PR #4, opened 2016-07-21
Adding check for httpoxy

### LINKED ISSUE #1: Warn on cgi usage
Go has a CGI package, and in some versions could be vulnerable to https://httpoxy.org/.

We could potentially warn about this..

callidus:
Go code running under CGI is vulnerable to httpoxy attack. See
https://httpoxy.org/ this checks for an import of net/http/cgi
that might indicate code may be run under CGI.

closes #1

### DIFF
--- a/rules/httpoxy.go
+++ b/rules/httpoxy.go
@@ -1,0 +1,50 @@
+// (c) Copyright 2016 Hewlett Packard Enterprise Development LP
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


rules/httpoxy.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	return
+}
+
+func NewHttpoxyTest() (r gas.Rule, n ast.Node) {
+	r = &HttpoxyTest{
+		MetaData: gas.MetaData{
+			Severity:   gas.High,
+			Confidence: gas.Low,
+			What:       "Go code running under CGI is vulnerable to Httpoxy attack.",
```
gcmurphy: I think maybe append (CVE-2016-5386) to the message. Will help people track down what the issue is in more detail.

callidus:
thanks @gcmurphy updated

PR #11, opened 2016-07-25
Make sure -fmt json produces valid output

csstaub:
r: @gcmurphy 

This makes sure that `-fmt json` produces output that is a valid JSON document. 

Before:
```
$ gas -fmt json main.go | head -n6
{
        "metrics": [
            Files: 1,
            Lines: 508,
            Nosec: 0,
            Issues: 1],
```

After:
```
$ gas -fmt json main.go | head -n7
{
        "metrics": {
            "files": 1,
            "lines": 508,
            "nosec": 0,
            "issues": 1
        },
```

In the future, would probably be better to serialize a struct with `encoding/json` instead of using templates.

### DIFF
--- a/output/formatter.go
+++ b/output/formatter.go
@@ -45,12 +45,12 @@ Summary:
 `
 
 var json = `{
-        "metrics": [
-            Files: {{.Stats.NumFiles}},
-            Lines: {{.Stats.NumLines}},
-            Nosec: {{.Stats.NumNosec}},
-            Issues: {{.Stats.NumFound}}],
-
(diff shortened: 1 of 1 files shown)


gcmurphy:
👍 thanks

csstaub:
FYI: This fixes the obvious issue, but some lines in code are still causing issues for me. I think `"` might not be properly escaped.

gcmurphy:
I agree we should probably look at using a proper JSON encoder for this output type.

PR #22, opened 2016-07-28
Use encoding/csv for CSV output

csstaub:
@gcmurphy @callidus
The encoding/csv package will take care of quoting, double-quoting,
and other CSV quirks -- avoids having to fiddle with text templates.


callidus:
LGTM, thanks @csstaub

PR #23, opened 2016-07-28
Detect use of rand.Read from math/rand

csstaub:
@gcmurphy @callidus 
Detects the use of `rand.Read` from math/rand in a code base. It's easy to accidentally import math/rand instead of crypto/rand for such uses since they both have a read function with the same type signature. It doesn't warn on other uses, because those APIs are explicitly different from the crypto/rand ones so I think they're less likely to be misused. 

Minimal sample that triggers an issue:
```Go
package main

import (
	"math/rand"
)

func main() {
	bad, err := rand.Read(nil)
}
```

Whereas this does not trigger:
```Go
package main

import (
	"crypto/rand"
)

func main() {
	good, err := rand.Read(nil)
}
```

### DIFF
--- a/rulelist.go
+++ b/rulelist.go
@@ -53,6 +53,7 @@ func newRulelist() rulelist {
 	rs.register("exec", rules.NewSubproc)
 	rs.register("errors", rules.NewNoErrorCheck)
 	rs.register("httpoxy", rules.NewHttpoxyTest)
+	rs.register("rand", rules.NewWeakRandCheck)
 	return rs
 }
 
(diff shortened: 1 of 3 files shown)


callidus:
@csstaub this looks great, but could you add a test file as per the other rules? you should be able to create one from your minimal sample easily.

csstaub:
Yes, can do. I'll update this with a test later.

callidus:
awesome, thanks

csstaub:
Added a couple of tests now

gcmurphy:
👍  thanks!

PR #24, opened 2016-07-28
Smarter hard-coded credentials check

csstaub:
@gcmurphy @callidus 
Check right-hand side expr for literals when looking for hard-coded credentials.

Example file:
```Go
package main

func main() {
  passwordBad := "password"
  passwordOk := getPassword()
}
```

With these changes, the first line in `main` will still warn, but the second line will now be fine. Previously both lines would produce a warning.

Anecdotally (based on running this on Square's internal Go code base) the hard-coded credentials check almost always yields false positives, since it's a common pattern to read a secret from e.g. a file and then load it into a variable like "password" or "token" or such.

### DIFF
--- a/rules/hardcoded_credentials.go
+++ b/rules/hardcoded_credentials.go
@@ -15,9 +15,10 @@
 package rules
 
 import (
-	gas "github.com/HewlettPackard/gas/core"
 	"go/ast"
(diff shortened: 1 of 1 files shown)


callidus:
nice 👍

PR #26, opened 2016-07-28
Fixing annotations

### LINKED ISSUE #25: #nosec comment not working
Adding a #nosec comment doesn't work. The code below is still reported as unsafe:

Code:
```go
cmd := exec.Command("sh", "-c", config.Command) // #nosec
```

And Log:
```
Subprocess launching with variable. (Confidence: HIGH, Severity: HIGH)
  > exec.Command(&#34;sh&#34;, &#34;-c&#34;, config.Command)
```

callidus:
The logic around annotations (nosec) was broken, meaning they were
ignored by default and would not skip sub-blocks. This fixes the
problem and also adds a test to make sure it wont be broken in the
future. Closes #25

### DIFF
--- a/core/analyzer.go
+++ b/core/analyzer.go
@@ -47,7 +47,7 @@ type Metrics struct {
 }
 
 type Analyzer struct {
-	annotations bool
+	ignoreNosec bool
 	ruleset     RuleSet
 	context     Context
 	logger      *log.Logger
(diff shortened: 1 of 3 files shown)


gcmurphy:
Might make more sense to s/!gas.annotations/gas.annotations/ here: https://github.com/HewlettPackard/gas/blob/master/core/analyzer.go#L139

The flag has a default of 'off'. That should propagate through rather than being inverted two times?

callidus:
The logic here was really broken (sorry my bad). Fixed it properly now

gcmurphy:
Thanks @callidus! 👍

Issue #28, opened 2016-07-30
Subprocess checks should apply to syscall.Exec() as well


marienz:
The checks for calling a subprocess look for `exec.Command`. It would probably make sense to apply the same checks to `syscall.Exec`.

callidus:
makes sense, will do. Thanks @marienz

PR #30, opened 2016-08-01
Extending subprocess launching tests to include "syscall.Exec"

callidus:
closes: #28

### DIFF
--- a/rules/subproc.go
+++ b/rules/subproc.go
@@ -15,10 +15,11 @@
 package rules
 
 import (
-	gas "github.com/HewlettPackard/gas/core"
 	"go/ast"
 	"regexp"
 	"strings"
+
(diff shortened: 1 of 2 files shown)


gcmurphy:
The syscall.Exec function has a different signature than os.Exec. The second argument (an array) will not match the *ast.BasicLit check and cause failures. 

```func Exec(argv0 string, argv []string, envv []string) (err error)```

callidus:
Good point, TBH the check is actually a bit lame. For example it will trip on exec.Cmmand("ls" + " -l") we should have a utility to determine if an expression contains any variables that we cant deduce the value of. We can iterate towards that though. I'll fix the problem with syscall first though

callidus:
OK, we need something much smarter here. I'll add the utility and then re-open this.

PR #31, opened 2016-08-01
Adding a config block to the analyzer, parsed from JSON

callidus:
A CLI option can now be given to tell GAS it should parse data
from a JSON file. Fatal errors are given if the file is not
readable or is not valid JSON.


gcmurphy:
Thanks @callidus.

PR #32, opened 2016-08-03
Try to resolve all elements in an expression to a known const

### LINKED ISSUE #28: Subprocess checks should apply to syscall.Exec() as well

callidus:
This is used in the subprocess launching test but will be added to
others as applicable.

I am also adding more (all) possible fields to the context info. Not
used in this patch but will be in the future.

This also closes #28

### DIFF
--- a/core/analyzer.go
+++ b/core/analyzer.go
@@ -92,9 +92,12 @@ func (gas *Analyzer) process(filename string, source interface{}) error {
 
 		// here we get type info
 		gas.context.Info = &types.Info{
-			Types: make(map[ast.Expr]types.TypeAndValue),
-			Defs:  make(map[*ast.Ident]types.Object),
-			Uses:  make(map[*ast.Ident]types.Object),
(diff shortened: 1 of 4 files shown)


callidus:
some examples:

OK
`        a := 1 + 1
        cmd := exec.Command("/bin/sleep", a)
`

OK
`
        b := [2]int{"/bin/", "sh"}
        a := 1
        cmd := exec.Command(b, a)
`


BAD
`
        b := someFunc()
        a := 1 + b
        cmd := exec.Command("cmd", a)
`

gcmurphy:
Looks good. One minor nit: 

I would like to see a test using syscall.Exec("/bin/cat", []string{ "/etc/passwd" }, nil)  or similar. It shows that we successfully handle arrays and nil values.

callidus:
added

PR #34, opened 2016-08-05
Creating blacklist import rules

callidus:
Creating a new generic blacklist rule and removing the older
specific ones. This will need configuration integration when
we have some.

The new test is immune to import aliasing but not shadowing


gcmurphy:
Looks good @callidus thanks!

PR #35, opened 2016-08-05
Configuration

callidus:
This re-works the way that CLI options are passed through to the
analyzer so that they can act as overrides for config options. If
not given on the CLI, options will come from a config file. If no
file is used then a default value is chosen.

Two lists are also populated with tests to include or exclude.
These lists are not used for now but will eventually replace the
way we select test to run in a future patch to follow.  NOTE that
the "exclude" CLI option has been repurposed to mean
excluded rules, a new "skip" option has been added to replace
the old behaviour.

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,8 +15,10 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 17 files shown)


main.go
```diff
@@ -57,6 +59,52 @@ USAGE:
...
+	} else {
+		val, ok := config["ignoreNosec"]
+		if !ok {
+			config["ignoreNosec"] = false
+		} else if _, ok := val.(bool); !ok {
+			logger.Fatal("Config value must be a bool: 'ignoreNosec'")
+		}
+	}
+
+	fmt.Println(config)
```
gcmurphy: Did you mean to leave the fmt.Println here?

callidus: ah good catch, will fix

PR #39, opened 2016-08-10
Rule selection rules

callidus:
This makes the following changes:
- riles are identified by an ID
- include / exclude list now work
- rules are selected based on these lists
- blacklist rules are broken out into methods
- rule constructors now take the config map
- config file can be used to select rules
- CLI options embelish config selection options

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	gas "github.com/HewlettPackard/gas/core"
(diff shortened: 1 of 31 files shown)


callidus:
There will be more tweaks to this to come, but it works and the interface shouldn't change too much so this is ok to merge if you have no issues with it.

gcmurphy:
Think you need to update the test cases? Tests currently not building.

callidus:
Heh. well good to know our CI is working well :P

PR #41, opened 2016-08-11
Fix usage information

gcmurphy:
Mostly a tidy up. Fixed a couple of spelling errors as well.


callidus:
Thanks Grant, good stuff

PR #46, opened 2016-08-24
prefix patterns with **/ to match subdirectories

### LINKED ISSUE #45: Gas checks _test files when multiple packages are found
src/pkg/violation_test.go has security violations

    cd src/pkg; gas .  // no errors are reported

    cd src; gas ./... // gas reports errors in violation_test.go files

    cd src; gas -skip=*_test.go ./... // gas repots errors in violation_test.go

drewwells:
fixes #45 

filepath.Match() requires the entire path is matched, prefixing ** allows `*_test.go` to work in multiple packages.

### DIFF
--- a/filelist.go
+++ b/filelist.go
@@ -57,6 +57,11 @@ func (f *filelist) Contains(pathname string) bool {
 			return true
 		}
 
+		// match file suffixes ie. *_test.go
+		if matched, _ := filepath.Match(filepath.Join("**", pattern), pathname); matched {
+			return true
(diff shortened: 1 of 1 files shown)


gcmurphy:
👍  Thanks. Hopefully we've caught all the edge cases around this now 😄

drewwells:
it is a somewhat overloaded function now. It's still not possible to skip a directory found anywhere in the path ie. `vendor` or `build`

PR #60, opened 2016-11-02
rand: resolve math/rand package

tam7t:
work in progress - could use a lot of refactoring

Fix for https://github.com/GoASTScanner/gas/issues/43 by resolving full package paths instead of regex on package name.

### DIFF
--- a/rules/rand.go
+++ b/rules/rand.go
@@ -16,8 +16,8 @@ package rules
 
 import (
 	"go/ast"
+	"go/types"
 	"regexp"
-	"strings"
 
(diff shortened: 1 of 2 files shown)


gcmurphy:
I think this is a nice start. I have been experimenting with using the typechecker to resolve this class of bug rather than pattern matching. Something like this would be a nice fit until I can figure out how all that will work.

tam7t:
Ah that makes a lot mores sense to just use that package. I updated the PR to use `go/types`

gcmurphy:
Perfect. This looks good.

PR #70, opened 2016-11-06
rand: use a MatchCallByPackage helper

tam7t:
Updated the `rand` rule to use the helper added by https://github.com/GoASTScanner/gas/pull/67

Also changed `MatchCallByObject` to `MatchCallByPackage` since it only matches functions on a package, not objects.

### DIFF
--- a/core/helpers.go
+++ b/core/helpers.go
@@ -52,11 +52,11 @@ func MatchCall(n ast.Node, r *regexp.Regexp) *ast.CallExpr {
 // package and identifier name matches the passed in parameters.
 //
 // Usage:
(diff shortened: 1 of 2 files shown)


gcmurphy:
Hmm not sure why this fails on 1.5 but passes on tip. Think it's probably an incompatibility issue with my helper function. I'll try to figure out what's happening here.

gcmurphy:
@tam7t Checkout this change 7a275fd0a. When I did some experimentation here it seems as though there is a big difference between 1.5.x and 1.7.x as far as the information in types.Object's. I've put together this slightly less optimal change to work around this and maintain support for the older 1.5.x release.

PR #92, opened 2016-11-18
Resolve issues with error rules

gcmurphy:


gcmurphy:
Going to merge and see if there are any complaints. The default whitelist will no doubt need to be improved.

PR #98, opened 2016-12-02
adding support for arbitrary paths with ellipses

endophage:
Changes:

- any path suffixed with `/...` will now work.
- changes to fileList and how the `-skip` values are handled necessary to support this.
- Golang does not support arbitrary length globbing in filepaths so inclusion of go-glob for simple string glob matching on `-skip` paths.
- Addition of vndr tool to manage dependencies.
- fixed capitalization of `filelist`. Some comments had `fileList` and it's technically 2 words so updating per Go defacto style conventions to `fileList`.

I'm a big fan of including the vendor directory as it protects projects from dependencies becoming unavailable. This isn't my project though so if you don't want it, or want a tool other than `vndr`, just let me know and I'll update.

~~Still working on some tests, will remove WIP later today once I've finished.~~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,5 @@ _testmain.go
 *.prof
 
 .DS_Store
+
+.vscode
--- a/filelist.go
+++ b/filelist.go
@@ -15,60 +15,58 @@
 package main
 
 import (
-	"fmt"
-	"os"
-	"path/filepath"
+	"log"
 	"strings"
+
+	"github.com/ryanuber/go-glob"
 )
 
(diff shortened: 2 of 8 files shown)


endophage:
@gcmurphy I've added all the tests that don't require fixtures. I can add tests for the file walking part but would suggest adding a test fixtures directory to create a known file structure that can be used for running those tests.

Let me know if you'd like me to do that, otherwise this is good to go.

gcmurphy:
👍  This looks good thanks!

PR #121, opened 2017-03-15
Add a check for PreferServerCipherSuites flag of tls.Config

cosmincojocar:
Having the server select the best available cipher suite, it is more secured than allowing the client to pick a weak cipher configuration.


gcmurphy:
Thanks for the PR!

Issue #124, opened 2017-03-26
The skip option does not work properly


cosmincojocar:
I tired the following combinations to skip the vendor folder but none of them was successful:

gas -skip vendor ./...
gas -skip vendor/* ./...
gas -skip \*vendor\* ./...
gas -skip vendor/*.go ./...

The vendor folder was always included into the scan.

gcmurphy:
The [changes here]( https://github.com/GoASTScanner/gas/blob/experimental/cmd/gas/main.go#L201) will address this. Effectively GAS _only_ scans complete packages, not individual files. By default it excludes the vendor directory and the files returned by a [built package](https://golang.org/src/go/build/build.go#L390) This will automatically exclude test files.

PR #128, opened 2017-04-10
Add support for partial path match in the skip option

cosmincojocar:
This is a fix for issue #124. The `skip` option is extended to support partial path match. Currently it is possible to exclude the vendor folder with this pattern only:

```
gas -skip "*vendor*" ./..
```

I extended the `fileList` to support also partial paths like this:

```
gas -skip vendor ./..
```


gcmurphy:
Getting the skip / inclusions lists right has been a bit of a nightmare. This looks like a sensible change

PR #129, opened 2017-04-10
Add a rule which audits the use of math/big.Int.Exp function call

cosmincojocar:
The big.Int.Exp might lead to DoS attacks if is used with a modulus equal with zero.

More details: https://www.cryptologie.net/article/347/my-first-cve-o-common-vulnerabilities-and-exposures/


gcmurphy:
Wasn't aware of this attack vector. Thanks!

PR #133, opened 2017-08-03
Adding Docker container and updating README

awiens:
This container can be built and run locally, in order to provide a simple way for users to run Gas on their code. The README has been updated to give information and provide build/run commands.


awiens:
I think we are intending to publish to dockerhub at some point, but wanted to just get it up on github first

PR #136, opened 2017-09-29
output/formatter: Issue.Line was already a string

lanzafame:
I also removed the `gas` binary that had been committed to the cmd/gas directory.


gcmurphy:
👍 Thanks!

PR #138, opened 2017-10-05
Adjust SQL format-string rules to ignore inherently safe formats

jonmcclintock:
If the only format specifiers in the format-string used to generate the SQL statement are ones that are guaranteed to only print safe characters (such as digits and decimals), then we shouldn't create an issue.


gcmurphy:
This is a great submission. Would you mind rebasing your change?

jonmcclintock:
Rebased to current master.

gcmurphy:
Thanks so much for your contribution!

PR #142, opened 2017-10-12
Add support for #excluding specific rules

jonmcclintock:
Create the ability to exclude specific rules, rather than all of them (with "#nosec"). Works like:

                cmd := exec.Command("sh", "-c", os.Getenv("FOO")) // #exclude !G001

You can specify an arbitrary number of exclusions, and they have the same scoping semantics as "#nosec". You can also add comments to explain your exclusions:

                cmd := exec.Command("sh", "-c", os.Getenv("FOO")) // #exclude !G001: Doesn't apply here


jonmcclintock:
Have you had a chance to take a look at this?

gcmurphy:
Sorry I've not been getting many cycles to OSS work. I'd like to get https://github.com/GoASTScanner/gas/pull/146 merged before this lands. 

Awesome work though.

jonmcclintock:
Sure thing, just give me a heads up when it lands and I can refactor it to
work with the new revision.

-Jon

gcmurphy:
@jonmcclintock not sure if i pinged at you about the rebasing this. if you don't have time I can look at doing it.

cosmincojocar:
I am also interested in this feature. I think we need only to exclude the rules which produce false positive results. Do you see any use case which requires specific inclusion? Typically all rules are enabled.

What do you think about this syntax since there is already `#nosec` directive:

```
#sec !G101,!G102
```

The build constrains in Go could be a source of inspiration: https://golang.org/pkg/go/build/

jonmcclintock:
I will be able to get to this next week. Feel free to do if it you've got
the time before then.

-Jon

jonmcclintock:
Finally got around to rebasing this. Sorry for the delay.

jonmcclintock:
@cosmincojocar I'm not wed to "exclude" by any means, and "sec" may make more sense. I'll leave it to @gcmurphy to figure out what fits in with his vision of things...

gcmurphy:
This is an awesome change and thanks for getting back to it. I'm wondering if it is worth tweaking it a little bit so it is more of an extension to the #nosec directive.

Effectively: `#nosec` = Ignore all rules
And: `#nosec G101, G202` = ignore specifically these rules. 

@jonmcclintock thoughts? Would that be a difficult change to make? If so I'm happy to merge this as is and iterate on it.

jonmcclintock:
Wouldn't be that hard at all, just a couple of quick changes here:

  https://github.com/GoASTScanner/gas/pull/142/files#diff-21c43657b6fe53d9635ac2f42896ee61R152

-Jon

jonmcclintock:
Pushed an updated PR with the changed exclusion syntax.

PR #146, opened 2017-12-13
Merge experimental / refactor

gcmurphy:
I've had a number of changes that I've been sitting on for a while that I think need to get merged to move forward. 

The most notable:

* gas root repository acts more like a library, with the cmd being moved off to cmd/gas directory.
* gas will only process valid go packages rather than individual files. This significantly simplifies the file include/exclude file logic as it works as the compiler does.  To recursively evaluate packages you can still run: gas $GOPATH/src/github.com/GoASTScanner/gas/... for example. The vendor directory and any test files will be excluded by default.
* Moved to using ginkgo for writing tests. 

There is still much to do but this change is getting too big, and am getting pinged with issues & PR's that won't be relevant under this model.


cmd/gas/main.go
```diff
@@ -0,0 +1,232 @@
...
+	logger = log.New(logWriter, "[gas] ", log.LstdFlags)
+
+	// Load config
+	config, err := loadConfig(*flagConfig)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
+	// Load enabled rule definitions
+	ruleDefinitions := loadRules(*flagRulesInclude, *flagRulesExclude)
```
cosmincojocar: Are all rules included if both flags are empty? If all are included, I would print a info messages here. If all are excluded, I would print an error.

cmd/gas/main.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// Create the analyzer
+	analyzer := gas.NewAnalyzer(config, logger)
+	analyzer.LoadRules(ruleDefinitions.Builders()...)
+
+	vendor := regexp.MustCompile(`[\\/]vendor([\\/]|$)`)
```
cosmincojocar: Maybe is worth make it this configurable. I think we can get a list of folders which could be potentially excluded. I would add **vendor** folder by defualt to this list.

cmd/gas/main.go
```diff
@@ -0,0 +1,232 @@
...
+
+		// Skip vendor directory
+		if vendor.MatchString(pkg) {
+			continue
+		}
+
+		abspath, _ := filepath.Abs(pkg)
+		logger.Println("Searching directory:", abspath)
+		if err := analyzer.Process(pkg); err != nil {
+			logger.Fatal(err)
```
cosmincojocar: Maybe it is worth continuing the processing if only one package fails.

config.go
```diff
@@ -0,0 +1,82 @@
...
+}
+
+// Set section in the configuration to specified value
+func (c Config) Set(section string, value interface{}) {
+	c[section] = value
+}
+
+// GetGlobal returns value associated with global configuration option
+func (c Config) GetGlobal(option string) (string, error) {
+	if globals, ok := c["global"]; ok {
```
cosmincojocar: I would extract this string into a constant.

config.go
```diff
@@ -0,0 +1,82 @@
...
+			return "", fmt.Errorf("global setting for %s not found", option)
+		}
+	}
+	return "", fmt.Errorf("no global config options found")
+
+}
+
+// SetGlobal associates a value with a global configuration ooption
+func (c Config) SetGlobal(option, value string) {
+	if globals, ok := c["global"]; ok {
```
cosmincojocar: The same here.

rules/rulelist.go
```diff
@@ -0,0 +1,93 @@
...
+		builders = append(builders, def.Create)
+	}
+	return builders
+}
+
+type RuleFilter func(string) bool
+
+func NewRuleFilter(action bool, ruleIDs ...string) RuleFilter {
+	rulelist := make(map[string]bool)
+	for _, rule := range ruleIDs {
```
cosmincojocar: I would validate if the ruleID is really defined in the list. If not, I will skip it.

testutils/source.go
```diff
@@ -0,0 +1,193 @@
+package testutils
+
+// CodeSample encapsulates a snippet of source code that compiles, and how many errors should be detected
+type CodeSample struct {
+	Code   string
+	Errors int
+}
+
+var (
```
cosmincojocar: Do you plan to add more code samples here to cover all the rules? I am asking this because the tests of each rule were removed.

analyzer.go
```diff
@@ -0,0 +1,180 @@
...
+	builtPackage, err := packageConfig.Load()
+	if err != nil {
+		return err
+	}
+
+	for _, pkg := range builtPackage.Created {
+		gas.logger.Println("Checking package:", pkg.String())
+		for _, file := range pkg.Files {
+			gas.logger.Println("Checking file:", builtPackage.Fset.File(file.Pos()).Name())
+			gas.context.FileSet = builtPackage.Fset
```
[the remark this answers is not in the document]
gcmurphy: I think this is something that we can look at, although  some of the context like type information is relevant between files in a package I think. Parallel processing is something I'd like to work towards.

analyzer_test.go
```diff
@@ -0,0 +1,113 @@
...
+		logger   *log.Logger
+		output   *bytes.Buffer
+	)
+	BeforeEach(func() {
+		logger, output = testutils.NewLogger()
+		analyzer = gas.NewAnalyzer(nil, logger)
+	})
+
+	Context("when processing a package", func() {
+
```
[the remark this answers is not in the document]
gcmurphy: Agreed. Done.

cmd/gas/main.go
gcmurphy: Agreed. Done.

cmd/gas/main.go
gcmurphy: The file inclusion / exclusion logic in the past has been a major source of bugs and confusion. The aim here was to basically make gas work as simply as possible. 

However I agree  it would make sense to be able to configure package directories in the recursive case. Do you think we should fix this as a separate PR or try to sneak it in with change?

cmd/gas/main.go
gcmurphy: I think the causes of errors here are: 

1. It is a valid go package but it doesn't compile
2. It is not a valid go package

I think I was trying to mimic the compiler behaviour here. For the purposes of a mass scan I 
could see how this would be useful to continue processing. 

However I would typically do something like to scan everything in my workspace: 

```bash
# (mac osx)
$ cd $GOPATH/src && go list ./... | xargs  -I {} gas $GOPATH/src/{}/...
```

testutils/source.go
gcmurphy: Yes. This is something I need to finish doing.  There are also some placeholder tests that I need to get finished also (currently skipping).

cmd/gas/main.go
cosmincojocar: In my opinion the vendor folder exclusion is enough. A more complex logic can be added after the merge.

cmd/gas/main.go
cosmincojocar: Agreed. The workaround is fair enough.

testutils/source.go
cosmincojocar: Cool!

helpers_test.go
```diff
@@ -2,13 +2,13 @@ package gas_test
 
 import (
```
cedturk, line 3: ben hiç bir şey anlamadım işe geç kalacağım yardım

cedturk: yardım edin biraz

PR #151, opened 2018-01-07
Cache built packages in analyzer

ashanbrown:
This change avoids reloading (using `packageConfig.Load()`) any packages that have already been loaded as a side-effect of previously loaded packages.  This branch runs about 30x times faster for a large application I work on, and runs gas on itself nearly 2x faster.  This change also only caches packages that we want to include in the report.  If we want to run even faster I think it would be necessary to cache all packages -- not just the packages we are targeting -- but I think that would require not using the loader as is.  I'm not quite sure I've handled relative paths quite correctly, but it works for the project I'm testing it on as well as on gas itself.

I'm pretty sure gas used to run much faster (possibly before switching to package imports) but I'm not sure exactly what version I was using before (unfortunately for us it was cached in CI so I didn't notice initially and I've had some trouble running old versions).  It could also be that some other dependency changed and is responsible for slowness (several seconds of load time) when loading entire packages.

Here's the results when I run `time gas` on itself using gas built from this branch:

```
time gas -log /dev/null ./...
2018/01/07 13:41:50 including rules: default
2018/01/07 13:41:50 excluding rules: default
Results:
...
Summary:
   Files: 33
   Lines: 4181
   Nosec: 10
  Issues: 7

gas -log /dev/null ./...  9.59s user 1.37s system 369% cpu 2.969 total
```

Below are the results when running `time gas` on itself from master:
```
time gas -log /dev/null ./...
2018/01/07 13:42:11 including rules: default
2018/01/07 13:42:11 excluding rules: default
Results:
...
Summary:
   Files: 33
   Lines: 4135
   Nosec: 10
  Issues: 7

gas -log /dev/null ./...  14.28s user 1.60s system 321% cpu 4.940 total
```

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -54,13 +54,19 @@ type Metrics struct {
 // Analyzer object is the main object of GAS. It has methods traverse an AST
 // and invoke the correct checking rules as on each node as required.
 type Analyzer struct {
-	ignoreNosec bool
-	ruleset     RuleSet
-	context     *Context
-	config      Config
-	logger      *log.Logger
-	issues      []*Issue
-	stats       *Metrics
+	ignoreNosec   bool
+	ruleset       RuleSet
+	context       *Context
+	config        Config
+	logger        *log.Logger
+	issues        []*Issue
+	stats         *Metrics
+	builtPkgCache map[string]*buildPackageCacheEntry
+}
+
+type buildPackageCacheEntry struct {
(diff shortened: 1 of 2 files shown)


ashanbrown:
Closing this in favor of https://github.com/GoASTScanner/gas/pull/152, which just does one load instead of trying to cache targeted packages.

PR #152, opened 2018-01-07
Do a single build for all packages

ashanbrown:
This is much faster because the loader can reuse packages.

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -27,6 +27,8 @@ import (
 	"reflect"
 	"strings"
 
+	"path/filepath"
+
 	"golang.org/x/tools/go/loader"
 )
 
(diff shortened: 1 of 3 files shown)


ashanbrown:
This branch runs gas on itself over 2x faster, according to the results below.

Time to run gas on itself with this branch:
```
>time gas -out /dev/null  ./...
...
gas -out /dev/null ./...  3.32s user 0.53s system 307% cpu 1.253 total
```

Time to run gas on itself in master:
```
>time gas -out /dev/null  ./...
...
gas -log /dev/null ./...  14.28s user 1.60s system 321% cpu 4.940 total
```

gcmurphy:
Thanks for this! I did notice the performance took a huge hit from the last big change but hadn't had a chance to dig into it. It makes sense that this would be a large contributor to slow downs.

Issue #158, opened 2018-01-25
JUnit XML output format


wongherlung:
Hi,

I've integrate gas into my CI workflow, which requires JUnit XML formatted results. Currently, I've created a simple program to parse the `json` output from gas into Junit XML, which I thought it would be nice as an enhancement for gas.

Would it be okay if I do a PR to implement an option to output the results in Junit XML? Also there is an issue of multiple types of XML format, as mentioned in #120. What are your thoughts on this if we were to implement `xml` format for gas?

gcmurphy:
Sounds like a great addition I look forward to the PR. There shouldn't be any problem with multiple XML formats, we could name one as -junit and the other -checkstyle if it became confusing.

gcmurphy:
Fixed by #160

PR #160, opened 2018-01-27
JUnit XML output

wongherlung:
This PR creates a new output format, namely: `JUnit XML` format, following [IBM's definition](https://www.ibm.com/support/knowledgecenter/en/SSQ2R2_9.5.0/com.ibm.rsar.analysis.codereview.cobol.doc/topics/cac_useresults_junit.html).

This allows the results generated by `gas` to be used by CI tools, like Jenkins, which are able to parse/read `JUnit XML` results in their various workflows.

Example command to output `JUnit XML`:
```
gas -exclude=G104 -fmt=junit-xml -out=report.xml ./...
```

This PR is in reference to #158.

### DIFF
--- a/output/formatter.go
+++ b/output/formatter.go
@@ -74,8 +74,8 @@ func CreateReport(w io.Writer, format string, issues []*gas.Issue, metrics *gas.
 		err = reportJSON(w, data)
 	case "csv":
 		err = reportCSV(w, data)
-	case "xml":
-		err = reportXML(w, data)
+	case "junit-xml":
+		err = reportJUnitXML(w, data)
 	case "html":
 		err = reportFromHTMLTemplate(w, html, data)
 	case "text":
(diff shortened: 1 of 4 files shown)


output/formatter.go
```diff
@@ -112,6 +118,25 @@ func reportCSV(w io.Writer, data *reportInfo) error {
...
+	raw, err := xml.MarshalIndent(junitXMLStruct, "", "\t")
+	if err != nil {
+		panic(err)
+	}
+
+	xmlHeader := []byte("<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n")
+	raw = append(xmlHeader, raw...)
+	_, err = w.Write(raw)
+	if err != nil {
+		panic(err)
```
gcmurphy: could we just return the error here instead of panic?

wongherlung: opps. yes, should return an error here.

gcmurphy:
Thanks this looks a great addition.

Issue #159, opened 2018-01-26
Gas can't find functions if the "C" import is used


jonmcclintock:
If a source file imports "C", gas loses the ability to resolve functions in that file. In the attached example:

> gas repro/
> [gas] 2018/01/26 22:37:13 including rules: default
> [gas] 2018/01/26 22:37:13 excluding rules: default
> [gas] 2018/01/26 22:37:13 Searching directory: /home/jonm/go/src/gas-package-repro/repro
> /home/jonm/go/src/gas-package-repro/repro/target.go:9:7: undeclared name: testFunction
> [gas] 2018/01/26 22:37:13 couldn't load packages due to errors: repro

However, if you comment out the `import "C"` line in def.go, it runs fine:

> gas repro/
> [gas] 2018/01/26 22:40:42 including rules: default
> [gas] 2018/01/26 22:40:42 excluding rules: default
> [gas] 2018/01/26 22:40:42 Searching directory: /home/jonm/go/src/gas-package-repro/repro
> [gas] 2018/01/26 22:40:42 Checking package: repro
> [gas] 2018/01/26 22:40:42 Checking file: /home/jonm/go/src/gas-package-repro/repro/def.go
> [gas] 2018/01/26 22:40:42 Checking file: /home/jonm/go/src/gas-package-repro/repro/target.go
> Results:
> 
> Summary:
>    Files: 2
>    Lines: 19
>    Nosec: 0
>   Issues: 0

[gas-package-repro.tar.gz](https://github.com/GoASTScanner/gas/files/1669416/gas-package-repro.tar.gz)

gcmurphy:
This looks to be related to https://github.com/golang/go/issues/11627

gcmurphy:
Fixed by #161

PR #161, opened 2018-01-29
[Issue 159] Allow loader errors so that processing continues

jonmcclintock:
This allows processing to continue to work around the loader issue identified in #159 .


gcmurphy:
👍 thanks for this.

PR #164, opened 2018-02-06
Add a rule to audit the usage of ssh.InsecureIgnoreHostKey

cosmincojocar:
cc @gcmurphy

### DIFF
--- a/rules/ssh.go
+++ b/rules/ssh.go
@@ -1,17 +1,3 @@
-// (c) Copyright 2016 Hewlett Packard Enterprise Development LP
-//
-// Licensed under the Apache License, Version 2.0 (the "License");
-// you may not use this file except in compliance with the License.
-// You may obtain a copy of the License at
-//
-//     http://www.apache.org/licenses/LICENSE-2.0
-//
-// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


rules/ssh.go
```diff
@@ -0,0 +1,47 @@
+// (c) Copyright 2016 Hewlett Packard Enterprise Development LP
```
gcmurphy: I am no longer am associated with HP and since this contribution is from you I don't think this header needs to be here. 

I probably need to create a LICENSE, AUTHORS, and COPYRIGHT files to better track contributions and copyright information.

testutils/source.go
```diff
@@ -183,6 +183,15 @@ func main() {
...
 }`, 1}}
 
+	// SampleCodeG106 - ssh InsecureIgnoreHostKey
+	SampleCodeG106 = []CodeSample{{`
+package main
+import (
+        "golang.org/x/crypto/ssh"
+)
+func main() {
+        _ :=  ssh.InsecureIgnoreHostKey()
```
gcmurphy: Think you need to change ':=' to '=' for the test to run.

cosmincojocar:
Thanks for reviewing my changes. I pushed the fixes.

gcmurphy:
Thanks this is a great addition.

PR #167, opened 2018-02-08
Sort the issues by severity in descending order

cosmincojocar:
Sort the issues by severity in descending order before creating the report. This will print the most critical issues at the top of the report. 

This is useful when doing security audit in large code bases with hundred of warnings. 

cc @gcmurphy

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,9 @@ language: go
 before_script:
   - go vet $(go list ./... | grep -v /vendor/)
 go:
-  - 1.5
+  - 1.7
+  - 1.8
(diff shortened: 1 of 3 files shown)


gcmurphy:
sort.Slice was introduced in Go 1.8, the travis job will try to build against 1.5 and latest. I'm not sure if we need to support versions as far back as 1.5..but I think it's reasonable to support N-2 which would include 1.7.

cosmincojocar:
I switched to old style sort interface. In addition, I update the build file to validate the tool from go version 1.7 onward.

PR #168, opened 2018-02-08
Rule/lfi

coredefend:
This rule flags potential file inclusion bugs. It's modeled after the `subproc` rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ or to specify a set of rules to explicitly exclude using the '-exclude=' flag.
   - G103: Audit the use of unsafe block
   - G104: Audit errors not checked
   - G105: Audit the use of math/big.Int.Exp
+  - G106: Audit the use of ssh.InsecureIgnoreHostKey
   - G201: SQL query construction using format string
   - G202: SQL query construction using string concatenation
   - G203: Use of unescaped data in HTML templates
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ $ gas -nosec=true ./...
 
 ### Output formats
 
-Gas currently supports text, json and csv output formats. By default
+Gas currently supports text, json, csv and JUnit XML output formats. By default
 results will be reported to stdout, but can also be written to an output
 file. The output format is controlled by the '-fmt' flag, and the output file is controlled by the '-out' flag as follows:
 
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ or to specify a set of rules to explicitly exclude using the '-exclude=' flag.
   - G301: Poor file permissions used when creating a directory
   - G302: Poor file permisions used with chmod
   - G303: Creating tempfile using a predictable path
+  - G304: Audit use of reading files
   - G401: Detect the usage of DES, RC4, or MD5
   - G402: Look for bad TLS connection settings
   - G403: Ensure minimum RSA key length of 2048 bits
@@ -113,4 +114,3 @@ file. The output format is controlled by the '-fmt' flag, and the output file is
 # Write output in json format to results.json
 $ gas -fmt=json -out=results.json *.go
 ```
-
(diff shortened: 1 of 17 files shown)


jovonuber:
@gcmurphy Howdy! I created a rule to check for file inclusion bugs. Can you review and provide feedback? Thanks!

README.md
```diff
@@ -49,6 +49,7 @@ or to specify a set of rules to explicitly exclude using the '-exclude=' flag.
   - G301: Poor file permissions used when creating a directory
   - G302: Poor file permisions used with chmod
   - G303: Creating tempfile using a predictable path
+  - G304: Audit use of reading files
```
cosmincojocar: I would call this rule `File path provided as taint input`.

rules/readfile.go
```diff
@@ -0,0 +1,50 @@
...
+	"go/types"
+
+	"github.com/GoASTScanner/gas"
+)
+
+type readfile struct {
+	gas.CallList
+}
+
+// Match inspects AST nodes to determine if the match the methods `os.Open` or `ioutil.ReadFile`
```
cosmincojocar: Maybe it makes sense to catch the entire family of function which can modify files, such as os.Create, os.Truncate, os.Rename, os.Remove ...

rules/rulelist.go
```diff
@@ -77,6 +77,7 @@ func Generate(filters ...RuleFilter) RuleList {
 		"G301": {"Poor file permissions used when creating a directory", NewMkdirPerms},
 		"G302": {"Poor file permisions used when creation file or using chmod", NewFilePerms},
 		"G303": {"Creating tempfile using a predictable path", NewBadTempFile},
+		"G304": {"Audit use of file inclusion", NewReadFile},
```
cosmincojocar: See my suggestion above regarding the name of this rule.

rules/readfile.go
```diff
@@ -0,0 +1,50 @@
...
+	if node := r.ContainsCallExpr(n, c); node != nil {
+		for _, arg := range node.Args {
+			if ident, ok := arg.(*ast.Ident); ok {
+				obj := c.Info.ObjectOf(ident)
+				if _, ok := obj.(*types.Var); ok && !gas.TryResolve(ident, c) {
+					return gas.NewIssue(c, n, "File inclusion launched with variable", gas.Medium, gas.High), nil
+				}
+			}
+		}
+		return gas.NewIssue(c, n, "File inclusion launching should be audited", gas.Low, gas.High), nil
```
cosmincojocar: Isn't a bit to much this warning. This will trigger an issues for all files regardless if they are provided as a taint input or not. This will produce potentially to many false positive warnings.

testutils/source.go
```diff
@@ -459,6 +459,49 @@ func main() {
 	ioutil.WriteFile("/tmp/demo2", []byte("This is some data"), 0644)
 }`, 2}}
 
+	// SampleCodeG304 - potential file inclusion vulnerability
```
cosmincojocar: I would add also a test case for `flags`.

rules/readfile.go
jovonuber: Yeah, I was thinking this also. But it might make sense to separate out vulns wrt Read and Write. However both read/write follow the same vuln pattern (e.g., tainted input going to the sensitive function)

rules/rulelist.go
jovonuber: will do

rules/readfile.go
jovonuber: I followed the same pattern for the `subproc` rule.  What should I do in this case?

Does GoAST have a lib that can do some data-flow analysis that pinpoints where the variable was assigned and possibly tainted? I played around with the `Context` struct but wasn't able to get the results I wanted.

cosmincojocar: > I followed the same pattern for the subproc rule. What should I do in this case?

I would return nil instead.

> Does GoAST have a lib that can do some data-flow analysis that pinpoints where the variable was assigned and possibly tainted? I played around with the Context struct but wasn't able to get the results I wanted.

I found this package which does some data flow analysis https://github.com/godoctor/godoctor/tree/master/analysis/dataflow. It would be great if you could try it out and come up with a more generic solution for taint input vulnerability.

Issue #77, opened 2016-11-08
Add version information in builds


gcmurphy:
In preparation to move towards Beta and first official release need to embed version information with each build and move towards proper release management using milestones.

PR #176, opened 2018-02-15
Add a semantic version to the usage text

cosmincojocar:
Fix for #77 

cc @gcmurphy

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,8 +113,40 @@ file. The output format is controlled by the '-fmt' flag, and the output file is
 # Write output in json format to results.json
 $ gas -fmt=json -out=results.json *.go
 ```
+### Development
 
-### Generate TLS rule
+#### Release Build
+
+Gas can be released as follows:
+
+```bash
+# Prepare the build variables
+export VERSION=2.0.0
+export GIT_TAG=$(git describe --always --tags)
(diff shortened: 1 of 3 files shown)


gcmurphy:
Thanks this looks good and will help with debugging issues. I'm wondering if we could pass this in somehow at build time rather than hardcoding it? Similar to this https://medium.com/@joshroppo/setting-go-1-5-variables-at-compile-time-for-versioning-5b30a965d33e or maybe we should investigate using something like https://github.com/ahmetb/govvv 

I'd like to start using git tags and doing proper releases soon.

cosmincojocar:
These approaches seem interesting. I'll have a look to them.

I am also working on a Makefile which will automate the build, test, release and docker image build steps. The script will get the version from the latest git tag as follows:

```
VERSION ?= $(shell git describe --always --tags)
BIN = gas
BUILD_CMD = go build -o build/$(BIN)-$(VERSION)-$${GOOS}-$${GOARCH} &
```
I hope to create a PR till tomorrow.

cosmincojocar:
@gcmurphy I will follow up with the Makefile in a different PR

cosmincojocar:
@gcmurphy I am closing this in favor of PR #179.

Issue #173, opened 2018-02-13
Support a yaml encoding formatter


ericwb:
It would be nice to have a yaml output formatter in addition to json, csv, html, xml, etc.

gcmurphy:
Fixed in #177

PR #177, opened 2018-02-16
Add support for YAML output format

cosmincojocar:
This is a fix for #173. 

cc @gcmurphy


gcmurphy:
Thanks this is great.

PR #178, opened 2018-02-19
Add a tool to generate the TLS configuration form Mozilla's ciphers recommendation

### LINKED ISSUE #15: Use hosted JSON guidelines for TLS evaluation
Howdy folks. Author of the Mozilla TLS guidelines here. Thanks for the great tool! I'm actually looking at using it in a number of Go projects we run.

One improvement I'd suggest to make would be to use the JSON version of the guidelines instead of hardcoding the recommendations. We host the file at https://statics.tls.security.mozilla.org/server-side-tls-conf.json You can refer to it in the code and fall back to hardcoded defaults if the retrieval fails.

As an example, here is how we use the JSON recommendations in the evaluation worker in the TLS Observatory works: https://github.com/mozilla/tls-observatory/blob/master/worker/mozillaEvaluationWorker/mozillaEvaluationWorker.go

cosmincojocar:
This tool can generate the TLS configuration from Mozilla's ciphers recommendation and it can be invoked with `go generate`.  

The configuration of TLS rule was generated using the tool. I updated also the README file with some instructions how to do this.

Fix #15 

cc @gcmurphy


gcmurphy:
Awesome work!

PR #186, opened 2018-03-08
Switch to valuespec instead of gendecl for hardcoded credential rule

andyleap:
The current hardcoded credentials rule fails to handle #nosec comments in const decl blocks, like
```
const (
    Token = "qw12r4jjwq1rdvsp" //#nosec
)
```
This change allows this to be handled by the existing analyzer ignore system.


gcmurphy:
Thanks. Nice pickup.

PR #201, opened 2018-04-18
Add an option for Go build tags

### LINKED ISSUE #157: Option to pass build parameters
Is there an option to pass build parameters to `gas` as passed to `go build` command?

There are some source files in our repository that don't get built unless some tags are passed, e.g. we build them with `go build -tags "debug"`. Since we cannot pass the tags to gas, they are not checked. Are there any ways to pass them, if not, is it reasonable to implement it? Thanks!

cosmincojocar:
Add an option to Go build tags (https://golang.org/pkg/go/build/)

@Fixes #157 

cc @gcmurphy

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,14 @@ can do the following:
 ```
 $ gas -nosec=true ./...
 ```
+#### Build tags
+
+Gas is able to pass your [Go build tags](https://golang.org/pkg/go/build/) to the analyzer.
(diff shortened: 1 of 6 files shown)


cosmincojocar:
I updated the README file with a section which describes how to use the build tags.

gcmurphy:
Awesome thanks.

PR #206, opened 2018-04-27
Integrate the goreleaser tool

### LINKED ISSUE #194: Look at integrating https://goreleaser.com/
In addition to the recent build changes it might be worth looking at integrating with goreleaser

cosmincojocar:
Fixes #194 

cc @gcmurphy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 _obj
 _test
 vendor
+dist
 
 # Architecture specific extensions/prefixes
 *.[568vq]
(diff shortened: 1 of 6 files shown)


cosmincojocar:
@gcmurphy I think we could drop the support for 1.8 since the go teams already did it. For instance, the crypto/ssh is not anymore compatible with 1.8 because of the math/bits package introduced in 1.9. 

https://github.com/golang/go/issues/23353

I seems that Travis still builds the 1.8 even though was removed from the configuration. I will create new a new PR.

Issue #85, opened 2016-11-14
Need a new project / org name


gcmurphy:
This project has been about for longer than ours: https://github.com/go-gas/gas so I guess we should look at coming up with a different name. 

@GoASTScanner/committers  Any suggestions for new project name / org name?

endophage:
Something that emphasizes the security aspect rather than the AST aspect would be great 👍

alaska:
Cogminder
Rearguard - Connotes vigilence / pun for CYA
Argus - Greek mythology, a many-eyed giant. Guardian, sees all.
Daimon - Sort of personal spirit guardians/tutors from Greek mythology.
Securitas - yeah, turns out there's a god for that.
Stata Mater - Another Roman pantheon goddess. She protected against fires. And, hey, what do security bugs cause? :D
Coeus - titan of rational knowledge
Aegis - shield/protective artifact

cosmincojocar:
What about **gosec**, it sounds more appropriate for this tool.

thiderman:
The target project is comptely empty now though? Maybe a name change is not needed?

gcmurphy:
Finally got around to doing this!

PR #207, opened 2018-04-30
Integrate the goreleaser tool

### LINKED ISSUE #194: Look at integrating https://goreleaser.com/

ccojocar:
Fixes #194

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 
 go:
-  - 1.8
   - 1.9
   - "1.10"
   - tip


gcmurphy:
@cosmincojocar  I'm thinking of just renaming the github org to something shorter and keeping the name of 'gas'.  maybe changing the acronym to mean 'Go Application Security' or something similar as per #85 

Will it be better to do this before / after merging the goreleaser work?

cosmincojocar:
@gcmurphy I think, it's a good idea to do the renaming before publishing any release. Do you have any specific name in mind? `gosec` sounds good to me. It fits the current purpose of the tool and it's inline with the naming conversion used by other go tools. 

I am also fine to keep the current name and just rename the organization to `gas`.

ccojocar:
@gcmurphy Do you intend to merge this PR?

gcmurphy:
I've been meaning to do the rename (at least to the github org) first, but haven't had much time to devote to this project due to other life commitments. I'll add you as a core dev to the project so things are less blocked on my approvals.

ccojocar:
Thanks! I joined the GitHub org. Which name would you like to use for the GitHub org?

gcmurphy:
@ccojocar renamed to 'securego' as it was available I thought we could create another repo under it securego/guide which is documentation about the logic behind the rules. This is a common request, and we could actually publish it as a git book or something similar and reference urls in the rule output. 

I think we can probably rename the tool if you like as well? Your suggestion was gosec right?

ccojocar:
It sounds great! Yes, my suggestion was `gosec`, which sounds a bit more aligned with other Go tools.

ccojocar:
You can also rename the repo. I'll try to fix the code and documentation.

gcmurphy:
Rename of repo is done. @ccojocar f you can handle the code / docs update from gas -> gosec that'd be great!

ccojocar:
I am closing this in favour of #219.

PR #208, opened 2018-05-01
Rule which detects a potential path traversal when extracting zip archives

### LINKED ISSUE #205: Potential rule: path traversal when extracting zip archives
It would be fairly typical to include the file path from the zip file when extracting it's contents to disk
like in this: 

```go
import (
	"archive/zip"
	"io"
	"os"
	"path/filepath"
	"strings"
)

func unzip(archive, target string) error {
	reader, err := zip.OpenReader(archive)
	if err != nil {
		return err
	}

	if err := os.MkdirAll(target, 0755); err != nil {
		return err
	}

	for _, file := range reader.File {
		path := filepath.Join(target, file.Name)
		if file.FileInfo().IsDir() {
			os.MkdirAll(path, file.Mode())
			continue
		}

		fileReader, err := file.Open()
		if err != nil {
			return err
		}
		defer fileReader.Close()

		targetFile, err := os.OpenFile(path, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, file.Mode())
		if err != nil {
			return err
		}
		defer targetFile.Close()

		if _, err := io.Copy(targetFile, fileReader); err != nil {
			return err
		}
	}

	return nil
}
```
[(original source)](https://gist.githubusercontent.com/svett/424e6784facc0ba907ae/raw/2fd88e050ce48023226a54c5ab39ee93447b8d20/extract.go)

It could be possible to create a zip archive with: `../../../../../some_file` to achieve path traversal here. 
We should investigate whether this could be detected.  I think if we detect filepath.Join inclusive of zip.File type, without sanitisation, and the use of that identifier in a write operation you could get a fairly high confidence here.

ccojocar:
@gcmurphy I am giving a try to implement a rule which detects a potential path traversal when extracting zip archives. I would be happy to get your feedback. Thanks

Fixes #205

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 
 go:
-  - 1.8
   - 1.9
   - "1.10"
   - tip
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ or to specify a set of rules to explicitly exclude using the '-exclude=' flag.
   - G302: Poor file permisions used with chmod
   - G303: Creating tempfile using a predictable path
   - G304: File path provided as taint input
+  - G305: File traversal when extracting zip archive
   - G401: Detect the usage of DES, RC4, or MD5
   - G402: Look for bad TLS connection settings
   - G403: Ensure minimum RSA key length of 2048 bits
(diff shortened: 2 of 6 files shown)


ccojocar:
@gcmurphy Please could you have a look at this PR?

PR #211, opened 2018-06-03
Make G204 look for CommandContext calls

WillAbides:
[exec.CommandContext](https://golang.org/pkg/os/exec/#CommandContext) is essentially the same thing as exec.Command.  This PR makes gas treat it the same.


WillAbides:
Merged master and fixed the build

PR #216, opened 2018-07-19
Rename gas with gosec

ccojocar:


vprithvi:
This rename breaks builds installing gas using `go get`, consider keeping the `gas` cmd (or perhaps aliasing it) so that users have time to migrate?

PR #217, opened 2018-07-23
Derive the package from given files

ccojocar:
Move some utility functions into the helper


ccojocar:
@gcmurphy This fix is required in order to fix the gometalinter test which provides some files relative path.

PR #223, opened 2018-07-30
Add a flag to specify the severity for which the scanning will be failed

### LINKED ISSUE #212: Option to suppress issues which have medium and low priority
Want an option to suppress errors which are medium and low and not to fail the build/execution

Go version  go version go1.10.3 darwin/amd64
 Operating system / Environment -MAC

Expected behavior  - Want an option to suppress errors which are medium and low so that it does not  fail the build/execution
Actual behavior - Even medium and low issues fail the build execution

ccojocar:
fixes #212

### DIFF
--- a/cmd/gosec/main.go
+++ b/cmd/gosec/main.go
@@ -95,7 +95,7 @@ var (
 	flagScanVendor = flag.Bool("vendor", false, "Scan the vendor folder")
 
 	// fail by severity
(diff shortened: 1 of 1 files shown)


ccojocar:
This was also my impression, but I think is still useful to have the lower severity issues in the repot. We can add later also such a filter.

PR #228, opened 2018-08-14
Small update to G201 and added ConcatString Function

cschoenduve-splunk:
This is a small update to the sql ruleset. `ConcatString` takes a binary expression and forms a string of all of its literals. In rules/sql.go, the function is used when evaluating format string vulnerabilities if the first argument in `fmt.Sprintf` is a binary expression. eg. `fmt.Sprintf("SELECT *" + " FROM foo" + " WHERE user='%s'", user)`

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -261,19 +261,19 @@ func GetPkgAbsPath(pkgPath string) (string, error) {
 func ConcatString(n *ast.BinaryExpr) (string, bool) {
 	var s string
 	// sub expressions are found in X object, Y object is always last BasicLit
-	if right_operand, ok := n.Y.(*ast.BasicLit); ok {
-		if str, err := GetString(right_operand); err == nil {
+	if rightOperand, ok := n.Y.(*ast.BasicLit); ok {
+		if str, err := GetString(rightOperand); err == nil {
 			s = str + s
 		}
 	} else {
 		return "", false
 	}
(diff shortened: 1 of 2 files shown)


cschoenduve-splunk:
Will fix the lint errors

ccojocar:
LGTM. Please could you rebase against latest master? Thanks

cschoenduve-splunk:
Should be rebased on latest master now

ccojocar:
@cschoenduve-splunk The build runs the latest `gosec` against gosec source code. It seems that your rule generates a false positive:

```
[/home/travis/gopath/src/github.com/securego/gosec/cmd/tlsconfig/tlsconfig.go:130-131] - G201: SQL string formatting (Confidence: HIGH, Severity: MEDIUM)
  > fmt.Sprintf("Could not load the Server Side TLS configuration from Mozilla's website. Check the URL: %s. Error: %v\n",
			TLSConfURL, err)
```

I think the issue is caused by the `i` in the regex, because the text does not contain any SQL keyword but it is still evaluated.

```
regexp.MustCompile("(?i)(SELECT|DELETE|INSERT|UPDATE|INTO|FROM|WHERE)
```

Please could you try to fix it?

cschoenduve-splunk:
My bad. It should be fixed now.

ccojocar:
Awesome! Thanks

PR #232, opened 2018-08-20
Added Fprintf to Rule G201

cschoenduve-splunk:
This is another small update to the SQL ruleset. In certain cases, `fmt.Fprintf` can be used to push a vulnerable SQL statement into a buffer which then gets passed to a SQL query function. I gave rule G201 the ability to identify vulnerable `fmt.Fprintf` calls. If `os.Stdout` or `os.Stderr` are used as the buffer, then we mark it as a false positive. 
Example: `fmt.Fprintf(buff, "SELECT * FROM foo WHERE user='%s', bar)`

Please let me know if I should also add code samples to `testutils/source.go`

### DIFF
--- a/rules/sql.go
+++ b/rules/sql.go
@@ -98,16 +98,35 @@ func NewSQLStrConcat(id string, conf gosec.Config) (gosec.Rule, []ast.Node) {
 
 type sqlStrFormat struct {
 	sqlStatement
-	calls gosec.CallList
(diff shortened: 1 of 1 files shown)


cschoenduve-splunk:
@ccojocar Yeah that definitely seems like a better solution in identifying SQL statements. I will look into it!

PR #233, opened 2018-08-25
update to G304 which adds binary expressions and file joining

cschoenduve-splunk:
This update to rule G304 gives gosec the ability to evaluate two path joining functions as well as binary expressions when inside `ioutil.Readfile` or `os.Open`. Gosec can now flag cases like:
`ioutil.Readfile("/tmp/" + input + "/blob")` and `os.Open(filepath.Join("/tmp/", file))`. To do this, a new function called `FindVarIdentities` was added to helpers.go. The function will return all variable identities found in a binary expression.

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -297,7 +297,6 @@ func FindIdentities(n *ast.BinaryExpr) ([]*ast.Ident, bool) {
 	if len(identities) > 0 {
 		return identities, true
 		// if nil or error, return false
-	} else {
-		return nil, false
 	}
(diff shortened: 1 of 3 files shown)


gcmurphy:
Mind also adding test source code here 

https://github.com/securego/gosec/blob/master/testutils/source.go#L475

cschoenduve-splunk:
Sure! I added 3 new tests.

gcmurphy:
Thanks. 👍

PR #238, opened 2018-09-04
Removed wrapping feature from G107 (for now)

### LINKED ISSUE #237: G107 Incorrect behaviour
First of all, thanks for the library, is great 🙌

### Summary 

After the gosec update that adds G107 I'm seeing an issue for every call to a "Get" function. Most of them are not "net/http" Get calls. 

Example:

```go
package main

import (
	"math/rand"
	"strconv"
)

type Foo struct{}

func (f Foo) Get(foo string) {
}

func main() {
	bar := Foo{}
	baz := strconv.Itoa(rand.Intn(4))
	bar.Get(baz)
}
```

### gosec version

1.1.0

1.10.2

darwin/amd64

### Expected behavior

Only catch net/http url calls

### Actual behavior

Get a warning for any get call

cschoenduve-splunk:
I removed the second check from G107 that looks just at the selectors based on issue 237. From my testing, it yielded just as many false positives as the other rules but I guess my testing wasn't broad enough. I want to add the go-resty library to G107 but I need to play around with `GetCallInfo` a little bit more. @ccojocar Sorry for the confusion.   

fixes #237

### DIFF
--- a/rules/ssrf.go
+++ b/rules/ssrf.go
@@ -40,17 +40,6 @@ func (r *ssrf) Match(n ast.Node, c *gosec.Context) (*gosec.Issue, error) {
 			return gosec.NewIssue(c, n, r.ID(), r.What, r.Severity, r.Confidence), nil
 		}
 	}
-	// Look at the last selector identity for methods matching net/http's
-	if node, ok := n.(*ast.CallExpr); ok {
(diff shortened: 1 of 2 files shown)


ccojocar:
Thanks for your quick fix.

Issue #234, opened 2018-08-28
Go Modules support 


kr:
### Summary 

Now that Go 1.11 is released with module support, more Go code throughout the ecosystem will be written as modules. Currently, gosec is unable to analyze code in modules. It would be great to update it to do so.

If this is something the gosec project is interested in, we (my company) are very interested in contributing a patch.

### Steps to reproduce the behavior

Here's an example. (For a real-world sample module, I followed along with https://dave.cheney.net/2018/07/14/taking-go-modules-for-a-spin on @davecheney's httpstat package, and pushed the changes to https://github.com/kr/httpstat/tree/mod.)

You can check out this module outside your `GOPATH` directory:

```
$ pwd
/Users/kr/demo
$ echo $GOPATH
/Users/kr/lib/go
$ git clone -b mod https://github.com/kr/httpstat.git
$ cd httpstat
$ go test
PASS
ok  	github.com/davecheney/httpstat	0.013s
```

But gosec isn't happy:

```
$ gosec *.go
[gosec] 2018/08/28 16:34:08 including rules: default
[gosec] 2018/08/28 16:34:08 excluding rules: default
2018/08/28 16:34:08 main.go is not within the $GOPATH and cannot be processed
```

### gosec version

`7fd94463edcbf12df2b5579682c003c3bffd6749`

### Go version (output of 'go version')

```
go version go1.11 darwin/amd64
```

### Operating system / Environment

```
$ go env GOOS GOARCH
darwin
amd64
```

### Expected behavior

Gosec analyzes the code and reports possible errors, as usual.

### Actual behavior

Error message.

gcmurphy:
Will look at moving to https://godoc.org/golang.org/x/tools/go/packages to address this issue.

ccojocar:
Yap, package loading was updated in 1.11 https://golang.org/doc/go1.11#gopackages.

MVrachev:
I tried it today and is still not fixed.
I downloaded the latest version of Golang and reinstalled gosec but still I get errors from gosec when I scan code outside the GOPATH.

d0znpp:
```
docker run -it -v /tmp/src/xxx_api/:/go/src/xxx_api securego/gosec ./...
[gosec] 2018/10/05 18:41:33 including rules: default
[gosec] 2018/10/05 18:41:33 excluding rules: default
2018/10/05 18:41:33 ./... is not within the $GOPATH and cannot be processed
```

ricardoatsouza:
[...]

PR #239, opened 2018-09-10
Support Go 1.11

dhui:


MVrachev:
Maybe I make a mistake somewhere but I still don't see the node module support Go 1.11.
Here is an example: 
image

ccojocar:
@MVrachev This is just for `gosec` itself, which now uses Go modules to manage its dependencies. The modules are not yet supported by the analyser.  See this open issue #234.

MVrachev:
I am sorry. I misunderstood. :))

PR #240, opened 2018-09-10
Allow quoted strings to be used to format SQL queries

dhui:


dhui:
@ccojocar 

Sure thing!

I looked at `rules/rules_test.go` but couldn't find an example where a negative test was being performed. e.g. where sample code is being processed by gosec but no errors are found

Could you point me to an example of a negative test and/or help me navigate the test suite?

ccojocar:
>I looked at rules/rules_test.go but couldn't find an example where a negative test was being performed. e.g. where sample code is being processed by gosec but no errors are found
> Could you point me to an example of a negative test and/or help me navigate the test suite?

@dhui for instance you can have a look at `SampleCodeG202` in `source.go` and try to add an additional sample for this use case. There, you find samples which generate warnings as well as samples which doesn't.

dhui:
@ccojocar Thanks for the pointer! I've added a test

dhui:
Looks like tests failed b/c the `pq` package isn't in the Go env.

Is the `install` section of `.travis.yml` the best place to `go get` the `pq` package?

ccojocar:
> Is the install section of .travis.yml the best place to go get the pq package?

Yes, you can add it there.

dhui:
3/4 of the test environment passed. The one that failed had a network error.

dhui:
Should I squash my commits and force push to re-run tests?

PR #247, opened 2018-09-27
Fix false positives for SQL string concatenation with constants from another file

### LINKED ISSUE #169: SQL String concatenation with string constants
Currently the following example SQL query constructs will trigger rule `G202`.

```golang
const allowed = ">=3" // Could be in another file
...
query := `SELECT name FROM  users WHERE status ` + allowed
```

Another example:
```golang
var allowed string
switch role {
case "admin":
    allowed = "<3"
case "user":
    allowed = ">=3"
}
query := `SELECT name FROM  users WHERE status ` + allowed
```

Would it possible/feasible to allow `gas` to ignore such SQL string concatenation constructs as they are inherently safe?

I don't mind trying to do a PR for `gas` to safely ignore such cases.

wongherlung:
Fixes #169. Sorry, it's been a while, was busy with work. 😅 

This PR aims to handle the following case:

main.go
```go
package main

func main() {
  ...
  rows, err := db.Query("SELECT * FROM foo WHERE gender = " + gender)
  ...
}
```

constants.go (for example)
```go
package main

...
const gender = "M"
...

```

Currently, gosec will flag this this out as an issue in `main.go` as the `gender` identifier is unresolved. 

This PR will add `pkg.Files` to `gosec.context` so that we will be able to resolve that identifier if it happens to be in the global scope of that package.

### DIFF
--- a/rules/sql.go
+++ b/rules/sql.go
@@ -55,11 +55,11 @@ func (s *sqlStrConcat) checkObject(n *ast.Ident, c *gosec.Context) bool {
 	if n.Obj != nil {
 		return n.Obj.Kind != ast.Var && n.Obj.Kind != ast.Fun
 	}
+
+	// Try to resolve unresolved identifiers using other files in same package
 	for _, file := range c.PkgFiles {
 		if node, ok := file.Scope.Objects[n.String()]; ok {
-			if node.Kind == ast.Con {
-				return true
-			}
+			return node.Kind == ast.Con
 		}
 	}
 	return false
--- a/rules/sql.go
+++ b/rules/sql.go
@@ -59,7 +59,7 @@ func (s *sqlStrConcat) checkObject(n *ast.Ident, c *gosec.Context) bool {
 	// Try to resolve unresolved identifiers using other files in same package
 	for _, file := range c.PkgFiles {
 		if node, ok := file.Scope.Objects[n.String()]; ok {
-			return node.Kind == ast.Con
+			return node.Kind != ast.Var && node.Kind != ast.Fun
 		}
 	}
 	return false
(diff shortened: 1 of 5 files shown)


rules/sql.go
```diff
@@ -51,10 +51,17 @@ func (s *sqlStrConcat) ID() string {
 }
 
 // see if we can figure out what it is
-func (s *sqlStrConcat) checkObject(n *ast.Ident) bool {
+func (s *sqlStrConcat) checkObject(n *ast.Ident, c *gosec.Context) bool {
 	if n.Obj != nil {
 		return n.Obj.Kind != ast.Var && n.Obj.Kind != ast.Fun
 	}
+
```
ccojocar: It would be nice to have a test for this use case.

wongherlung: 👍 I was thinking so too. But this test case spans across two files and I'm not sure how to do it.

ccojocar: I think you can extend the `sample.Code` to support multiple files. The test package is built here: https://github.com/securego/gosec/blob/5f98926a7be81fd2fcb25b535e2b4e30406a8af0/rules/rules_test.go#L35

In this file are the code samples defined:
https://github.com/securego/gosec/blob/master/testutils/source.go

ccojocar: Looks fine. You could have also use a `Code []string`. To keep multiple files. I will refactoring afterwards.

Thanks for providing the tests.

wongherlung: ahh yes. It's much better to use `Code []string`. If you are okay, can I help with the refactoring? for https://hacktoberfest.digitalocean.com/ 😂

ccojocar: I refactored it in the PR https://github.com/securego/gosec/pull/248

PR #261, opened 2018-10-29
Whitelist strings builder

### LINKED ISSUE #255: G104: Whitelist string.Builder Write methods
### Summary 

Go 1.10 added [Builder](https://golang.org/pkg/strings/#Builder) to the `strings` package for efficiently building strings. It implements `Write`, `WriteByte`, `WriteRune`, and `WriteString` which all always return nil. Add them to the G104 whitelist to ignore unchecked errors.

```go
var buf strings.Builder
buf.WriteString("test string")
```

### gosec version

master (e0a150b)

go version go1.11 linux/amd64

Debian buster

### Expected behavior

No matching rules.

### Actual behavior

```
G104: Errors unhandled. (Confidence: HIGH, Severity: LOW)
  > buf.WriteString("test string")
```

endorama:
Hello, first of all thank you for this project!

I would like to contribute, and I've started addressing issue #255.

I added a test case and the implementation, but the test are failing, and I can't get why. I tried reverse engineering the behaviour from other tests but it's clearly not working as expected.

Any help would be greatly appreciated :)

Thank you!

Fixes #255

### DIFF
--- a/rules/errors.go
+++ b/rules/errors.go
@@ -81,6 +81,7 @@ func NewNoErrorCheck(id string, conf gosec.Config) (gosec.Rule, []ast.Node) {
 	whitelist := gosec.NewCallList()
 	whitelist.AddAll("bytes.Buffer", "Write", "WriteByte", "WriteRune", "WriteString")
 	whitelist.AddAll("fmt", "Print", "Printf", "Println", "Fprint", "Fprintf", "Fprintln")
+	whitelist.AddAll("strings.Builder", "Write", "WriteByte", "WriteRune", "WriteString")
 	whitelist.Add("io.PipeWriter", "CloseWithError")
 
 	if configured, ok := conf["G104"]; ok {
--- a/testutils/source.go
+++ b/testutils/source.go
@@ -165,6 +165,8 @@ func main() {
 	e := test()
 	fmt.Println(e)
 }`}, 0}, {[]string{`
+// +build go1.10
+
 package main
 import "strings"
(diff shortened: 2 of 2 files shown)


rules/errors.go
```diff
@@ -81,6 +81,7 @@ func NewNoErrorCheck(id string, conf gosec.Config) (gosec.Rule, []ast.Node) {
 	whitelist := gosec.NewCallList()
 	whitelist.AddAll("bytes.Buffer", "Write", "WriteByte", "WriteRune", "WriteString")
 	whitelist.AddAll("fmt", "Print", "Printf", "Println", "Fprint", "Fprintf", "Fprintln")
+	whitelist.AddAll("strings.Builder", "Write", "WriteByte", "WriteRune", "WriteString")
```
ccojocar: I think that we have to make this backward compatible with Go 1.9 since these functions were introduced only in Go 1.10. We are trying to keep compatibility with last 3 major versions.

ccojocar:
@endorama Please could you rebase?

rules/errors.go
endorama: How can this be achieved in this case? As this is a string-based whitelist, why is not backward compatible? From my understanding this is not compile code but just strings used for string-matching when comparing AST with the whitelist.

endorama:
@ccojocar now that I can see the error seems so obvious :sweat_smile: 

Anyway thank you for merging this :+1:

PR #262, opened 2018-11-04
Make G201 ignore CallExpr with no args

mururu:


mururu:
The CI should fixed by #263.

PR #275, opened 2019-01-14
Update the README with some details about the configuration file

### LINKED ISSUE #269: Documentation about how to use config file
### Summary 
I want to set up a config file for Gosec. I see it is possible because there is a flag: "-conf string : Path to optional config file" but it is unclear for me what should be the syntax of that file.
Can you give an example? 

I want ot have the ability to set some of those options (it will be cool all to support all of them): 
https://github.com/PyCQA/bandit#per-project-command-line-args

So my question is can you documented it somewhere? 

### Expected behavior
Easy to understand with an example documentation on how to set up custom config file.

### Actual behavior
I don't find anywhere a documentation about it.

ccojocar:
fixes #269

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ A number of global settings can be provided in a configuration file as follows:
 ```
 
 - `nosec`: this setting will overwrite all `#nosec` directives defined throughout the code base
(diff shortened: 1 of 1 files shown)


ccojocar:
@gcmurphy Please could you have a look over this PR? Thanks

PR #284, opened 2019-02-25
Report for Golang errors

MVrachev:
Right now if you use Gosec to scan invalid go file and if you report the result in a text, JSON, CSV or another file format you will always receive 0 issues.
The reason for that is that Gosec can't parse the AST of invalid go files and thus will not report anything.

The real problem here is that the user will never know about the issue if he generates the output in a file.

I fix this problem in this patch and I added those Golang errors in the output.
Now when there are Golang errors found Gosec will exit with code 1.

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -64,6 +64,7 @@ type Analyzer struct {
 	logger      *log.Logger
 	issues      []*Issue
 	stats       *Metrics
+	errors      map[string][]Error
 }
 
 // NewAnalyzer builds a new analyzer.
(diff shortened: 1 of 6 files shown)


MVrachev:
Here is how the output when there are Golang issues + other vulnerable looks:

text:
image

JSON:
image

happypork:
Great idea! But I wonder I get a bunch of errors from standard library files like: 
/usr/local/go/src/runtime/mem_linux.go:21:6: sysAlloc redeclared in this block
/usr/local/go/src/runtime/mem_aix.go:14:6: 	other declaration of sysAlloc

... and many more lying in GOROOT.

Do I have to exclude them?

MVrachev:
Hi, do you mean those errors in the red box:
image

or do you mean that you want to ignore our report for Golang errors as shown here in the green box: 
image

happypork:
Red box :)

MVrachev:
Yes, you are right. That's something annoying from the golang parser library.
There is no solution until now how to silence those errors.

I will try to look at this as soon as I can.

Issue #287, opened 2019-03-11
Add Sonarqube Generic-Issue as format


kmcrawford:
### Summary 
I would like to add to the `-fmt` a `sonarqube` option that would export the data in a sonarqube format.

https://docs.sonarqube.org/latest/analysis/generic-issue/

gcmurphy:
This was addressed via #288

PR #288, opened 2019-03-11
Add sonarqube output

kmcrawford:
For Issue #287

### DIFF
--- a/cmd/gosec/main.go
+++ b/cmd/gosec/main.go
@@ -65,7 +65,7 @@ var (
 	flagIgnoreNoSec = flag.Bool("nosec", false, "Ignores #nosec comments when set")
 
 	// format output
-	flagFormat = flag.String("fmt", "text", "Set output format. Valid options are: json, yaml, csv, junit-xml, html, or text")
+	flagFormat = flag.String("fmt", "text", "Set output format. Valid options are: json, yaml, csv, junit-xml, html, sonarqube, or text")
 
 	// output file
 	flagOutput = flag.String("out", "", "Set output file for results")
@@ -165,19 +165,19 @@ func loadRules(include, exclude string) rules.RuleList {
 	return rules.Generate(filters...)
 }
 
-func saveOutput(filename, format string, issues []*gosec.Issue, metrics *gosec.Metrics, errors map[string][]gosec.Error) error {
+func saveOutput(filename, format, rootPath string, issues []*gosec.Issue, metrics *gosec.Metrics, errors map[string][]gosec.Error) error {
 	if filename != "" {
 		outfile, err := os.Create(filename)
 		if err != nil {
 			return err
 		}
 		defer outfile.Close()
-		err = output.CreateReport(outfile, format, issues, metrics, errors)
+		err = output.CreateReport(outfile, format, rootPath, issues, metrics, errors)
 		if err != nil {
 			return err
 		}
 	} else {
-		err := output.CreateReport(os.Stdout, format, issues, metrics, errors)
+		err := output.CreateReport(os.Stdout, format, rootPath, issues, metrics, errors)
 		if err != nil {
 			return err
 		}
@@ -318,6 +318,7 @@ func main() {
 	if *flagBuildTags != "" {
 		buildTags = strings.Split(*flagBuildTags, ",")
 	}
+
 	if err := analyzer.Process(buildTags, packages...); err != nil {
 		logger.Fatal(err)
 	}
@@ -342,9 +343,9 @@ func main() {
 	if !issuesFound && *flagQuiet {
 		os.Exit(0)
 	}
-
+	rootPath := packages[0]
 	// Create output report
-	if err := saveOutput(*flagOutput, *flagFormat, issues, metrics, errors); err != nil {
+	if err := saveOutput(*flagOutput, *flagFormat, rootPath, issues, metrics, errors); err != nil {
 		logger.Fatal(err)
 	}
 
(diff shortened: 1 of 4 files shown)


output/sonarqube_format.go
```diff
@@ -0,0 +1,36 @@
...
+	EtartColumn int `json:"endColumn,omitempty"`
+}
+type location struct {
+	Message   string    `json:"message"`
+	FilePath  string    `json:"filePath"`
+	TextRange textRange `json:"textRange,omitempty"`
+}
+
+type sonarIssue struct {
+	EngineId           string     `json:"engineId"`
```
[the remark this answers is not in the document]
kmcrawford: updated

kmcrawford:
travis keeps failing for silly go get network issues.

gcmurphy:
The latest travis-ci run fails on the tip: https://travis-ci.org/securego/gosec/jobs/504930446. `gosec `runs against itself, and on the tip it seems like a package is missing (which is triggering failure):

`could not import golang.org/x/sys/cpu (cannot find package "golang.org/x/sys/cpu" in any of..`

Likely introduced by https://github.com/golang/go/issues/24843

So I'm ok with merging this with the ci job failure.

cc/ @MVrachev

cmd/gosec/main.go
```diff
@@ -165,19 +165,19 @@ func loadRules(include, exclude string) rules.RuleList {
 	return rules.Generate(filters...)
 }
 
-func saveOutput(filename, format string, issues []*gosec.Issue, metrics *gosec.Metrics, errors map[string][]gosec.Error) error {
+func saveOutput(filename, format, rootPath string, issues []*gosec.Issue, metrics *gosec.Metrics, errors map[string][]gosec.Error) error {
```
gcmurphy: I'm not sure modifying the interface to achieve a truncated file path is the best solution here. Is there any specific reason for it? Sonarqube only recognize relative paths?

This is certainly something we can look at fixing though.

output/formatter.go
```diff
@@ -91,12 +93,50 @@ func CreateReport(w io.Writer, format string, issues []*gosec.Issue, metrics *go
...
+			EngineID: "gosec",
+			RuleID:   issue.RuleID,
+			PrimaryLocation: location{
+				Message:   issue.What,
+				FilePath:  strings.Replace(issue.File, rootPath+"/", "", 1),
+				TextRange: textRange{StartLine: startLine, EndLine: endLine},
+			},
+			Type:          "VULNERABILITY",
+			Severity:      getSonarSeverity(issue.Severity.String()),
+			EffortMinutes: 5,
```
gcmurphy: Why hardcode a value here? We could potentially actually capture time elapsed here as part of the reporting process. (start_time, finish_time) to the stats area.

output/formatter.go
```diff
@@ -91,12 +93,50 @@ func CreateReport(w io.Writer, format string, issues []*gosec.Issue, metrics *go
...
+		}
+		si.SonarIssues = append(si.SonarIssues, s)
+	}
+	raw, err := json.MarshalIndent(si, "", "\t")
+	if err != nil {
+		panic(err)
```
gcmurphy: panic? could you return the error instead?

output/formatter.go
```diff
@@ -91,12 +93,50 @@ func CreateReport(w io.Writer, format string, issues []*gosec.Issue, metrics *go
...
+	}
+
+	_, err = w.Write(raw)
+	if err != nil {
+		panic(err)
```
gcmurphy: same here..

output/formatter.go
kmcrawford: No reason, I just didn’t want to change your project too much when I actually have no idea how long it will take to fix.

output/formatter.go
kmcrawford: That’s a copy and paste from the json formatter.

output/formatter.go
kmcrawford: Same here.

cmd/gosec/main.go
kmcrawford: Yes, sonarqube reports from the root of the project. I needed a way to remove the full path.  Happy to refactor, if you can think of a better way of doing this.

kmcrawford: Example of this working:
image

output/formatter.go
kmcrawford: image

ccojocar: My understanding is that this is an estimation for developers which indicates how long it would take to fix the issue. In this case, I don't see how we can measure it, therefore I am fine with a constant. Maybe in future, it can be defined per issue for a more granular estimation.

I would extract this in a constant with a clear meaning.

kmcrawford: done.

output/formatter.go
gcmurphy: I think we should probably return the error. I'll submit an issue to fix the json formatter and do so after this has been merged.

output/formatter.go
```diff
@@ -41,6 +43,9 @@ const (
 
 	// ReportJUnitXML set the output format to junit xml
 	ReportJUnitXML // JUnit XML format
+
+	//Sonarqube effort to fix
```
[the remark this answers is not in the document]
kmcrawford: getting picky, done.

output/formatter.go
kmcrawford: error returned and fixed the json formatter.

kmcrawford:
All open requests have been updated.

kmcrawford:
bump

Issue #304, opened 2019-04-26
Failure when running in a directory without Go packages, but with subdirectories with Go packages


jsullivan3:
### Summary 

Since commit https://github.com/securego/gosec/commit/85d180848dbb8f86cba5d71bbbe06e1b40ce2705, following invocation in a directory that contains no Go packages but contains subdirectories with Go packages fails:

`gosec -vendor -exclude=G104 ./...`

with the following error:

`[gosec] 2019/04/26 11:02:42 strconv.Atoi: parsing " no Go files in /home/jsullivan3/path/to/my/project": invalid syntax`

### Steps to reproduce the behavior

```
$ go get -u github.com/securego/gosec/cmd/gosec
$ go install github.com/securego/gosec/cmd/gosec/...
$ gosec -vendor -exclude=G104 ./...
```

### gosec version

Build from commit `85d180848dbb8f86cba5d71bbbe06e1b40ce2705`.

### Go version (output of 'go version')

`go version go1.12.4 linux/amd64`

### Operating system / Environment

RHEL 7.6 x86_64

### Expected behavior

`gosec` analyzes all packages within subdirectories of the current directory, even if the current directory contains no Go packages.

### Actual behavior

`gosec` searches all subdirectories of the current directory, then complains that no Go packages are found in the current directory.

ccojocar:
@jsullivan3 I am closing this since this issue seems to be fixed after refactoring. Please feel free to reopen it if it does not work as expected. Thanks!

jeffdc:
This does not seem to be fixed as part of the latest refactoring.

`go get-u github.com/securego/gosec/cmd/gosec`

and then run `gosec` on a directory without a go file in the root and this failure will still happen.

`[gosec] 2019/04/29 15:59:17 strconv.Atoi: parsing " no Go files in /path/to/my/project": invalid syntax`

sanmarco:
@ccojocar I'm getting the same error after updating to the latest master. I built a simple case that hopefully helps reproducing it:

Create a folder with this structure:

```
- gosec-issue-304
  |- cmd
     |- main.go
  |- tests
     | - my_test.go
```

main.go:
```go
package main

import (
	"fmt"
)

func main() {
	fmt.Println("main")
}
```

my_test.go:
```go
package tests

import (
	"testing"
)

func TestMy(t *testing.T) {
}
```

From the folder `gosec-issue-304` run `go ./...`
It should produce the error:
[...]

PR #306, opened 2019-04-27
Fix failure when a folder does not contain go files

sanmarco:
Fix for issue #304

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -126,13 +126,14 @@ func (gosec *Analyzer) Process(buildTags []string, packagePaths ...string) error
 			packageFiles = append(packageFiles, path.Join(packagePath, filename))
 		}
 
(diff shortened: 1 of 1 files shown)


ccojocar:
@sanmarco Thanks for this PR! We are refactoring a bit the analyser in this PR #307, and it think this is conflicting with those changes. 

I believe that #301 is addressing this issue. Let's keep this PR a bit on hold until the refactoring gets merged.

sanmarco:
@ccojocar I re-tested after the merge of #301 but I'm still having the issue reported in #304. Will #304 be handled in #307? As a side note, I noticed that test files (*_test,go) are not checked but I believe this preexisted #301. Thanks.

ccojocar:
Yes. I have just tested with the master version and it seems to be fixed. Thanks for sending a fix anyhow!

PR #317, opened 2019-05-14
print version string

szuecs:
We validate that a proper installation was done by printing version string and exit with code 0.
Signed-off-by: Sandor Szücs <sandor.szuecs@zalando.de>


szuecs:
Go 1.11 and Go1.12 are fine only tip has some weird errors.

szuecs:
Anything that blocks from merging this PR? @ccojocar

ccojocar:
I think we can merge this PR since the unit test failure is not related. It needs a separate investigation.

Issue #338, opened 2019-07-16
Rule G204 too restrictive


MVrachev:
### Summary 
Rule G204 seems to be too restrictive. It seems to me that it just annotates every function used for external processes.

In issues [#292](https://github.com/securego/gosec/issues/292) and [#204](https://github.com/securego/gosec/issues/204) we can see the frustration of the users.
Of course, a big part of the frustration for them is that currently there is no good documentation for the rule.

### gosec version
2.0.0
### Go version (output of 'go version')
go version go1.11 darwin/amd64
### Operating system / Environment
mac OS Mojave
### Expected behavior
Command injection suggests that the program starts a new process with input provided by the user. This, of course,  should be annotated.
Here is a good example provided for that from `gosec/testutils/source.go` 
```
package main
import (
	"log"
	"os"
	"os/exec"
)
func main() {
	run := "sleep" + os.Getenv("SOMETHING")
	cmd := exec.Command(run, "5")
	err := cmd.Start()
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Waiting for command to finish...")
	err = cmd.Wait()
	log.Printf("Command finished with error: %v", err)
}
```
### Actual behavior
Here are two examples which are taken from the `gosec/testutils/source.go` file which is showcasing this restrictive behavior:
1) 
```
package main
import "syscall"
func main() {
	syscall.Exec("/bin/cat", []string{ "/etc/passwd" }, nil)
}
```

2) 
```
package main
import (
	"log"
	"os/exec"
)
func main() {
	cmd := exec.Command("sleep", "5")
	err := cmd.Start()
 	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Waiting for command to finish...")
  	err = cmd.Wait()
  	log.Printf("Command finished with error: %v", err)
}
```

MVrachev:
I found a comment in `rules/subproc.go` which suggests that change should be made:
![image](https://user-images.githubusercontent.com/16246778/61287156-f1003380-a7cc-11e9-8025-0b1e79ab4ab2.png)

PR #339, opened 2019-07-30
Make g204 less restrictive

MVrachev:
Change rule G204 to be less restrictive

Currently, rule G204 warns you about every single use of the
functions syscall.Exec, os.exec.CommandContext and os.Exec.Command.
This can create false positives and it's not accurate because you can
use those functions with perfectly secure arguments like hardcoded
strings for example.

With this change, G204 will warn you in 3 cases when passing arguments
to a function which starts a new process the arguments:
1) are variables initialized by calling another function
2) are functions
3) are command-line arguments or environmental variables

Closes: https://github.com/securego/gosec/issues/338


testutils/source.go
```diff
@@ -679,7 +678,40 @@ func main() {
...
+	"syscall"
+)
+func main() {
+	err := syscall.Exec("/bin/cat", []string{"/etc/passwd"}, nil)
+	if err != nil {
+		fmt.Printf("Error: %v\n", err)
+	}
+}`}, 0, gosec.NewConfig()},
+		{[]string{`
+// starting a process with a variable as an argument
```
ccojocar: Please could you add one more test for the case when a variable is a function argument. Something like:

```Go
func RunCmd(args ...strings) {
     cmd := exec.Command("cmd", args...)
     err := cmd.Start()
     if err != nil {
          log.Fatal(err)
     }
}

fun main() {
      RunCmd(os.Args...)
}
```

```

Comment (MVrachev) [reply testutils/source.go]: While I was working on this new test I noticed this issue:
https://github.com/securego/gosec/issues/343

Comment (ccojocar) [reply testutils/source.go]: Yeah, this needs more advanced data flow analysis in order to reliably solve it. It should recourse down the graph to find the origin of the initialisation.

I think for now, it's fine if there is a warning in this case and we let the user to investigate whether or not is a vulnerability.

Comment (MVrachev) [reply testutils/source.go]: So, do I need other unit tests or something?

Comment (ccojocar) [reply testutils/source.go]: It would be nice to handle this case in the rule and return a warning. Also it would be great to have a unit test for this which checks that in this case the rule returns a warning.

Comment (ccojocar) [reply testutils/source.go]: @MVrachev Are you still working on this? I would love to get this fix merged. I hope we can do a release soon, and it would be nice to include also this one. Thanks!

Comment (MVrachev) [reply testutils/source.go]: Yes, I haven't abandoned it. It's just I have to finish other work.
I will do my best to finish it this week. 
Sorry for the inconvenience.

Comment (ccojocar) [reply testutils/source.go]: Excellent!

Comment (MVrachev) [reply testutils/source.go]: For this code:

```
package main

import (
	"log"
	"os/exec"
)

func RunCmd(command string) {
	cmd := exec.Command(command, "5")
	err := cmd.Start()
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Waiting for command to finish...")
	err = cmd.Wait()
}

func main() {
	RunCmd("sleep")
}
```
Gosec will return the following warning: `G204: Subprocess launched with variable (Confidence: HIGH, Severity: MEDIUM)
  > exec.Command(command, "5")`

What you want me to do is just to add a unit test for this instance?

ccojocar: Yeah, it would be great to add this code sample to the list of samples tested on G204 here https://github.com/securego/gosec/blob/98749b7357954795b616fc335c98fb7bcb144576/testutils/source.go#L616. Thanks!

MVrachev: Done. I updated my pr with the new unit test.

PR #352, opened 2019-09-09
Remove rule G105 which detects the use of math/bit.Int.Exp

### LINKED ISSUE #283: G105 issue cannot be fixed?
### Summary 
I get an error G105 - Use of math/big.Int.Exp function should be audited for modulus == 0 (Confidence: HIGH, Severity: LOW) but I can't see any way to fix it?

code sample:
```
import "math/big"

lower := new(big.Int)
lower.Exp(big.NewInt(2), big.NewInt(255), nil)
```

I would expect no error as modulas is hard coded to `nil` which means ignored

### gosec version
1.2.0
go version go1.11.2 linux/amd64
ubuntu 18
### Expected behavior
Have a way to get rid of this error, not sure if its even valid as setting m==nil explictitly ignores m according to the exported func
### Actual behavior
no way to get rid of warning

ccojocar:
This function used to have a vulnerability in the older versions of Go, which
is had been fixed (https://github.com/golang/go/issues/15184).

Also reorder the remaining rules.

fixes #283


ccojocar:
Agree. I had the same thoughts on this. It is somehow a breaking change. I'll keep the IDs and just add a note in the README.

ccojocar:
Closing this in favour of https://github.com/securego/gosec/pull/352.

PR #353, opened 2019-09-09
Add support to exclude arbitrary folders from scanning

### LINKED ISSUE #342: Ignore abitrary directory
### Summary 
By default gosec ignores files in the 'tests' directory. However my project uses the name 'test' to store tests. It would be nice if we could specify arbitrary directory name to be ignored.

Scan a project recursively with a 'test' folder.

### gosec version
commit 4b59c948083cd711b6a8aac8f32721b164899f57

go version go1.12.7 linux/amd64

ccojocar:
fixes #342

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,13 +137,21 @@ of functions which will be skipped when auditing the not checked errors:
 gosec will fetch automatically the dependencies of the code which is being analyzed when go modules are turned on (e.g.` GO111MODULE=on`). If this is not the case,
 the dependencies need to be explicitly downloaded by running the `go get -d` command before the scan.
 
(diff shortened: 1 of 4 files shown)


ccojocar:
This is a simple approach to exclude an entire folder. A path is excluded from scanning if one segment of the path matches the regex.

We can improve it in case there will be confusion.

gcmurphy:
Ok I think that's better. Previously I think we tried matching whole regex.

arangamani:
Is there an ETA on seeing this change on a releasee tag?

ccojocar:
I hope we will cut soon a release.

PR #366, opened 2019-09-11
Expose analyzer API

jirfag:
Make it possible to use gosec from e.g. golangci-lint
without modification of gosec.


jirfag:
thank you!

PR #377, opened 2019-09-17
Remove redundant unit test for rule G304

MVrachev:
There are two identical unit tests for rule G304.

The first and the third one in `testutils/source.go`

### DIFF
--- a/testutils/source.go
+++ b/testutils/source.go
@@ -795,22 +795,6 @@ func main() {
 	})
 	log.Fatal(http.ListenAndServe(":3000", nil))
 }`}, 1, gosec.NewConfig()}, {[]string{`
-package main
-
-import (
-	"log"
-	"os"
-	"io/ioutil"
-)
-
-	func main() {
-		f2 := os.Getenv("tainted_file2")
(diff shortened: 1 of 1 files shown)


testutils/source.go
```diff
@@ -795,22 +795,6 @@ func main() {
 	})
 	log.Fatal(http.ListenAndServe(":3000", nil))
 }`}, 1, gosec.NewConfig()}, {[]string{`
-package main
```
ccojocar: Do you mean this is a duplicate of:

```Go
package main
import (
"os"
"io/ioutil"
"log"
)
func main() {
f := os.Getenv("tainted_file")
body, err := ioutil.ReadFile(f)
if err != nil {
 log.Printf("Error: %v\n", err)
}
log.Print(body)
```

I think the `ReadFile` has a string concatenation in the path: `"/tmp/" + f2`.

MVrachev: Yes, you are right.

My mistake.

MVrachev:
There is no duplication.
The third unit tests concatenate `"/tmp/" + f2` as an argument to the ReadFile function instead of just `taintInput` as it's in the first unit test.

PR #381, opened 2019-09-18
Change unit tests to check for one thing

MVrachev:
The unit tests should check for a single thing at a time.
This was not true for some the tests.

### DIFF
--- a/testutils/source.go
+++ b/testutils/source.go
@@ -771,36 +771,123 @@ func main() {
 	// SampleCodeG301 - mkdir permission check
 	SampleCodeG301 = []CodeSample{{[]string{`
 package main
-import "os"
+
+import (
+	"fmt"
+	"os"
+)
+
 func main() {
-	os.Mkdir("/tmp/mydir", 0777)
-	os.Mkdir("/tmp/mydir", 0600)
-	os.MkdirAll("/tmp/mydir/mysubidr", 0775)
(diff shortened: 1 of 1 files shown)


MVrachev:
> I don't quite get what's the goal of this change. In my opinion, it is fine to have multiple warnings per check. I don't see where do you cover the tests which you removed.

There are a few things.
I will demonstrate my points with the unit test for G301:
- in the test `os.Mkdir("/tmp/mydir", 0777)` is used which has an error and `os.Mkdir("/tmp/mydir", 0600)` which hasn't. Those should be two seprated tests. 
- when the `os.Mkdir` and `os.MkdirAll` are used they are not checked for returned errors and this leads to other G104 errors
- `os.Mkdir` and `os.MkdirAll` are two different instances of the same rule. They should be checked separately.

So, in summary, I think that one who creates a unit test for Gosec should:
- make a unit test as simple as possible without simulating the same error twice in a unit test but instead split the test in two if he wants to check two different instances of a rule
- use a unit tests to do only one of the two things: check for an error or check that Gosec doesn't report any errors
- make sure that Gosec will catch errors only related to the rule which is tested

MVrachev:
>  I don't see where do you cover the tests which you removed.

For G301: all three statements are checked in three different unit tests.

For  G302: all four statements are checked in three different unit tests.

For G303: I just removed the file opening because it was redundant.

PR #399, opened 2019-10-01
Convert the global settings to correct type when reading them from file

### LINKED ISSUE #394: Global Config File is not respected
### Summary 
When attempting to use a config file, the config file options are not actually being used.

Sample code
```
package main

import (
	"fmt"
)

func main() {
	/* #nosec */
	password := "hhend77dyyydbh&^psNSSZ)JSM--_%"
	fmt.Println("hello, from the vulnerable app" + password)
}

```
Note the password above is not real and a sample to trigger the gosec scanner.

Creating a config.json with the settings described below does not actually disable the nosec comments.

### gosec version
Head of master

`go version go1.12.4 darwin/amd64`

OSX

### Expected behavior
That either 
```
{
    "global": {
        "nosec": "<options>",
    }
}```
where options is true, false, enable, enabled, disable, or disabled would result in disabling #nosec comments. 

### Actual behavior
None of the above options actually does anything. The scanner still respects the #nosec comments. (Note the command line flag to disable nosec is respected though.)

ccojocar:
fixes #394

### DIFF
--- a/config.go
+++ b/config.go
@@ -38,6 +38,22 @@ func NewConfig() Config {
 	return cfg
 }
 
+func (c Config) keyToGlobalOptions(key string) GlobalOption {
+	return GlobalOption(key)
+}
+
+func (c Config) convertGlobals() {
+	if globals, ok := c[Globals]; ok {
+		if settings, ok := globals.(map[string]interface{}); ok {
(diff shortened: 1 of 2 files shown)


ccojocar:
@gcmurphy Please can you have a look also to this one?

PR #400, opened 2019-10-01
Replace the deprecated load mode with more specific flags are recommended in the packages docs

ccojocar:


ccojocar:
It is available in all Go versions supported by gosec. It is defined anyhow in the `packages` package.

PR #405, opened 2019-10-20
Julian add cwe rulemappings

### LINKED ISSUE #368: Map gosec rules to CWEs
### Summary 

This is more of a feature request than an actual issue. 

At the moment, `gosec` scans and reports security issues based on the rules listed [here](https://github.com/securego/gosec/blob/29fddff3b488cfa854fa94cf4e277ac5aa136001/rules/rulelist.go#L59). 

However, in order to use a more standard way of reporting vulnerabilities, it would be great to link these rules to CWEs. I would be happy to work on a patch. I am just wondering whether you think that adding this information would be useful.

I was thinking about adding the CWE mapping information right into the `RuleList` in [this function](https://github.com/securego/gosec/blob/29fddff3b488cfa854fa94cf4e277ac5aa136001/rules/rulelist.go#L59)

Kindly let me know your opinion.

julianthome:
Fixes #368

Add CWE rule mappings to gosec rules.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -34,10 +34,50 @@ const (
 	High
 )
 
+// Cwe id and url
+type Cwe struct {
+	ID  string
+	URL string
+}
+
+// GetCwe creates a cwe object for a given RuleID
+func GetCwe(id string) Cwe {
+	return Cwe{ID: id, URL: fmt.Sprintf("https://cwe.mitre.org/data/definitions/%s.html", id)}
+}
+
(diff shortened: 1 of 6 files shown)


julianthome:
> Thanks for this contribution! It looks great. I am just wondering if you can add a unit test in the `formatter_test.go` to validate the mapping. Thanks

Yes sure! Will add some unit tests.

PR #412, opened 2019-11-08
cmd/tlsconfig: remove support for deprecated tls.VersionSSL30

alrs:
The Go authors deprecated support for SSLv3 in go 1.13. They've decided to remove it in go 1.14, and have already merged the commit. 

See: https://golang.org/issue/32716

This PR, or something like it, should probably land before February 2020.

### DIFF
--- a/cmd/tlsconfig/tlsconfig.go
+++ b/cmd/tlsconfig/tlsconfig.go
@@ -4,7 +4,6 @@ package main
 
 import (
 	"bytes"
-	"crypto/tls"
 	"encoding/json"
 	"errors"
 	"flag"
@@ -14,7 +13,6 @@ import (
 	"log"
 	"net/http"
 	"path/filepath"
-	"sort"
 	"strings"
 
 	"github.com/mozilla/tls-observatory/constants"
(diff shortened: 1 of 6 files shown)


cmd/tlsconfig/tlsconfig.go
```diff
@@ -124,8 +124,6 @@ func mapTLSVersions(tlsVersions []string) []int {
 			versions = append(versions, tls.VersionTLS11)
 		case "TLSv1":
 			versions = append(versions, tls.VersionTLS10)
-		case "SSLv3":
-			versions = append(versions, tls.VersionSSL30)
```
ccojocar: This is used to map the TLS version published by Mozilla (https://statics.tls.security.mozilla.org/server-side-tls-conf.json) to Go version  when generating the TLS rule. 

We cannot directly remove it since we support 2 major older Go versions from latest.  Most likely, we have to make it conditional with a build tag  when it is entirely removed form newer Go versions.

alrs: I reworked the PR to use build tags, with  some comments in the source to help the future know when to discard the older of the `mapTLSVersions()`.

ccojocar: Looks good! Thanks!

ccojocar:
@alrs Please can you fix the build? Thanks!

alrs:
My understanding of build tags is that `go1.13` means "go 1.13 and later", and `!go1.12` means "not go 1.12 or later."

https://golang.org/pkg/go/build/

ccojocar:
Thanks! You are right,  I'm looking now at the build tags for go version compatibility. Just a small nit, you have go1.14 twice  in `tls_version_go12_go13.go`. Please can you remove one tag? Thanks

alrs:
> Please can you remove one tag?

Done.

PR #422, opened 2019-12-24
Add a rule which detects when there is potential integer overflow

rung:
### What
- Add rule to detect potential integer overflow 
  * (this is my first PR for gosec.)

### Why
- strconv.Atoi may cause Integer Overflow. But gosec doesn't check this. So I would like to add this rule.

- Details
  - In a third-party security audit of the Kubernetes project, many potential integer overflow bugs were found. So I think this bug is common and should be detected by gosec.
  - Kubernetes Audit Report
    - P93 ("B. strconv.Atoi result conversion may cause integer overflows")
    - https://github.com/kubernetes/community/blob/master/wg-security-audit/findings/Kubernetes%20Final%20Report.pdf
  - As written in the report, developer should use strconv.ParseInt or strconv.ParseUint to parse string

### Special Note for reviewers
- This rule checks code by every ast.FuncDecl (by Function)
  - This rule has to check multiple expressions in the same function because strconv.Atoi is a very common function, so it's not good just to detect this func. To reduce false positives, my code checks int32/int16 conversion too.
  ~- Match() func can't return multiple issues, so this rule detects one issue only in the same function. This is a limitation to reduce false positives.~
  - When there are other ideas, I'm very happy to modify Match().

### DIFF
--- a/rules/integer_overflow.go
+++ b/rules/integer_overflow.go
@@ -1,0 +1,87 @@
+// (c) Copyright 2016 Hewlett Packard Enterprise Development LP
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package rules
+
+import (
+	"github.com/securego/gosec"
+	"go/ast"
+)
+
+type integerOverflowCheck struct {
+	gosec.MetaData
+	calls gosec.CallList
+}
+
+func (i *integerOverflowCheck) ID() string {
+	return i.MetaData.ID
+}
+
+func (i *integerOverflowCheck) Match(node ast.Node, ctx *gosec.Context) (*gosec.Issue, error) {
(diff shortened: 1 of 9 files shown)


rules/integer_overflow.go
```diff
@@ -0,0 +1,87 @@
...
+						if n, ok := varName[idt.Name]; ok {
+							// Detect int32(v) and int16(v)
+							issue = gosec.NewIssue(ctx, n, i.ID(), i.What, i.Severity, i.Confidence)
+							return false
+						}
+					}
+				}
+			}
+		}
+		return true
```
ccojocar: I am wondering how efficient is this since this will make the `ast.Insepct` to traverse recursively all children of every node.

rules/integer_overflow.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	ast.Inspect(node, match)
```
ccojocar: I think that it is possible to leverage the `gosec.Context` which keeps the global state during analysis instead of calling again the `Inspect` on the entire tree.

rung: @ccojocar Hi! Thank you for your review.
I think your idea is more efficient.

I'll consider how to know whether `strconv.Atoi` and `int32()` are same function using gosec.Context.

rung:
@ccojocar 
Hi! Thank you for your review.

I modified this rule to use Context. (The idea made the rule more efficient! Thank you.)
https://github.com/securego/gosec/pull/422/commits/809e0a6c0fff81380ea18a6c305399851f155103

Could you check my code again? I would like to know wether we are getting closer.

rung:
@ccojocar Thank you!! I'm very happy to work for gosec 👍

PR #426, opened 2019-12-27
Check for both default and alternative nosec tags

### LINKED ISSUE #425: Check for both tags when handling false positives
### Summary 

Some repositories that are using [huskyCI](http://husky.ci) and already have the `#nosec` tag are having some false positive issues. What happens is that [huskyCI hardcodes](https://github.com/globocom/huskyCI/blob/master/api/config.yaml#L67) the `-nosec-tag` flag and the default one is not being checked anymore.

```sh
gosec -nosec-tag dontanalyze ./...
```

```go
import "md5" // #dontanalyze


func main(){

    /* #dontanalyze */
    if x > y {
        h := md5.New() // this will be ignored
    }

    returnError() // #nosec 
    // this will fail as the tag #dontanalyze is set
}

func returnError() error {
   return nil
}
```

### Expected behavior

Both tags should be checked as the `#nosec` is considered to be the default one.

rafaveira3:
## Closes #425 

This PR aims to add an `OR` statement when checking a specific line of code for both default and alternative flag. If any of them is found, the line should be considered as a false positive.

### Changes

* `analyze.go`: Added an `OR` logic to check for both tags and updated the function description.
* `analyze_test.go`: Updated one function description and adjusted another test as the default tag should also be considered when the alternative is set.
* `Makefile`: Added a few aliases to avoid errors when working outside the `GOPATH`


ccojocar:
@rafaveira3 Thank for this contribution! It seems that one test is failing:
```
Analyzer when processing a package 
  should not ignore vulnerabilities
  /home/travis/gopath/src/github.com/securego/gosec/analyzer_test.go:292
• Failure [0.238 seconds]
Analyzer
/home/travis/gopath/src/github.com/securego/gosec/analyzer_test.go:19
  when processing a package
  /home/travis/gopath/src/github.com/securego/gosec/analyzer_test.go:32
    should not ignore vulnerabilities [It]
    /home/travis/gopath/src/github.com/securego/gosec/analyzer_test.go:292
    Expected
        <[]*gosec.Issue | len:0, cap:16>: []
    to have length 1
    /home/travis/gopath/src/github.com/securego/gosec/analyzer_test.go:312
```

rafaveira3:
Sure, @ccojocar! Sorry about that. I have adjusted the failing test to now pass as the default `#nosec` tag is being used at b00176a.

I have also proposed at 694b37d changes in the Makefile to add a few aliases to avoid errors when working outside the `GOPATH`.

PR #441, opened 2020-02-19
Detecting unsafe `defer` of `os.File.Close()`

caccavale:
This adds a rule for `defer`ing the call to a function which returns an error which should be checked. 
 I've used a mostly arbitrary rule label `G307`.  

Please let me know what changes can/should be made if this is desired upstream.

### DIFF
--- a/rules/bad_defer.go
+++ b/rules/bad_defer.go
@@ -1,0 +1,69 @@
+package rules
+
+import (
+	"fmt"
+	"go/ast"
+	"strings"
+
+	"github.com/securego/gosec"
+)
+
+type deferType struct {
+	typ     string
+	methods []string
+}
+
+type badDefer struct {
+	gosec.MetaData
+	types []deferType
+}
+
+func (r *badDefer) ID() string {
+	return r.MetaData.ID
+}
+
+func normalize(typ string) string {
+	return strings.TrimPrefix(typ, "*")
+}
+
+func contains(methods []string, method string) bool {
+	for _, m := range methods {
+		if m == method {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *badDefer) Match(n ast.Node, c *gosec.Context) (*gosec.Issue, error) {
+	if deferStmt, ok := n.(*ast.DeferStmt); ok {
+		for _, deferTyp := range r.types {
+			if typ, method, err := gosec.GetCallInfo(deferStmt.Call, c); err == nil {
+				if normalize(typ) == deferTyp.typ && contains(deferTyp.methods, method) {
+					return gosec.NewIssue(c, n, r.ID(), fmt.Sprintf(r.What, typ, method), r.Severity, r.Confidence), nil
+				}
+			}
+		}
+
+	}
+
+	return nil, nil
+}
+
+// NewDeferredClosing detects unsafe defer of error returning methods
+func NewDeferredClosing(id string, conf gosec.Config) (gosec.Rule, []ast.Node) {
+	return &badDefer{
+		types: []deferType{
+			{
+				typ:     "os.File",
+				methods: []string{"Close"},
+			},
+		},
+		MetaData: gosec.MetaData{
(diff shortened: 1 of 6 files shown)


rules/bad_defer.go
```diff
@@ -0,0 +1,58 @@
...
+// NewDeferredClosing detects unsafe defer of os.Close
+func NewDeferredClosing(id string, conf gosec.Config) (gosec.Rule, []ast.Node) {
+	return &badDefer{
+		typ:     "os.File",
+		methods: []string{"Close"},
+		MetaData: gosec.MetaData{
+			ID:         id,
+			Severity:   gosec.Medium,
+			Confidence: gosec.High,
+			What:       fmt.Sprintf("Defering closing a file."),
```
ccojocar: I wondering if this message shouldn't be a bit more generic something like:
```
Deferring unsafe method %q on type %q
```

and then wen the issue is created,  you can fill up the method and type.

rules/bad_defer.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"go/ast"
+	"strings"
+
+	"github.com/securego/gosec"
+)
+
+type badDefer struct {
+	gosec.MetaData
+	typ     string
```
ccojocar: I would make this a bit more generic to support multiple types and methods.

```Go
type deferType struct {
    type string
    methods []string
}
```
and use it as:

```Go
type badDefer struct {
  gosec.MetaData
  types []deferType
}
```

rules/rulelist.go
```diff
@@ -81,6 +81,7 @@ func Generate(filters ...RuleFilter) RuleList {
 		{"G303", "Creating tempfile using a predictable path", NewBadTempFile},
 		{"G304", "File path provided as taint input", NewReadFile},
 		{"G305", "File path traversal when extracting zip archive", NewArchive},
+		{"G307", "Unsafe defer of closing a file", NewDeferredClosing},
```
ccojocar: I would change this message to something like: `Unsafe defer call of a method returning an error`

ccojocar:
The build seems to fail because gosec itself has some defer calls to os.Close(). I would ignore the warnings.

testutils/source.go
```diff
@@ -1375,6 +1375,52 @@ func unzip(archive, target string) error {
 	}
 
 	return nil
+}`}, 1, gosec.NewConfig()}}
```
ccojocar: I would be nice to have also a test for happy use case when the error is checked.

caccavale:
Thanks for the feedback!

I believe I have made all the requested changes, please let me know if anything else should be changed.

rules/bad_defer.go
caccavale: Changes made.

rules/bad_defer.go
caccavale: Changes made.

rules/rulelist.go
caccavale: Changes made.

testutils/source.go
caccavale: I've added a trivial `defer` case but I'm not sure exactly what is a happy case for deferring `File.Close()`, do you have a suggestion?

ccojocar: i would add something like:

```Go
defer func() {
  err := f.Close()
  if err != nil {
     log.Printf("error closing the file: %s", err)
  }
}()
```

ccojocar:
Thanks for fixing the review comments! Please could you have a look at the build and fix what's failing? Thanks again!

testutils/source.go
caccavale: This should be done!

caccavale:
Squashed with a better commit message-- should pass CI 🤞

ccojocar:
@caccavale Please could you rebase this pull request? There are some conflicts after merging the other one. Thanks!

PR #442, opened 2020-02-19
Add a check for `WriteFile` permissions.

caccavale:
`io/ioutil`'s `WriteFile` will create a file if one does not exist.  This PR adds a rule for checking whether the permissions of the file exceed a value (default `0600`).

Please let me know if I can make any changes (will squash and rename before merge).

### DIFF
--- a/testutils/source.go
+++ b/testutils/source.go
@@ -1377,6 +1377,55 @@ func unzip(archive, target string) error {
 	return nil
 }`}, 1, gosec.NewConfig()}}
 
+	// SampleCodeG306 - Poor permissions for WriteFile
+	SampleCodeG306 = []CodeSample{
+		{[]string{`package main
+
+import (
+	"bufio"
+	"fmt"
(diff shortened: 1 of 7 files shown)


testutils/source.go
```diff
@@ -1375,6 +1375,52 @@ func unzip(archive, target string) error {
...
+func check(e error) {
+	if e != nil {
+		panic(e)
+	}
+}
+
+func main() {
+	
+	d1 := []byte("hello\ngo\n")
+	err := ioutil.WriteFile("/tmp/dat1", d1, 0744)
```
ccojocar: Please add a block of code which writes a file with `0600` permissions to verify that no warning is generated. Thanks

caccavale: This is done.

caccavale:
I've squashed the requested changes, and added `nogosec` to two places in the project which break this rule.  CI should pass? 🤞

Issue #438, opened 2020-02-12
Add a rule to detect implicit aliasing issues with RangeStmt's


caccavale:
### Summary 
Elements yielded from RangeStmt's have the same memory address.  Not realizing this and storing these addresses for later use is an unsafe behavior.

### Steps to reproduce the behavior
```go
package main

import "fmt"

var vector []*string

func appendVector(s *string) {
	vector = append(vector, s)
}

func printVector() {
	for _, item := range vector {
		fmt.Printf("%s", *item)
	}
	fmt.Println()
}

func foo() (int, *string, string) {
	for _, item := range vector {
		return 0, &item, item
	}
}

func main() {
	for _, item := range []string{"A", "B", "C"} {
		appendVector(&item)
	}

	printVector()

	zero, c_star, c := foo()
	fmt.Printf("%d %v %s", zero, c_start, c)
}
```

### Expected behavior

foo's behavior shouldn't cause any issues, but `appendVector(&item)` will.
I propose adding the following rule:

```go
func (r *implicitAliasing) Match(n ast.Node, c *gosec.Context) (*gosec.Issue, error) {
	if rangeStmt, ok := n.(*ast.RangeStmt); ok && rangeStmt.Value != nil && rangeStmt.Tok.String() == ":=" {
		if ident, ok := rangeStmt.Value.(*ast.Ident); ok {
			name := ident.Name

			var issue *gosec.Issue
			var fun func(n ast.Node) bool
			fun = func(n ast.Node) bool {
				if ret, ok := n.(*ast.ReturnStmt); ok {
					for _, element := range ret.Results {
						if isDereferenceOf(element, name) {
							continue
						}

						ast.Inspect(element, fun)
						if issue != nil {
							return false
						}
					}

					return false
				}

				if isDereferenceOf(n, name) {
					issue = gosec.NewIssue(c, n, r.ID(), r.What, r.Severity, r.Confidence)
					return false
				}

				return issue == nil
			}

			ast.Inspect(rangeStmt.Body, fun)
			return issue, nil
		}
	}

	return nil, nil
}
```

It does have false positives but seems like it is a good stab of what can be a nasty bug.  I haven't participated in much open source software so please advise on how I should go about this; I can open a PR with the work I have done if that is desired.  Thanks.

gcmurphy:
Thanks for raising this happy to review a PR on this!

caccavale:
Working a little to clean this up/reduce false positives-- will PR asap.

PR #443, opened 2020-02-26
Rule which detects aliasing of values in RangeStmt

caccavale:
fixes https://github.com/securego/gosec/issues/438

### DIFF
--- a/rules/implicit_aliasing.go
+++ b/rules/implicit_aliasing.go
@@ -1,0 +1,116 @@
+package rules
+
+import (
+	"fmt"
+	"github.com/securego/gosec/v2"
+	"go/ast"
+	"go/token"
+)
+
+type implicitAliasing struct {
+	gosec.MetaData
+	aliases         map[*ast.Object]struct{}
+	rightBrace      token.Pos
+	acceptableAlias []*ast.UnaryExpr
+}
+
+func (r *implicitAliasing) ID() string {
+	return r.MetaData.ID
+}
+
+func containsUnary(exprs []*ast.UnaryExpr, expr *ast.UnaryExpr) bool {
+	for _, e := range exprs {
+		if e == expr {
+			return true
+		}
+	}
+	return false
+}
+
+func (r *implicitAliasing) Match(n ast.Node, c *gosec.Context) (*gosec.Issue, error) {
+	switch node := n.(type) {
+	case *ast.RangeStmt:
+		// When presented with a range statement, get the underlying Object bound to
+		// by assignment and add it to our set (r.aliases) of objects to check for.
+		if key, ok := node.Value.(*ast.Ident); ok {
+			if assignment, ok := key.Obj.Decl.(*ast.AssignStmt); ok {
+				if len(assignment.Lhs) < 2 {
+					return nil, nil
+				}
+
+				if object, ok := assignment.Lhs[1].(*ast.Ident); ok {
+					r.aliases[object.Obj] = struct{}{}
+
+					if r.rightBrace < node.Body.Rbrace {
+						r.rightBrace = node.Body.Rbrace
+					}
+				}
+			}
+		}
+	case *ast.UnaryExpr:
+		// If this unary expression is outside of the last range statement we were looking at
+		// then clear the list of objects we're concerned about because they're no longer in
+		// scope
+		if node.Pos() > r.rightBrace {
+			r.aliases = make(map[*ast.Object]struct{})
+			r.acceptableAlias = make([]*ast.UnaryExpr, 0)
+		}
+
+		// Short circuit logic to skip checking aliases if we have nothing to check against.
+		if len(r.aliases) == 0 {
+			return nil, nil
+		}
+
+		// If this unary is at the top level of a return statement then it is okay--
+		// see *ast.ReturnStmt comment below.
+		if containsUnary(r.acceptableAlias, node) {
+			return nil, nil
+		}
+
+		// If we find a unary op of & (reference) of an object within r.aliases, complain.
+		if ident, ok := node.X.(*ast.Ident); ok && node.Op.String() == "&" {
+			if _, contains := r.aliases[ident.Obj]; contains {
+				return gosec.NewIssue(c, n, r.ID(), r.What, r.Severity, r.Confidence), nil
+			}
+		}
+	case *ast.ReturnStmt:
+		// Returning a rangeStmt yielded value is acceptable since only one value will be returned
+		for _, item := range node.Results {
+			if unary, ok := item.(*ast.UnaryExpr); ok && unary.Op.String() == "&" {
+				r.acceptableAlias = append(r.acceptableAlias, unary)
+			}
+		}
+	}
+
+	return nil, nil
+}
+
+// NewImplicitAliasing detects implicit memory aliasing of type: for blah := SomeCall() {... SomeOtherCall(&blah) ...}
+func NewImplicitAliasing(id string, conf gosec.Config) (gosec.Rule, []ast.Node) {
+	return &implicitAliasing{
+		aliases:         make(map[*ast.Object]struct{}),
+		rightBrace:      token.NoPos,
+		acceptableAlias: make([]*ast.UnaryExpr, 0),
+		MetaData: gosec.MetaData{
+			ID:         id,
+			Severity:   gosec.Medium,
+			Confidence: gosec.Medium,
+			What:       fmt.Sprintf("Implicit memory aliasing in for loop."),
+		},
+	}, []ast.Node{(*ast.RangeStmt)(nil), (*ast.UnaryExpr)(nil), (*ast.ReturnStmt)(nil)}
+}
+
+/*
+This rule is prone to flag false positives.
+
+Within GoSec, the rule is just an AST match-- there are a handful of other
+implementation strategies which might lend more nuance to the rule at the
+cost of allowing false negatives.
+
+From a tooling side, I'd rather have this rule flag false positives than
+potentially have some false negatives-- especially if the sentiment of this
+rule (as I understand it, and Go) is that referencing a rangeStmt-yielded
+value is kinda strange and does not have a strongly justified use case.
+
+Which is to say-- a false positive _should_ just be changed.
+*/
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ directory you can supply `./...` as the input argument.
 - G503: Import blacklist: crypto/rc4
 - G504: Import blacklist: net/http/cgi
 - G505: Import blacklist: crypto/sha1
+- G601: Implicit memory aliasing of items from a range statement
 
 ### Retired rules
 
--- a/rules/rulelist.go
+++ b/rules/rulelist.go
@@ -96,6 +96,9 @@ func Generate(filters ...RuleFilter) RuleList {
 		{"G503", "Import blacklist: crypto/rc4", NewBlacklistedImportRC4},
 		{"G504", "Import blacklist: net/http/cgi", NewBlacklistedImportCGI},
 		{"G505", "Import blacklist: crypto/sha1", NewBlacklistedImportSHA1},
+
+		// memory safety
+		{"G601", "Implicit memory aliasing in RangeStmt", NewImplicitAliasing},
 	}
 
 	ruleMap := make(map[string]RuleDefinition)
--- a/rules/rules_test.go
+++ b/rules/rules_test.go
@@ -170,8 +170,13 @@ var _ = Describe("gosec rules", func() {
 		It("should detect blacklisted imports - CGI (httpoxy)", func() {
 			runner("G504", testutils.SampleCodeG504)
 		})
+
 		It("should detect blacklisted imports - SHA1", func() {
 			runner("G505", testutils.SampleCodeG505)
+		})
+
+		It("should detect implicit aliasing in ForRange", func() {
+			runner("G601", testutils.SampleCodeG601)
 		})
 
 	})
--- a/testutils/source.go
+++ b/testutils/source.go
@@ -1848,6 +1848,42 @@ func main() {
 		fmt.Printf("%x - %s\n", sha1.Sum([]byte(arg)), arg)
 	}
 }`}, 1, gosec.NewConfig()}}
+
+	// SampleCodeG601 - Implicit ForRange aliasing
+	SampleCodeG601 = []CodeSample{{[]string{`package main
+
+import "fmt"
+
+var vector []*string
+
+func appendVector(s *string) {
+	vector = append(vector, s)
+}
+
+func printVector() {
+	for _, item := range vector {
+		fmt.Printf("%s", *item)
+	}
+	fmt.Println()
+}
+
+func foo() (int, *string, string) {
+	for _, item := range vector {
+		return 0, &item, item
+	}
+}
+
+func main() {
+	for _, item := range []string{"A", "B", "C"} {
+		appendVector(&item)
+	}
+
+	printVector()
+
+	zero, c_star, c := foo()
+	fmt.Printf("%d %v %s", zero, c_start, c)
+}`}, 1, gosec.NewConfig()}}
+
 	// SampleCode601 - Go build tags
 	SampleCode601 = []CodeSample{{[]string{`
 // +build tag


rules/implicit_aliasing.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func (r *implicitAliasing) Match(n ast.Node, c *gosec.Context) (*gosec.Issue, error) {
+	if rangeStmt, ok := n.(*ast.RangeStmt); ok && rangeStmt.Value != nil && rangeStmt.Tok.String() == ":=" {
+		if ident, ok := rangeStmt.Value.(*ast.Ident); ok {
+			name := ident.Name
+
+			var issue *gosec.Issue
+			var fun func(n ast.Node) bool
+			fun = func(n ast.Node) bool {
```
ccojocar: I think that you can compact these two lines buy just:
```
fun := func(n ast.Node) bool {
}
```

rules/implicit_aliasing.go
```diff
@@ -0,0 +1,92 @@
...
+				if ret, ok := n.(*ast.ReturnStmt); ok {
+					for _, element := range ret.Results {
+						if isDereferenceOf(element, name) {
+							continue
+						}
+
+						ast.Inspect(element, fun)
```
ccojocar: We try to avoid `Inspect` in the rules, because will effective traverse again the tree which is not very efficient. Maybe have a look at this rule which uses a cache to traverse only once the tree https://github.com/securego/gosec/blob/a305f10eb98e67d7fd0259d369ded9d3caf31ded/rules/integer_overflow.go#L33.

caccavale:
@ccojocar , is there a way to tell when GoSec leaves an AST node?  The issue I'm having is the following:
```go
for _, item := range []string{"A", "B", "C"} {
    foo(item) // This is safe.
}
item := "D"
bar(&item) // Also safe but problematic.
```
For ease of discussion, let "bad identifier" be an identifier yielded by a range statement which should not be referenced (found as the expression in a unary op with the '&' operator).

I'm trying to collect all bad identifiers and verify they are not referened within the body of the range loop.  With the `go/ast` interface, I push an empty list of bad ID's onto a stack whenever I enter a node-- I add the name of any rangeStmt yielded values onto the top most environment (list in the stack) if the node is a rangeStmt.  I pop the stack whenever the `visitor.Visit(nil)`, emulating scope.  Anytime I see a unary expression with `&` I check whether it is referencing a bad ID.  It's a similar dance to what you all do with the ignore list in this project.  Here's an example where `*implicitAliasing` has the operative data structure:
```go
func (r *implicitAliasing) Match(n ast.Node, c *gosec.Context) (*gosec.Issue, error) {
	if n == nil {
		r.pop()
		return nil, nil
	}

	r.push([]string{})

	switch node := n.(type) {
	case *ast.RangeStmt:
		if node.Value != nil && node.Tok.String() == ":=" {
			if ident, ok := node.Value.(*ast.Ident); ok {
				r.add(ident.Name)
			}
		}
	case *ast.UnaryExpr:
		if ident, ok := node.X.(*ast.Ident); ok && node.Op.String() == "&" {
			if r.has(ident.Name) {
				return gosec.NewIssue(c, n, r.ID(), r.What, r.Severity, r.Confidence), nil
			}
		}

	return nil, nil
}
```

From what I can tell, GoSec doesn't have a way for a rule to visit all the `nil`'s during the AST walk.  I'm not sure I can do a very good job emulating scoping rules without it.  

An alternative would be to carry a list of the bad IDs and invalidate them when they are rebound (filter on assignment statements) but this seems like a less desirable solution since the implementation (based on the names last use) in no way represents the problem it is attempting to solve (scope).  I'd also feel less confident that this would catch all the ways that a name could be recycled and it would also be strange that this is invalid Go code but theoretically could get flagged:
```go
for _, item := range []string{"A", "B", "C"} {
    foo(item) // This is safe.
}
bar(&item) // Not valid.
```

ccojocar:
@caccavale I believe we can catch the issue inside of the `RangeStmt` when an item yielded by the range statement is referenced. Let's try to work out an example:

```Go
package main

import "fmt"

var vector []*string

func appendVector(s *string) {
	vector = append(vector, s)
}

func printVector() {
	for _, item := range vector {
		fmt.Printf("%s", *item)
	}
	fmt.Println()
}

func main() {
	for _, item := range []string{"A", "B", "C"} {
		fmt.Printf("%p\n", &item)
		appendVector(&item)
	}

	printVector()
}
``` 

I run `./gosecutil -tool ast main.go > ast.txt` to generate the AST tree (see it attached). Let's look around the range statement:

```
   276  .  .  .  .  .  0: *ast.RangeStmt {
   277  .  .  .  .  .  .  For: main.go:19:2
   278  .  .  .  .  .  .  Key: *ast.Ident {
   279  .  .  .  .  .  .  .  NamePos: main.go:19:6
   280  .  .  .  .  .  .  .  Name: "_"
   281  .  .  .  .  .  .  .  Obj: *ast.Object {
   282  .  .  .  .  .  .  .  .  Kind: var
   283  .  .  .  .  .  .  .  .  Name: "_"
   284  .  .  .  .  .  .  .  .  Decl: *ast.AssignStmt {
   285  .  .  .  .  .  .  .  .  .  Lhs: []ast.Expr (len = 2) {
   286  .  .  .  .  .  .  .  .  .  .  0: *(obj @ 278)
   287  .  .  .  .  .  .  .  .  .  .  1: *ast.Ident {
   288  .  .  .  .  .  .  .  .  .  .  .  NamePos: main.go:19:9
   289  .  .  .  .  .  .  .  .  .  .  .  Name: "item"
   290  .  .  .  .  .  .  .  .  .  .  .  Obj: *ast.Object {
   291  .  .  .  .  .  .  .  .  .  .  .  .  Kind: var
   292  .  .  .  .  .  .  .  .  .  .  .  .  Name: "item"
   293  .  .  .  .  .  .  .  .  .  .  .  .  Decl: *(obj @ 284)
   294  .  .  .  .  .  .  .  .  .  .  .  }
   295  .  .  .  .  .  .  .  .  .  .  }
   296  .  .  .  .  .  .  .  .  .  }
   297  .  .  .  .  .  .  .  .  .  TokPos: main.go:19:14
   298  .  .  .  .  .  .  .  .  .  Tok: :=
   299  .  .  .  .  .  .  .  .  .  Rhs: []ast.Expr (len = 1) {
   300  .  .  .  .  .  .  .  .  .  .  0: *ast.UnaryExpr {
   301  .  .  .  .  .  .  .  .  .  .  .  OpPos: main.go:19:17
   302  .  .  .  .  .  .  .  .  .  .  .  Op: range
   303  .  .  .  .  .  .  .  .  .  .  .  X: *ast.CompositeLit {
   304  .  .  .  .  .  .  .  .  .  .  .  .  Type: *ast.ArrayType {
   305  .  .  .  .  .  .  .  .  .  .  .  .  .  Lbrack: main.go:19:23
   306  .  .  .  .  .  .  .  .  .  .  .  .  .  Elt: *ast.Ident {
   307  .  .  .  .  .  .  .  .  .  .  .  .  .  .  NamePos: main.go:19:25
   308  .  .  .  .  .  .  .  .  .  .  .  .  .  .  Name: "string"
   309  .  .  .  .  .  .  .  .  .  .  .  .  .  }
   310  .  .  .  .  .  .  .  .  .  .  .  .  }
   311  .  .  .  .  .  .  .  .  .  .  .  .  Lbrace: main.go:19:31
   312  .  .  .  .  .  .  .  .  .  .  .  .  Elts: []ast.Expr (len = 3) {
   313  .  .  .  .  .  .  .  .  .  .  .  .  .  0: *ast.BasicLit {
   314  .  .  .  .  .  .  .  .  .  .  .  .  .  .  ValuePos: main.go:19:32
   315  .  .  .  .  .  .  .  .  .  .  .  .  .  .  Kind: STRING
   316  .  .  .  .  .  .  .  .  .  .  .  .  .  .  Value: "\"A\""
   317  .  .  .  .  .  .  .  .  .  .  .  .  .  }
   318  .  .  .  .  .  .  .  .  .  .  .  .  .  1: *ast.BasicLit {
   319  .  .  .  .  .  .  .  .  .  .  .  .  .  .  ValuePos: main.go:19:37
   320  .  .  .  .  .  .  .  .  .  .  .  .  .  .  Kind: STRING
   321  .  .  .  .  .  .  .  .  .  .  .  .  .  .  Value: "\"B\""
   322  .  .  .  .  .  .  .  .  .  .  .  .  .  }
   323  .  .  .  .  .  .  .  .  .  .  .  .  .  2: *ast.BasicLit {
   324  .  .  .  .  .  .  .  .  .  .  .  .  .  .  ValuePos: main.go:19:42
   325  .  .  .  .  .  .  .  .  .  .  .  .  .  .  Kind: STRING
   326  .  .  .  .  .  .  .  .  .  .  .  .  .  .  Value: "\"C\""
   327  .  .  .  .  .  .  .  .  .  .  .  .  .  }
   328  .  .  .  .  .  .  .  .  .  .  .  .  }
   329  .  .  .  .  .  .  .  .  .  .  .  .  Rbrace: main.go:19:45
   330  .  .  .  .  .  .  .  .  .  .  .  .  Incomplete: false
   331  .  .  .  .  .  .  .  .  .  .  .  }
   332  .  .  .  .  .  .  .  .  .  .  }
   333  .  .  .  .  .  .  .  .  .  }
   334  .  .  .  .  .  .  .  .  }
   335  .  .  .  .  .  .  .  }
   336  .  .  .  .  .  .  }
   337  .  .  .  .  .  .  Value: *(obj @ 287)
   338  .  .  .  .  .  .  TokPos: main.go:19:14
   339  .  .  .  .  .  .  Tok: :=
   340  .  .  .  .  .  .  X: *(obj @ 303)
   341  .  .  .  .  .  .  Body: *ast.BlockStmt {
   342  .  .  .  .  .  .  .  Lbrace: main.go:19:47
   343  .  .  .  .  .  .  .  List: []ast.Stmt (len = 2) {
   344  .  .  .  .  .  .  .  .  0: *ast.ExprStmt {
   345  .  .  .  .  .  .  .  .  .  X: *ast.CallExpr {
   346  .  .  .  .  .  .  .  .  .  .  Fun: *ast.SelectorExpr {
   347  .  .  .  .  .  .  .  .  .  .  .  X: *ast.Ident {
   348  .  .  .  .  .  .  .  .  .  .  .  .  NamePos: main.go:20:3
   349  .  .  .  .  .  .  .  .  .  .  .  .  Name: "fmt"
   350  .  .  .  .  .  .  .  .  .  .  .  }
   351  .  .  .  .  .  .  .  .  .  .  .  Sel: *ast.Ident {
   352  .  .  .  .  .  .  .  .  .  .  .  .  NamePos: main.go:20:7
   353  .  .  .  .  .  .  .  .  .  .  .  .  Name: "Printf"
   354  .  .  .  .  .  .  .  .  .  .  .  }
   355  .  .  .  .  .  .  .  .  .  .  }
   356  .  .  .  .  .  .  .  .  .  .  Lparen: main.go:20:13
   357  .  .  .  .  .  .  .  .  .  .  Args: []ast.Expr (len = 2) {
   358  .  .  .  .  .  .  .  .  .  .  .  0: *ast.BasicLit {
   359  .  .  .  .  .  .  .  .  .  .  .  .  ValuePos: main.go:20:14
   360  .  .  .  .  .  .  .  .  .  .  .  .  Kind: STRING
   361  .  .  .  .  .  .  .  .  .  .  .  .  Value: "\"%p\\n\""
   362  .  .  .  .  .  .  .  .  .  .  .  }
   363  .  .  .  .  .  .  .  .  .  .  .  1: *ast.UnaryExpr {
   364  .  .  .  .  .  .  .  .  .  .  .  .  OpPos: main.go:20:22
   365  .  .  .  .  .  .  .  .  .  .  .  .  Op: &
   366  .  .  .  .  .  .  .  .  .  .  .  .  X: *ast.Ident {
   367  .  .  .  .  .  .  .  .  .  .  .  .  .  NamePos: main.go:20:23
   368  .  .  .  .  .  .  .  .  .  .  .  .  .  Name: "item"
   369  .  .  .  .  .  .  .  .  .  .  .  .  .  Obj: *(obj @ 290)
   370  .  .  .  .  .  .  .  .  .  .  .  .  }
   371  .  .  .  .  .  .  .  .  .  .  .  }
   372  .  .  .  .  .  .  .  .  .  .  }
   373  .  .  .  .  .  .  .  .  .  .  Ellipsis: -
   374  .  .  .  .  .  .  .  .  .  .  Rparen: main.go:20:27
   375  .  .  .  .  .  .  .  .  .  }
   376  .  .  .  .  .  .  .  .  }
   377  .  .  .  .  .  .  .  .  1: *ast.ExprStmt {
   378  .  .  .  .  .  .  .  .  .  X: *ast.CallExpr {
   379  .  .  .  .  .  .  .  .  .  .  Fun: *ast.Ident {
   380  .  .  .  .  .  .  .  .  .  .  .  NamePos: main.go:21:3
   381  .  .  .  .  .  .  .  .  .  .  .  Name: "appendVector"
   382  .  .  .  .  .  .  .  .  .  .  .  Obj: *(obj @ 59)
   383  .  .  .  .  .  .  .  .  .  .  }
   384  .  .  .  .  .  .  .  .  .  .  Lparen: main.go:21:15
   385  .  .  .  .  .  .  .  .  .  .  Args: []ast.Expr (len = 1) {
   386  .  .  .  .  .  .  .  .  .  .  .  0: *ast.UnaryExpr {
   387  .  .  .  .  .  .  .  .  .  .  .  .  OpPos: main.go:21:16
   388  .  .  .  .  .  .  .  .  .  .  .  .  Op: &
   389  .  .  .  .  .  .  .  .  .  .  .  .  X: *ast.Ident {
   390  .  .  .  .  .  .  .  .  .  .  .  .  .  NamePos: main.go:21:17
   391  .  .  .  .  .  .  .  .  .  .  .  .  .  Name: "item"
   392  .  .  .  .  .  .  .  .  .  .  .  .  .  Obj: *(obj @ 290)
   393  .  .  .  .  .  .  .  .  .  .  .  .  }
   394  .  .  .  .  .  .  .  .  .  .  .  }
   395  .  .  .  .  .  .  .  .  .  .  }
   396  .  .  .  .  .  .  .  .  .  .  Ellipsis: -
   397  .  .  .  .  .  .  .  .  .  .  Rparen: main.go:21:21
   398  .  .  .  .  .  .  .  .  .  }
   399  .  .  .  .  .  .  .  .  }
   400  .  .  .  .  .  .  .  }
   401  .  .  .  .  .  .  .  Rbrace: main.go:22:2
   402  .  .  .  .  .  .  }
   403  .  .  .  .  .  }
```

We can have something like this:
- On `RangeStmt` you can lookup for the `AssignStmt` and then you can get the `Obj` address:
```
   290  .  .  .  .  .  .  .  .  .  .  .  Obj: *ast.Object {
   291  .  .  .  .  .  .  .  .  .  .  .  .  Kind: var
   292  .  .  .  .  .  .  .  .  .  .  .  .  Name: "item"
   293  .  .  .  .  .  .  .  .  .  .  .  .  Decl: *(obj @ 284)
   294  .  .  .  .  .  .  .  .  .  .  .  }
```
- You can store in a map this address with the `RangStmt` address as value
- You can then keep watching for `UnaryExpr` and get their `Obj` address if the `Op` is a reference such as `&`:
```
   363  .  .  .  .  .  .  .  .  .  .  .  1: *ast.UnaryExpr {
   364  .  .  .  .  .  .  .  .  .  .  .  .  OpPos: main.go:20:22
   365  .  .  .  .  .  .  .  .  .  .  .  .  Op: &
   366  .  .  .  .  .  .  .  .  .  .  .  .  X: *ast.Ident {
   367  .  .  .  .  .  .  .  .  .  .  .  .  .  NamePos: main.go:20:23
   368  .  .  .  .  .  .  .  .  .  .  .  .  .  Name: "item"
   369  .  .  .  .  .  .  .  .  .  .  .  .  .  Obj: *(obj @ 290)
   370  .  .  .  .  .  .  .  .  .  .  .  .  }
   371  .  .  .  .  .  .  .  .  .  .  .  }
```
- Look up the object address in the range map, if the object is present, we can create a gosec issue.

I think we can optimise this a bit by storing also the position of the `Rbrace` of the`BlockStmt` of the `Body` from the `RangeStmt` in the map (see `Rbrace`):

```
	RangeStmt struct {
		For        token.Pos   // position of "for" keyword
		Key, Value Expr        // Key, Value may be nil
		TokPos     token.Pos   // position of Tok; invalid if Key == nil
		Tok        token.Token // ILLEGAL if Key == nil, ASSIGN, DEFINE
		X          Expr        // value to range over
		Body       *BlockStmt
	}
	// A BlockStmt node represents a braced statement list.
	BlockStmt struct {
		Lbrace token.Pos // position of "{"
		List   []Stmt
		Rbrace token.Pos // position of "}", if any (may be absent due to syntax error)
	}
```

On the first `UnaryExpr` out of this position, we can reset the map (see `OpPos`)
```
	UnaryExpr struct {
		OpPos token.Pos   // position of Op
		Op    token.Token // operator
		X     Expr        // operand
	}
```

From that point on, we don't check anymore `UnaryExpr` nodes as long as a new `RangeStmt` is not recorded in the map (e.g. map is empty). 

[ast.txt](https://github.com/securego/gosec/files/4317699/ast.txt)

ccojocar:
@caccavale Do you need any additional help to be able to move forward with this rule improvement? Thanks

caccavale:
Sorry for the delay!  Hows it look?

Thanks for the detailed explanation/walkthrough of the AST.  I didn't realize the strength in the `.Obj` references being intentionally aliased.  Let me know how to proceed from here.

PR #462, opened 2020-04-14
Use the latest color package to get the color working with tmux

ccojocar:


cmd/gosec/main.go
```diff
@@ -286,8 +283,9 @@ func main() {
 	}
 
 	// Color flag is allowed for text format
-	if *flagColor && *flagFormat != "text" {
-		logger.Fatalf("cannot set color with %s format. Only text format is accepted", *flagFormat)
+	var color bool
+	if *flagFormat != "text" {
```
marcosinger, line 287: shouldn't it be `== "text"`, otherwise the color won't be applied

ccojocar: Thanks! I fixed it in https://github.com/securego/gosec/pull/463.

Issue #465, opened 2020-04-17
G101: pattern matching validates RHS only if it matches LHS (for patterns)


ersuniljoshi:
### Summary 
matching patterns with RHS  is not supported or only supported if LHS matches the given pattern.
hardcoded_credentials.go  Line number : 69.

### Steps to reproduce the behavior
UseCase: Linting the hardcoded password usage.
example: val : = "abc1234" # note: abc1234 is a password.

### gosec version
master

### Go version (output of 'go version')
go version go1.14.2 darwin/amd64
### Operating system / Environment
go version go1.14.2 darwin/amd64

### Expected behavior
It should detect patterns on rhs as well.

### Actual behavior
Not detects the password abc1234 (untill val is provided as pattern)

ccojocar:
I am trying to make sense of your example:
```
val : = "abc1234"
```

What doesn't it hint that on the RHS is a password? It could be just an ordinary variable holding some string. 

The rule works by matching the variable name against the pattern and the entropy of the value from the RHS. 
For instance something like:
```
user_password := "78373q883"
```
it will be a match because the variable name matches the pattern and the `78373q883` has a high entropy. 

Could you please provide a more detailed example? Maybe I am missing something.

ersuniljoshi:
@ccojocar  Thanks for your prompt response. 
My UseCase: I want to pattern match (RHS) few generally used passwords patterns in a project!
Irrespective of what variable name i am using on LHS.
We can have the config parameters like 
lhs_pattern
rhs_pattern

IMHO, Doing this will extend the usage of gosec for more general cases and configurable parameters can help to leave this at users discretion.
Entropy can still be used if required.

ccojocar:
Usually on the RHS there is a random looking string as a password for which pattern matching won't quite work. 

Something like:
```
val := "password"
```
it is not a good practice from security perspective. 

but something like:
```
val := "d87183jajdjd"
```
it is difficult to reliably detected with pattern matching without generating false positives.  The rule is using entropy calculation on RHS in order to detect random looking strings in combination with the pattern matching on the variable name.

Also other tools such as sonarqube adopt a similar approach. See https://github.com/Skyscanner/sonar-secrets/blob/068bd8b822e66ed2a98dfc33fdcad0611852daea/java/src/main/java/org/sonar/skyscanner/java/checks/Passwords.java#L55

ccojocar:
[...]

PR #466, opened 2020-04-17
Enhance the hardcoded credentials rule to check the equality and non-equality of strings

ccojocar:
related to #465


go.mod
```diff
@@ -10,6 +10,7 @@ require (
 	github.com/nbutton23/zxcvbn-go v0.0.0-20180912185939-ae427f1e4c1d
 	github.com/onsi/ginkgo v1.12.0
 	github.com/onsi/gomega v1.9.0
+	github.com/shurcooL/graphql v0.0.0-20181231061246-d48a9a75455f
```
gcmurphy, line 13: how come the graphql library was added here?

ccojocar: I run `go mod tidy` should be removed now. Thanks for catching this!

Issue #369, opened 2019-09-12
G201 and G202 (SQL query construction) do not detect lowercase SQL keywords


nchelluri:
### Summary 
(I tested this for G201 but inspected the code for G202 and believe it has the same issue.)

G201 will detect SQL string formatting when I have `fmt.Sprintf("SELECT * FROM TEST UNION %s", "SELECT * FROM test")` in my code. However, it will not detect when I have `fmt.Sprintf("select * from test union %s", "select * from test")` in my code. I believe this is a bug.

See: https://github.com/securego/gosec/blob/2.0.0/rules/sql.go#L184 and https://github.com/securego/gosec/blob/2.0.0/rules/sql.go#L94

### Steps to reproduce the behavior
Create a file `sql.go` with the following content:
```
package main

import "fmt"

func sqlConstructionLowercase() string {
	return fmt.Sprintf("select * from test union %s", "select * from test")
}

func sqlConstructionUppercase() string {
	return fmt.Sprintf("SELECT * FROM TEST UNION %s", "SELECT * FROM test")
}

func main() {
	fmt.Println(sqlConstructionLowercase())
	fmt.Println(sqlConstructionUppercase())
}
```

Run `gosec` on it:

```
vagrant@ct100-test:~/dev/gosec/sql $ ../gosec .
[gosec] 2019/09/12 11:48:21 Including rules: default
[gosec] 2019/09/12 11:48:21 Excluding rules: default
[gosec] 2019/09/12 11:48:21 Import directory: /home/vagrant/dev/gosec/sql
[gosec] 2019/09/12 11:48:21 Checking package: main
[gosec] 2019/09/12 11:48:21 Checking file: /home/vagrant/dev/gosec/sql/sql.go
Results:


[/home/vagrant/dev/gosec/sql/sql.go:10] - G201: SQL string formatting (Confidence: HIGH, Severity: MEDIUM)
  > fmt.Sprintf("SELECT * FROM TEST UNION %s", "SELECT * FROM test")


Summary:
   Files: 1
   Lines: 16
   Nosec: 0
  Issues: 1

vagrant@ct100-test:~/dev/gosec/sql $ 
```

Only the uppercased version is shown as a problem. Both should be.

### gosec version
```
VERSION: 2.0.0
GIT TAG: 2.0.0
BUILD DATE: 2019-05-02T15:22:02Z
```

### Go version (output of 'go version')
```
go version go1.12.9 linux/amd64
```

### Operating system / Environment
[...]

Issue #345, opened 2019-09-03
Improve accuracy of SQL rule


ccojocar:
### Summary 

The sql rule is using only a regexp expression to identify when is a SQL query into a constant string. 

https://github.com/securego/gosec/blob/4b59c948083cd711b6a8aac8f32721b164899f57/rules/sql.go#L204

This generates quite some false positives (See issues  #341 and #264). It would be great to improve its accuracy. There are various options to achieve this:
 - Improve the heuristic of the rule by checking more acuratly the context - see @gcmurphy's suggestions in the comments below 
- Investigate if a SQL parser can validate additionally the syntax in certain use cases when the query can be reconstructed - possible SQL parse https://github.com/xwb1989/sqlparser. 

### Expected behavior

Reduce the number of false positive warnings generated by the rule.

### Actual behavior

It doesn't really differentiate between a plain string starting with a SQL keyword (e.g. SELECT) and a valid SQL query.

gdhameeja:
I'd like to work on this.

ccojocar:
Great! Feel free to join our slack channel if you have any question.

gcmurphy:
The places where we're creating issues are typically when a SQL query is being constructed by concatenation or string interpolation. In the examples below the SQL query we're matching on will not likely be considered valid by the SQL parser would they?

`q :=  fmt.Sprintf("SELECT * from foo where bar = %s", untrustedVar)`

or 

` q := "SELECT * from foo where bar = " + untrustedVar`

There might be a way around this but just an initial thought around actual parsing of the SQL syntax. The other potential issue is DB compatibility, but I'm less worried about that TBH. 

There's potentially a couple of things we could do to improve the confidence of the rule. Simple things like checking if 'database/sql' is imported of if the matched string is being used in a db.Query or db.Exec call could help reduce false positives.

ccojocar:
My initial thinking was that we can somehow reconstruct the query and run it through a SQL parser to validate it, but I realise that it is not really possible when parts of the query are stored in various variables - they are only available at runtime. 

I agree that adding some better heuristic in the detection can improve significantly the rule accuracy. 

I am going to change the title to reflect better the scope of this work.

ccojocar:
@gcmurphy I updated also the description to point first to you suggestions. Thanks for reviewing this!

ccojocar:
Fixed by https://github.com/securego/gosec/pull/479

PR #479, opened 2020-05-25
Various improvements in the SQL rules

### LINKED ISSUE #264: SQL String concatenation issues are ignored when multiple strings are concatenated
### Summary 
If a SQL query string is a concatenation of multiple strings gosec does not complain.

Check the following code snippet with `gosec`:
```
package main

import "database/sql"

func main() {
	query("; DELETE * FROM TABLE;")
}

func query(param string) {
	clt, _ := sql.Open("postgres", "")

	clt.Query("SELECT * FROM table" + "WHERE" + param)
}
```

### gosec version
Commit 41809946d461bcf365ab440a5d79bfbd1ce2ebbf
go version go1.11.2 linux/amd64
Linux
### Expected behavior
Gosec finds issue `G202: SQL string concatenation (Confidence: HIGH, Severity: MEDIUM)`
### Actual behavior
Gosec does not find an issue.

ccojocar:
- Improve the SQL concatenation and string formatting rules to be applied only in the database/sql context
- Make the heuristic case insensitive
- Handle multiple strings concatenation 

fixes #264 #369 #345

### DIFF
--- a/issue_test.go
+++ b/issue_test.go
@@ -98,7 +98,8 @@ import (
    	"net"
 )
 func main() {
-	_, _ := net.Listen("tcp", "0.0.0.0:2000")
+	_, _ := net.Listen("tcp", 
+	"0.0.0.0:2000")
 }
 `
 			pkg := testutils.NewTestPackage()
@@ -122,7 +123,7 @@ func main() {
 			Expect(err).ShouldNot(HaveOccurred())
 			Expect(issue).ShouldNot(BeNil())
 			Expect(issue.File).Should(MatchRegexp("foo.go"))
-			Expect(issue.Line).Should(MatchRegexp("7"))
+			Expect(issue.Line).Should(MatchRegexp("7-8"))
 			Expect(issue.Col).Should(Equal("10"))
 		})
 
--- a/issue_test.go
+++ b/issue_test.go
@@ -1,6 +1,7 @@
 package gosec_test
 
 import (
+	"fmt"
 	"go/ast"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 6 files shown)


issue_test.go
```diff
@@ -91,21 +91,23 @@ var _ = Describe("Issue", func() {
 		})
 
 		It("should provide accurate line and file information for multi-line statements", func() {
-			var target *ast.BinaryExpr
```
gcmurphy, line 94: Think the goal of this test was to make sure line range was accurate for findings that span multiple lines like this:

```go
2: 
3: q := `SELECT * FROM table WHERE` +
4:    os.Args[1] +  `= ?`
5:
```
So would return a range `3-4` rather than just `3` for example. The test code you've replaced here doesn't really do that afaict.

ccojocar: Thanks for catching this. I changed the test to generate a multi-line finding.

rules/sql.go
```diff
@@ -212,8 +293,11 @@ func NewSQLStrFormat(id string, conf gosec.Config) (gosec.Rule, []ast.Node) {
 			},
 		},
 	}
-	rule.calls.AddAll("fmt", "Sprint", "Sprintf", "Sprintln", "Fprintf")
+	rule.AddAll("*database/sql.DB", "Query", "QueryContext", "QueryRow", "QueryRowContext")
+	rule.AddAll("*database/sql.Tx", "Query", "QueryContext", "QueryRow", "QueryRowContext")
+	rule.fmtCalls.AddAll("fmt", "Sprint", "Sprintf", "Sprintln", "Fprintf")
 	rule.noIssue.AddAll("os", "Stdout", "Stderr")
 	rule.noIssueQuoted.Add("github.com/lib/pq", "QuoteIdentifier")
```
lukasaron, line 300: Why is here PostgreSQL import? It should be DB independent, right?

ccojocar: This package seems to be used by quite some people. It was already in there. I don't want to remove it as part of this refactoring since it appears to be useful to remove some false positives.

lukasaron: Do you feel this approach is a good practice? IMHO when the PostgreSQL hide some false alarms doesn't mean that for instance MySQL users can benefit from this.

ccojocar:
@gcmurphy I pushed the fixes. Do you mind having another look? Thanks

PR #498, opened 2020-06-29
Change naming rule from blacklist to blocklist

evalphobia:
This PR changes naming from `blacklist` to `blocklist`.
some codes are changed by formatter. (e.g. trailing spaces)


ccojocar:
The file where the blocklist is stored also needs to be renamed. Done in this PR https://github.com/securego/gosec/pull/499.

PR #508, opened 2020-07-27
Add io.CopyBuffer function to rule G110

### LINKED ISSUE #507: io.CopyBuffer  G110: Potential DoS vulnerability via decompression bomb.
### Summary 

I have Go code that untars a tarball, with the use of `io.Copy` in the `case tar.TypeReg` of the switch. There is like many examples of this online... here is one: https://medium.com/@skdomino/taring-untaring-files-in-go-6b07cf56bc07

The `io.Copy` is captured by gosec linter as a decompression-bomb.  This lint needs more "awareness".  It is possible to replace `io.Copy` with `_, err := io.CopyBuffer(f, tr, nil)` which is NOT captured by the linter, but is basically the same code.  

`Copy`  is go is `return copyBuffer(dst, src, nil)`

https://github.com/golang/go/blob/master/src/io/io.go#L368

Change any code that is identified by gosec with `io.Copy(x, y)` as `G110: Potential DoS vulnerability via decompression bomb (gosec)` to `io.CopyBuffer(x, y, nil)`.  It will work exactly the same but will not be captured by gosec :(

### gosec version
The version is the `golangci-lint has version 1.29.0 built from 6a68907`

`go version go1.14.5 darwin/amd64`

darwin Mac OSX

### Expected behavior
io.CopyBuffer(dst, src, nil) should fail as io.Copy does

ccojocar:
fixes #507

### DIFF
--- a/rules/decompression-bomb.go
+++ b/rules/decompression-bomb.go
@@ -95,6 +95,7 @@ func NewDecompressionBombCheck(id string, conf gosec.Config) (gosec.Rule, []ast.
 
 	copyCalls := gosec.NewCallList()
 	copyCalls.Add("io", "Copy")
+	copyCalls.Add("io", "CopyBuffer")
 
 	return &decompressionBombCheck{
 		MetaData: gosec.MetaData{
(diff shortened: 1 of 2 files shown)


kensipe:
👍

PR #514, opened 2020-08-17
Write the logs to stdout instead the stderr

### LINKED ISSUE #511: Proposal: Write log to stdout and not stderr
### Summary 

I understand there is a flag to write the log to a file instead of stderr, but is there a reason why the log is not written to stdout by either a flag or by default? I feel as though messages such as "2020/08/05 15:29:41 Import directory: C:\..." should be written to the stdout instead? 

### gosec version

2.4.0

go1.14.4 windows/amd64

Windows 10

### Expected behavior

Log messages get produced to stdout

### Actual behavior

Log messages get produced to stderr

ccojocar:
fixes #511

### DIFF
--- a/cmd/gosec/main.go
+++ b/cmd/gosec/main.go
@@ -273,7 +273,7 @@ func main() {
 	}
 
 	// Setup logging
-	logWriter := os.Stderr
+	logWriter := os.Stdout
 	if *flagLogfile != "" {
 		var e error
 		logWriter, e = os.Create(*flagLogfile)


gcmurphy:
I think I disagree with this change. What if you want to pipe the output of gosec with another program? You would not be able to distinguish between log output and result output easily with this change. 

For example thjs will break I think: 

```
gosec -fmt=json ./... | jq .Stats
```

ccojocar:
This makes sense. Let's keep it like this. Thanks for clarifications.

PR #527, opened 2020-09-21
Fix wrong local installation info

boratanrikulu:
There's no `v2` folder inside https://github.com/securego/gosec/tree/master/cmd.  
Local installation information was wrong, I fixed it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ jobs:
 ### Local Installation
 
 ```bash
-go get github.com/securego/gosec/v2/cmd/gosec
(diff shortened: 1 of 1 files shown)


ccojocar:
There is not folder but the module contains a `v2` in the path. See https://github.com/securego/gosec/blob/868556b846eafed2980f9708077c402be36ed28f/cmd/gosec/main.go#L26. 

This is how go modules manages major versions. The `v2` is required in the installation path.

ccojocar:
Closing this since the update is not required. Thanks!

PR #532, opened 2020-09-30
Update README.md

holgermetschulat:
Typo ?!? v2 doesn't exist as a path in go get...

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ jobs:
 ### Local Installation
 
 ```bash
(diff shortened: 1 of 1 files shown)


ccojocar:
Thank you for this pull request but it is the right path. This is how Go modules manages major versions. It is defined here https://github.com/securego/gosec/blob/6bd6e4ba2c6e34f83241ce62f54eb2a6e66f9903/go.mod#L1.

PR #536, opened 2020-10-26
Enable CodeQL Security Scan

VinodAnandan:
https://github.blog/2020-09-30-code-scanning-is-now-available/


ccojocar:
Thanks for this contribution but we scan already the source code with gosec. We like to eat our own dog food. Maybe we will consider to add code scanning later but I won't merge it for now.

Issue #549, opened 2020-12-17
Some sample code does not compile


cbandy:
### Summary 

The documentation on `github.com/securego/gosec/v2/testutils.CodeSample` says
> CodeSample encapsulates a snippet of source code that compiles, and how many errors should be detected

However, not all the code in these samples compiles. I think I see a test that is _supposed_ to verify this, so my concern is that tests are incorrect.

https://github.com/securego/gosec/blob/9fe0b2e21afb166f331098c282aa5de6c96d3253/rules/rules_test.go#L35-L41

### Steps to reproduce the behavior

1. Copy the first value of `SampleCodeG601` and paste it in Go playground: https://play.golang.org/p/MfxutjCC9YO
2. Press "Run" to find that it does not compile.

### gosec version

v2.5.0 and HEAD.

### Expected behavior

I expected the sample to run (and show what G601 protects against).

### Actual behavior

> Go build failed.

```
./prog.go:15:13: cannot use &item (type **string) as type *string in return argument
./prog.go:15:13: cannot use item (type *string) as type string in return argument
./prog.go:17:1: missing return at end of function
./prog.go:24:31: undefined: c_start
```

ccojocar:
Thanks for reporting this. Please could you send a pull request which fixes the test samples which doesn't compile and it suppose to? I'll be happy to merge it in.

ccojocar:
Fixed by https://github.com/securego/gosec/pull/554

PR #554, opened 2021-01-01
Assert that sample code compiles

cbandy:
This partially addresses #549. I fixed what seemed like typos to me, but there are still three failing specs that I'm not equipped to address:

1. `[Fail] gosec rules report correct errors for all samples [It] should detect pprof endpoint`

   The failing sample was introduced in 29341f6e. It seems to assert that importing `net/http/pprof` without an alias disables rule G108 when using `http.DefaultServeMux`. Is that the correct behavior?

   ```
   sample_1_0.go:8:2: "net/http/pprof" imported but not used
   ```

2. `[Fail] gosec rules report correct errors for all samples [It] should detect file path provided as taint input`

   The failing sample was introduced in f13b8bc6. It seems to assert that using `filepath.Rel` sanitizes path input, but that function returns an error. How should that error affect this sample?

   ```
   sample_8_0.go:11:37: too few arguments in call to filepath.Rel
   sample_8_0.go:11:16: cannot initialize 1 variables with 2 values
   ```

3. `[Fail] gosec rules report correct errors for all samples [It] should detect implicit aliasing in ForRange`

   The failing sample was introduced in ee3146e6. The `foo` function seems to be implemented for a `var vector []string`, while the rest of the sample is for a `var vector []*string`. What in this sample should change?

   ```
   sample_0_0.go:20:13: cannot use &item (value of type **string) as *string value in return statement
   sample_0_0.go:20:20: cannot use item (variable of type *string) as string value in return statement
   sample_0_0.go:22:1: missing return
   ```

### DIFF
--- a/rules/rules_test.go
+++ b/rules/rules_test.go
@@ -39,6 +39,7 @@ var _ = Describe("gosec rules", func() {
 				}
 				err := pkg.Build()
 				Expect(err).ShouldNot(HaveOccurred())
+				Expect(pkg.PrintErrors()).Should(BeZero())
 				err = analyzer.Process(buildTags, pkg.Path)
 				Expect(err).ShouldNot(HaveOccurred())
 				issues, _, _ := analyzer.Report()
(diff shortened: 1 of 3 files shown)


ccojocar (approved):
Thanks for looking into this and for adding the check for compilation errors. I fixed these failing code samples. It is good to be merged now.

PR #559, opened 2021-01-21
fix for sarif which translates from issue severity to sarif level

wolfeidau:
I followed the example from the sonarqube formatter and guidance from the spec, which i linked in the code.


ccojocar (approved):
LGMT, thanks for this contribution!

PR #561, opened 2021-01-22
fix sarif add default configuration set to correct level

wolfeidau:
i have been working on getting a good example of integration with the GitHub security scanning, which required a bit of research into how SARIF is processed by this service. One of the issues I have had is the checks seemed to default to warning so it wouldn't mark the PR as failed, this was resolved with the addition of a level on the rule.

For reference I got a hint about this information from https://github.com/hunt3rkillerz/Snyk-To-Sarif-JS/blob/main/snyk-to-sarif.js#L161-L163

Screenshot from 2021-01-22 20-18-24

### DIFF
--- a/output/sarif_format.go
+++ b/output/sarif_format.go
@@ -21,12 +21,17 @@ type sarifProperties struct {
 }
 
 type sarifRule struct {
-	ID               string           `json:"id"`
-	Name             string           `json:"name"`
-	ShortDescription *sarifMessage    `json:"shortDescription"`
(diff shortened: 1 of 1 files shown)


ccojocar (approved):

ccojocar:
@wolfeidau Thanks for this contribution and looking into geting gosec integrated with GitHub Code Scanning. 

It would be awesome if you could add a section in the README file which describes how to setup gosec with GitHub Code Scanning.

PR #565, opened 2021-01-30
Fix sarif formatting issues

### LINKED ISSUE #563: Gosec is generating invalid sarif files
### Summary 
I came across some issues regarding the SARIF format and, when trying to validate on [sarif validator](https://sarifweb.azurewebsites.net/Validation). Basically there is three issues:

1. `$schema` is outdated, should be `https://schemastore.azurewebsites.net/schemas/json/sarif-2.1.0-rtm.5.json` or `https://raw.githubusercontent.com/oasis-tcs/sarif-spec/master/Schemata/sarif-schema-2.1.0.json`;
2. driver object in `tools` should contains `version` or `semanticVersion`, following [§3.19.2 standard rule](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317535);
3. [rules array's elements in `driver` must be unique](https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317556), so it can be referenced on `results `.

1. Generate a sarif file from an analysis, preferable with many occurrences from same rule
2. Validate it on [sarif validator](https://sarifweb.azurewebsites.net/Validation)

### gosec version
v 2.0

go version go1.15.6 linux/amd64

Linux Ubuntu 20.04

### Expected behavior
The file should be validated without issues

### Actual behavior
The file has the issues described above

DimaSalakhov:
closes #563 

PR attempts to fix issues formatting issues with produced SARIF report.
Covered changes:
1. Use running tip version of the SARIF `$schema`
2. `tool` declares it's version. Used the same version as the `$schema`, might not be ideal but that fixes the immediate problem of invalid SARIF schema. The other option could be passing through commit SHA or gosec module version.
3. `tool`'s `rules` couldn't be duplicated
4. `results`' `locations` array should have only a single entry (https://docs.oasis-open.org/sarif/sarif/v2.1.0/os/sarif-v2.1.0-os.html#_Toc34317650).

### DIFF
--- a/output/formatter.go
+++ b/output/formatter.go
@@ -209,6 +209,7 @@ func convertToSarifReport(rootPaths []string, data *reportInfo) (*sarifReport, e
 	tool := &sarifTool{
 		Driver: &sarifDriver{
 			Name:           "gosec",
+			Version:        "2.1.0",
 			InformationURI: "https://github.com/securego/gosec/",
 			Rules:          rules,
 		},
--- a/output/formatter.go
+++ b/output/formatter.go
@@ -180,27 +180,36 @@ func convertToSonarIssues(rootPaths []string, data *reportInfo) (*sonarIssues, e
 func convertToSarifReport(rootPaths []string, data *reportInfo) (*sarifReport, error) {
 	sr := buildSarifReport()
 
-	var rules []*sarifRule
-	var locations []*sarifLocation
+	rules := make([]*sarifRule, 0)
 	results := []*sarifResult{}
 
-	for index, issue := range data.Issues {
-		rules = append(rules, buildSarifRule(issue))
+	for _, issue := range data.Issues {
+		sarifRule := buildSarifRule(issue)
+		ruleIndex := -1
+		for i, rule := range rules {
+			if rule.ID == sarifRule.ID {
+				ruleIndex = i
+				break
+			}
+		}
+		if ruleIndex == -1 {
+			rules = append(rules, sarifRule)
+			ruleIndex = len(rules) - 1
+		}
 
 		location, err := buildSarifLocation(issue, rootPaths)
 		if err != nil {
 			return nil, err
 		}
-		locations = append(locations, location)
 
 		result := &sarifResult{
 			RuleID:    fmt.Sprintf("%s (CWE-%s)", issue.RuleID, issue.Cwe.ID),
-			RuleIndex: index,
+			RuleIndex: ruleIndex,
 			Level:     getSarifLevel(issue.Severity.String()),
 			Message: &sarifMessage{
 				Text: issue.What,
 			},
-			Locations: locations,
+			Locations: []*sarifLocation{location},
 		}
 
 		results = append(results, result)
--- a/output/formatter.go
+++ b/output/formatter.go
@@ -180,21 +180,24 @@ func convertToSonarIssues(rootPaths []string, data *reportInfo) (*sonarIssues, e
 func convertToSarifReport(rootPaths []string, data *reportInfo) (*sarifReport, error) {
 	sr := buildSarifReport()
 
+	type rule struct {
+		index int
+		rule  *sarifRule
+	}
+
 	rules := make([]*sarifRule, 0)
+	rulesIndices := make(map[string]rule)
+	lastRuleIndex := -1
+
 	results := []*sarifResult{}
 
 	for _, issue := range data.Issues {
-		sarifRule := buildSarifRule(issue)
-		ruleIndex := -1
-		for i, rule := range rules {
-			if rule.ID == sarifRule.ID {
-				ruleIndex = i
-				break
-			}
-		}
-		if ruleIndex == -1 {
-			rules = append(rules, sarifRule)
-			ruleIndex = len(rules) - 1
+		r, ok := rulesIndices[issue.RuleID]
+		if !ok {
+			lastRuleIndex++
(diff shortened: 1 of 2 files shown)


output/formatter.go
```diff
@@ -180,27 +180,36 @@ func convertToSonarIssues(rootPaths []string, data *reportInfo) (*sonarIssues, e
...
 		result := &sarifResult{
 			RuleID:    fmt.Sprintf("%s (CWE-%s)", issue.RuleID, issue.Cwe.ID),
-			RuleIndex: index,
+			RuleIndex: ruleIndex,
 			Level:     getSarifLevel(issue.Severity.String()),
 			Message: &sarifMessage{
 				Text: issue.What,
 			},
-			Locations: locations,
+			Locations: []*sarifLocation{location},
```
ccojocar, line 212: This is a good fix. The current implementation keeps appending the location of each issue to the locations slice. This resulting in the last sarif result having all the locations of previous issues, which it seems to me like a bug.

output/formatter.go
```diff
@@ -180,27 +180,36 @@ func convertToSonarIssues(rootPaths []string, data *reportInfo) (*sonarIssues, e
...
 	sr := buildSarifReport()
 
-	var rules []*sarifRule
-	var locations []*sarifLocation
+	rules := make([]*sarifRule, 0)
 	results := []*sarifResult{}
 
-	for index, issue := range data.Issues {
-		rules = append(rules, buildSarifRule(issue))
+	for _, issue := range data.Issues {
```
ccojocar, line 186: This approach doesn't seem right to me because it is removing issues form the result produced by the same rule but at a different location. 

It seems that the recommended way to solve this in the spec (http://docs.oasis-open.org/sarif/sarif/v2.0/csprd01/sarif-v2.0-csprd01.html - see section `3.19.6 ruleId property`) is to append an index to to the duplicated rule id. In our case something like: 

```
G101 (CEW-123)
G101 (CEW-123)-1
G101 (CEW-123)-2
G101 (CEW-123)-3
....
```

Maybe the best way to solve this is to extend the signature of `buildSafariRule` function with an index like this:

```
func buildSarifRule(issue *gosec.Issue, index int) *sarifRule 
```

We can skip appending the index `0` to rule ID because is the first encountered rule but start appending it form `1`. 

Also I think that you can keep track of duplicated rules in `convertToSafariReport` function by using a map instead of iterating twice over the rules slice:

```
rules := []*safariRule{}
rulesMap := map[string]int


for index, issue := range data.Issues {

      ruleIdIndex, ok := rulesMap[issue.RuleID]; 
      if ok {
           ruleidIndex += 1
            rulesMap[issue.RuleID] = ruleIndex
      } else {
           rulesMap[issue.RuleID] = 0
      }

       rules = append(rules, buildSafariRule(issue, ruleidIndex))
....
```

ccojocar (changes requested):

output/formatter.go
```diff
@@ -180,27 +180,36 @@ func convertToSonarIssues(rootPaths []string, data *reportInfo) (*sonarIssues, e
...
+		}
 
 		location, err := buildSarifLocation(issue, rootPaths)
 		if err != nil {
 			return nil, err
 		}
-		locations = append(locations, location)
 
 		result := &sarifResult{
 			RuleID:    fmt.Sprintf("%s (CWE-%s)", issue.RuleID, issue.Cwe.ID),
```
ccojocar, line 206: We should change also this `RuleID` to:
```
result := &safariResult{
    RuleID: rule.ID,
....

File: output/formatter.go (hunk 1 above)
Comment (DimaSalakhov) [reply output/formatter.go:212]: that's right, only the first result had the correct location, all following results carried locations of previous item + self.

File: output/formatter.go (hunk 2 above)
Comment (DimaSalakhov) [reply output/formatter.go:186]: >  it is removing issues from the result produced by the same rule but at a different location.
I don't think it's true, but maybe I misunderstand you. Let me explain what I see happens in the report:
- we add deduplicated rules
- we add all rule violations one-by-one, each violation refers to the rule and the location

Here's the excerpt from the report:
```
{
	"runs": [
		{
			"tool": {
				"driver": {
					"rules": [
						{
							"id": "G404 (CWE-338)",
							...
						},
						{
							"id": "G402 (CWE-295)",
							...
						}
					]
				}
			},
			"results": [
				{
					"ruleId": "G404 (CWE-338)",
					"ruleIndex": 0,
					"locations": [
						{
							"physicalLocation": {
								"artifactLocation": {
									"uri": "api/chisel/tunnel.go"
								},
								...
							}
						}
					]
				},
				{
					"ruleId": "G404 (CWE-338)",
					"ruleIndex": 0,
					"locations": [
						{
							"physicalLocation": {
								"artifactLocation": {
									"uri": "api/chisel/proxy.go"
								},
								...
							}
						}
					]
				},
				{
					"ruleId": "G402 (CWE-295)",
					"ruleIndex": 1,
					"locations": [
						{
							"physicalLocation": {
								"artifactLocation": {
									"uri": "api/http/handler/websocket/proxy.go"
								},
								...
							}
						}
					]
				},
				...
			]
		}
	]
}

```
You can see 2 result items that have the same ruleId `G404 (CWE-338)` and different locations.

As for the dedup mechanics, I surely can re-write it using a map.

ccojocar: Thanks for this clarification and for improving the logic by using a map.

ccojocar (approved):

Issue #574, opened 2021-02-15
Generated SARIF is invalid 


pbalogh-sa:
### Summary 
The generated SARIF file is invalid using the Github actions.
```
Error: Unable to upload "gosec-results.sarif" as it is not valid SARIF:
- instance.runs[0].tool.driver.rules contains duplicate tem
```

The items that caused the error:
```
      },
      {
        "id": "G601 (CWE-118)",
        "name": "Implicit memory aliasing in for loop.",
        "shortDescription": {
          "text": "Implicit memory aliasing in for loop."
        },
        "fullDescription": {
          "text": "Implicit memory aliasing in for loop."
        },
        "help": {
          "text": "Implicit memory aliasing in for loop.\nSeverity: MEDIUM\nConfidence: MEDIUM\nCWE: https://cwe.mitre.org/data/definitions/118.html"
        },
        "properties": {
          "tags": [
            "CWE-118",
            "MEDIUM"
          ]
        },
        "defaultConfiguration": {
          "level": "error"
        }
      },
      {
        "id": "G601 (CWE-118)",
        "name": "Implicit memory aliasing in for loop.",
        "shortDescription": {
          "text": "Implicit memory aliasing in for loop."
        },
        "fullDescription": {
          "text": "Implicit memory aliasing in for loop."
        },
        "help": {
          "text": "Implicit memory aliasing in for loop.\nSeverity: MEDIUM\nConfidence: MEDIUM\nCWE: https://cwe.mitre.org/data/definitions/118.html"
        },
        "properties": {
          "tags": [
            "CWE-118",
            "MEDIUM"
          ]
        },
        "defaultConfiguration": {
          "level": "error"
        }
```

### Steps to reproduce the behavior
Set up gosec GitHub actions and try to upload the generated SARIF file.

### gosec version
securego/gosec@master

### Expected behavior
When the same issues are found, the IDs of the issues should be unique in the generated SARIF file.

### Actual behavior
The IDs of the issues can be the same.

felipe-avelar:
I think this issue is solve by #565, but I think docker hub was not updated...

pbalogh-sa:
Thaks

PR #575, opened 2021-02-16
Use uniq ID in the case of SARIF format

pbalogh-sa:
Signed-off-by: Peter Balogh <p.balogh.sa@gmail.com>

fixes: #574


pbalogh-sa:
Unit tests updates are needed.

pbalogh-sa:
duplicates: #565

PR #604, opened 2021-05-04
Migrate sonar types in a dedicated package

mmorel-35:
* Migrate sonar types in a dedicated package
* CWE field is not supported by sonar Report so I removed it


ccojocar (approved):
Thanks for this great refactoring!

PR #606, opened 2021-05-05
Define a report package with core and per format sub-packages

mmorel-35:
Others modifications like the use of GenerateReport function in every formatter file of each format


ccojocar (approved):
This is a great refactoring. Thanks!

PR #607, opened 2021-05-07
Refactor : Replace Cwe with cwe.Weakness

mmorel-35:


ccojocar (approved):

mmorel-35:
@ccojocar , It seems like only a part of the files are in the codecov report. Would you know why ?
I don't see report package and subpackages.

ccojocar:
> @ccojocar , It seems like only a part of the files are in the codecov report. Would you know why ?
> I don't see report package and subpackages. For example [here](https://codecov.io/gh/securego/gosec/tree/561cd724ce02d35230026d874fb1f35cf808d28d)

I am not sure what's the issue. The coverage is generated by the `make test-coverage`. Maybe we should look if there are any numbers in the report for those subpackages.

PR #610, opened 2021-05-07
Chore/golangci lint

mmorel-35:


mmorel-35:
I splitted some CWE descriptions but I think that they should be moved into a json file. Moreover, the templates (html and text) shall also be moved into template files. This way the the length of the lines wouldn't be a problem.

PR #611, opened 2021-05-08
ISSUE-600 | Adding -stdout -verbose Flag

### LINKED ISSUE #600: As a gosec user, I want my output is saved in a file as well as stdout.
For a CICD pipeline when pipeline is failing because of gosec results, I am getting either stdout or formatted file output. I am not able to see both set both the output formats at the same time.

By enabling the stdout while using any formatted output user can check the logs to quickly find out what is wrong and file output can be saved as an artifact.

I may going to file a PR  for this.

ShreyasSubhedar:
This PR adds up the `-stdout` flag to enable the results to be reported to stdout (console log) and at the same time it'll get stored inside the file provided by `-out` flag. 
It adds the `-verbose` flag to choose which format the stdout should follow while output format is present.
This helps in Devops CICD pipelines to store the generated report as an artifact and at the same time, makes it easily available in logs. 

fixes #600

### DIFF
--- a/README.md
+++ b/README.md
@@ -302,6 +302,12 @@ file. The output format is controlled by the `-fmt` flag, and the output file is
 $ gosec -fmt=json -out=results.json *.go
 ```
 
+Results will be reported to stdout as well as to the provided output file by `-stdout` flag.
+```bash
+# Write output in json format to results.json as well as stdout
+$ gosec -fmt=json -out=results.json -stdout *.go
+```
+
 **Note:** gosec generates the [generic issue import format](https://docs.sonarqube.org/latest/analysis/generic-issue/) for SonarQube, and a report has to be imported into SonarQube using `sonar.externalIssuesReportPaths=path/to/gosec-report.json`.
 
 ## Development
--- a/README.md
+++ b/README.md
@@ -302,8 +302,8 @@ file. The output format is controlled by the `-fmt` flag, and the output file is
 $ gosec -fmt=json -out=results.json *.go
 ```
 
-Results will be reported to stdout as well as to the provided output file by `-stdout` flag. The `-verbose` flag overrides the
-overrides the output format when stdout the results while saving them in the output file
+Results will be reported to stdout as well as to the provided output file by `-stdout` flag. The `-verbose` flag overrides the 
+output format when stdout the results while saving them in the output file
 ```bash
 # Write output in json format to results.json as well as stdout
 $ gosec -fmt=json -out=results.json -stdout *.go
--- a/README.md
+++ b/README.md
@@ -302,10 +302,14 @@ file. The output format is controlled by the `-fmt` flag, and the output file is
 $ gosec -fmt=json -out=results.json *.go
 ```
 
-Results will be reported to stdout as well as to the provided output file by `-stdout` flag.
+Results will be reported to stdout as well as to the provided output file by `-stdout` flag. The `-verbose` flag overrides the
+overrides the output format when stdout the results while saving them in the output file
 ```bash
 # Write output in json format to results.json as well as stdout
 $ gosec -fmt=json -out=results.json -stdout *.go
+
+# Overrides the output format to 'text' when stdout the results, while writing it to results.json
+$ gosec -fmt=json -out=results.json -stdout -verbose=text *.go
 ```
 
 **Note:** gosec generates the [generic issue import format](https://docs.sonarqube.org/latest/analysis/generic-issue/) for SonarQube, and a report has to be imported into SonarQube using `sonar.externalIssuesReportPaths=path/to/gosec-report.json`.
--- a/cmd/gosec/main.go
+++ b/cmd/gosec/main.go
@@ -193,6 +193,26 @@ func loadRules(include, exclude string) rules.RuleList {
 	return rules.Generate(filters...)
 }
 
+func getRootPaths(paths []string) []string {
+	rootPaths := []string{}
+	for _, path := range paths {
+		rootPath, err := gosec.RootPath(path)
+		if err != nil {
+			logger.Fatal(fmt.Errorf("failed to get the root path of the projects: %s", err))
+		}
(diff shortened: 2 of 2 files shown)


cmd/gosec/main.go
```diff
@@ -187,7 +190,7 @@ func loadRules(include, exclude string) rules.RuleList {
 	return rules.Generate(filters...)
 }
 
-func saveOutput(filename, format string, color bool, paths []string, issues []*gosec.Issue, metrics *gosec.Metrics, errors map[string][]gosec.Error) error {
```
mmorel-35, line 190: Nice! What do you think about replacing this parameter addition to the saveOutput method and replacing that with something like this
```go
rootpaths := getRootPaths(paths)
if filename ! = "" {
  saveReport(filename, ...)// this saves the report in a file
}
if filename == "" or stdout {
  printReport(... ) // this prints the report in the stdout 
}
```

ShreyasSubhedar: Absolutely! 
This is more adaptable.
adding the same!!

mmorel-35:
What about naming the flag `verbose`, if you use `stdout`, in implies that you  don't change the format and just redirect the stream

ShreyasSubhedar:
> What about naming the flag `verbose`, if you use `stdout`, it implies that you  don't change the format and just redirect the stream.
> And seems make more sense to print to the stdout before generating the report 

then 
`gosec ./... `
and 
`gosec -verbose ./... `
will be the same commands and user might raise an issue that verbosity is not being increased / decreased after using `verbose` flag.

which in case `-stdout` make more sense to me !

Happy to be corrected!

cmd/gosec/main.go
```diff
@@ -206,6 +209,14 @@ func saveOutput(filename, format string, color bool, paths []string, issues []*g
 		if err != nil {
 			return err
 		}
+		if stdout {
+			format = "text"
```
mmorel-35, line 213: IMHO format shouldn't be changed when `stdout` shall only means a redirect of the stream.

mmorel-35:
`gosec ./… ` and `gosec -stdout ./… ` will also be the same command and there is no apparent reason for the user to be less bothered by such behavior.
What I'm suggesting here is to add `-stdout` and `-verbose` when verbose is on the output format can be overriden, not when it's stdout

ShreyasSubhedar:
> `gosec ./… ` and `gosec -stdout ./… ` will also be the same command and there is no apparent reason for the user to be less bothered by such behavior.
> What I'm suggesting here is to add `-stdout` and `-verbose` when verbose is on the output format can be overriden, not when it's stdout

Got it! 
I believe, I have updated the logic, please add more suggestions!

cmd/gosec/main.go
```diff
@@ -363,8 +364,27 @@ func main() {
 	}
 
 	// Create output report
-	if err := saveOutput(*flagOutput, *flagFormat, color, flag.Args(), issues, metrics, errors); err != nil {
-		logger.Fatal(err)
+	rootPaths := []string{}
```
mmorel-35, line 367: Please create a getRootPaths(...)

ShreyasSubhedar: Okay

cmd/gosec/main.go
```diff
@@ -363,8 +364,27 @@ func main() {
...
+	for _, path := range flag.Args() {
+		rootPath, err := gosec.RootPath(path)
+		if err != nil {
+			logger.Fatal(fmt.Errorf("failed to get the root path of the projects: %s", err))
+		}
+		rootPaths = append(rootPaths, rootPath)
+	}
+	if *flagOutput == "" || *flagStdOut {
+		var fileFormat = *flagFormat
```
mmorel-35, line 376: same here with fileFormat = getPrintedFormat(...)

README.md
```diff
@@ -302,6 +302,16 @@ file. The output format is controlled by the `-fmt` flag, and the output file is
 $ gosec -fmt=json -out=results.json *.go
 ```
 
+Results will be reported to stdout as well as to the provided output file by `-stdout` flag. The `-verbose` flag overrides the
+overrides the output format when stdout the results while saving them in the output file
```
mmorel-35, line 306: "overrides the" is wrote twice?

ShreyasSubhedar: Opps fixing it

cmd/gosec/main.go
ShreyasSubhedar: Okay

cmd/gosec/main.go
```diff
@@ -291,7 +312,7 @@ func main() {
 
 	// Color flag is allowed for text format
 	var color bool
```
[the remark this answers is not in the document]
ShreyasSubhedar: True, color format should take it as an input from cmd, but what I see is, at present we are only supporting text format for color output, and not other formats.
So once we are can able to support other formats, then we can create:-
```golang
flagColor = flag.Bool("color", false, "enables the color output ")
```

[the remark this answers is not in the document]
ShreyasSubhedar: Also I am noticing that the results of the command :- 
`./gosec -out=results.txt ./...`
saves the results.txt as. :- 
```text
Results:
[1;36mSummary:[0m
   Files: 67
   Lines: 10465
   Nosec: 22
  Issues: [1;32m0[0m
``` 
I don't know, what will happen if we colored the sonarqube / xml output format & pass those files to any CI /analyzer tool.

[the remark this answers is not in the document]
mmorel-35: The color shall only be applied when the report is output in the console. But indeed colorisation is done by surrounding the text with characters that defines this colorisation.
If you do `gosec... > results.txt ` you'll end up with the same result.

IMHO, there shall be the flag you defined in your previous command.
The color flag shouldn't apply to file saved reports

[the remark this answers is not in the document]
ShreyasSubhedar: Agreed

ccojocar (approved):

PR #625, opened 2021-05-21
use go embed for templates

bakito:
Use go embed feature to make template maintenance easier.


bakito:
Just saw, that the package needs to run also against go 1.14 and 1.15.
I guess the PR might be closed then

ccojocar (commented):
go embed is not supported by Go 1.14 and 1.15. We maintain always support for two Go versions back from the current version. Thanks!

ccojocar:
We can add this improvement later when the support for 1.14 and 1.15 is not maintained. Are you fine to close it for now? Thanks for this contribution!

PR #636, opened 2021-06-01
Disable cache in golangci job

mmorel-35:


ccojocar (approved):
Thanks for fix this issue!

PR #637, opened 2021-06-01
Tidy up the dependencies

ccojocar:


mmorel-35:
That's weird that you have to do that after each dependencies updates. Isn't there a way to make it done by renovate automatically

ccojocar:
> That's weird that you have to do that after each dependencies updates. Isn't there a way to make it done by renovate automatically? Take a look at this https://docs.renovatebot.com/golang/

I agree, it would be cool to fix this stuff. It used to be fine a little while ago.

mmorel-35:
A solution might be to remove renovate json config file to trigger a new PR to restart configuration ?
This way we would see what we are doing. The created issue could then be pined and kept active as Babel did. What do you think ?

PR #638, opened 2021-06-01
Renovate/restart

mmorel-35:
This will force renovate to trigger a new configuration for the project
Something like that should suffice: 

```json
{
  "extends": [":preserveSemverRanges", "group:all", "schedule:weekly"],
  "lockFileMaintenance": {
    "commitMessageAction": "Update",
    "enabled": true,
    "extends": ["group:all", "schedule:weekly"]
  },
  "postUpdateOptions": ["gomodTidy", "gomodUpdateImportPaths"],
  "separateMajorMinor": false
}
```
The renovate dashboard will insure us that the configuration works as we expect


ccojocar:
@mmorel-35 It seems that you are removing the `renovate.json` file instead of updating it with the configuration proposed in the pull request description.

mmorel-35:
That's the point, after the removal, renovate will trigger an issue like a dashboard and a PR to assist the configuration

mmorel-35:
@ccojocar , There is a flag to enable dependency dashboard so we might not need to completely remove the renovate.json file

ccojocar (approved):

PR #643, opened 2021-06-01
Renovate/configure

mmorel-35:


mmorel-35:
@ccojocar, that should be good, now. The Dashboard /issue can be indefinitely kept open to supervise the PR of renovate. You can even pin it so it's easy to find it

ccojocar (approved):
Thanks, I'll do that.

PR #645, opened 2021-06-04
Fixes #644

### LINKED ISSUE #644: Format options for results on stdout not working
### Summary 
The `-fmt` flag does not work when the results are printed to `stdout` by default. 

Seems like we can only get the format we want on `stdout` only when we run `gosec -out output.json -stdout -verbose json ./...`. I suspect that https://github.com/securego/gosec/commit/a8b633f124f712e465ac2d7ee0e992b274cc40a6 changed the behavior.

Run `gosec -fmt json ./...`.

### gosec version
v2.8.0

`go version go1.14.1 linux/amd64`

Ubuntu 20.04

### Expected behavior
A JSON object will be printed out on `stdout`.

### Actual behavior
The `text` output is printed to `stdout`.

mmorel-35:
Hi @wongherlung , @ShreyasSubhedar would you mind checking this fix ?

### DIFF
--- a/cmd/gosec/main.go
+++ b/cmd/gosec/main.go
@@ -208,12 +208,13 @@ func getRootPaths(paths []string) []string {
 	return rootPaths
 }
 
+// If verbose is defined it overwrites the defined format
+// Otherwise the actual format is used
 func getPrintedFormat(format string, verbose string) string {
-	fileFormat := format
-	if format != "" && verbose != "" {
-		fileFormat = verbose
+	if verbose != "" {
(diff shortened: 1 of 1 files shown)


wongherlung:
LGTM! Thank you for the fix.

ShreyasSubhedar:
good catch

wongherlung (approved):

ShreyasSubhedar (approved):

ccojocar (approved):

Issue #657, opened 2021-06-24
Proposal: Closing files


elgohr:
Based on https://www.joeshaw.org/dont-defer-close-on-writable-files/ , I would like to add a new rule.
Would this be welcome as a PR or is behaviour like this out of scope?

ccojocar:
Hi @elgohr , There is already a rule which catches this issue: https://github.com/securego/gosec/blob/master/rules/bad_defer.go. Did you have a look at it? I think was based on the blog post mentioned above. Thanks

elgohr:
Maybe I should write another test for this role then. This rule will also be ok, when the error from close is handled in the defer function (even if the file is not synced with the filesystem).

ccojocar:
It would be great if you could write another test. I will be happy to merge it. Thanks!

PR #661, opened 2021-07-02
Add G307 sample code.

elgohr:
fixes https://github.com/securego/gosec/issues/657

The sample should reflect a defered close that leads to data loss.
Due to IDE auto-complete people tend to at least log errors, but not really care about handling.

To my point of view this could be fixed by seeing a deferred `f.Close()`, without a correspondig `f.Sync()`, generally as a bad pattern.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,8 +1,0 @@
-# Default ignored files
-/shelf/
-/workspace.xml
-# Datasource local storage ignored files
-/dataSources/
-/dataSources.local.xml
-# Editor-based HTTP Client requests
-/httpRequests/
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,8 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Datasource local storage ignored files
+/dataSources/
+/dataSources.local.xml
+# Editor-based HTTP Client requests
+/httpRequests/
--- a/.idea/gosec.iml
+++ b/.idea/gosec.iml
@@ -1,9 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<module type="WEB_MODULE" version="4">
-  <component name="Go" enabled="true" />
-  <component name="NewModuleRootManager">
-    <content url="file://$MODULE_DIR$" />
-    <orderEntry type="inheritedJdk" />
-    <orderEntry type="sourceFolder" forTests="false" />
-  </component>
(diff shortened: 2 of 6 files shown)


ccojocar (approved):

ccojocar:
@elgohr It seems that the test is failing. Could you please have a look? Thanks!

elgohr:
As said, I wrote the test. Do you want to discuss on the implementation or should I just give it a try?

ccojocar:
We can discuss on slack if you have questions. Otherwise you can give it a try, and I will review it when is ready.

elgohr:
I took a few tries and all ended up in turning G307 around and putting os.File in the center, instead of defer.
I don't think that's a good thing for the future of that rule. At the moment I don't see an easy to read solution for this, if we keep the structure as is... Do you have an idea?

ccojocar:
@elgohr Could you share what you have so far? I can have a look at your current implementation. Thanks

elgohr:
So this would be an idea for implementing - without turning the whole thing around

ccojocar (commented):
It looks good!

ccojocar:
It seems that there are some lint warnings to fix. Could you please have a look? Thanks

elgohr:
So I fixed the linting and added an example that I would like to discuss.

https://github.com/elgohr/gosec/commit/b1c5f2cd89c5efd979872085f80be5913198b1a8#diff-ac25ed6050bbf7e77dd739a9fc894d1ed3f2ab52dbf6a9b2b582577a91da1204R2061
This should not return an error, as it's savely synced.
Nevertheless it can't be tested, as the `Match` only gets defers...

ccojocar:
> So I fixed the linting and added an example that I would like to discuss.

Thanks for fixing the linting warnings. 

> This should not return an error, as it's savely synced.
Nevertheless it can't be tested, as the Match only gets defers..

I think the AST tree has to be traversed further into the defer statement in order to verify if the error is checked. `gosec` doesn't really have call graph tracing or taint analysis support. Do you want to work on this specific improvement for defer statement?

ccojocar:
The test is failing as you mentioned above. You can also remove this test in order for us to merge the pull request and contribute the improvement in a follow up pull request. What do you think?

elgohr:
So I rebased to upstream (that's why force push) and removed the failing test (for now).

ccojocar:
Thanks for following! I removed some IDEA related files which I think you committed accidentally.

PR #667, opened 2021-07-18
Use version `^v0` instead of `latest` for cyclonedx-gomod

nscuro:
To avoid breaking the build when cyclonedx-gomod introduces breaking changes in a new major version. See https://github.com/CycloneDX/gh-gomod-generate-sbom/releases/tag/v0.3.0

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
           json: true
           output: bom.json
           resolve-licenses: true
-          version: latest
+          version: ^v0
       - name: Release Binaries
         uses: goreleaser/goreleaser-action@v2
         with:


mmorel-35:
Nice! 
Would it also be pertinent to have the file named gosec_2.8.1_bom.json as the rest of the files are named in the release?

nscuro:
This comes down to personal preference, really. Most projects just use `bom.xml` or `bom.json`, but there's no official naming convention (ref https://github.com/aquasecurity/trivy/pull/1076#issuecomment-864575035).

mmorel-35:
I see. @ccojocar, what is your position on that matter then?

ccojocar (approved):
This makes sense.

PR #676, opened 2021-08-04
display rule it's in html report

bakito:
This PR adds rule ID and CWE reference to the html report.

### DIFF
--- a/report/html/template.go
+++ b/report/html/template.go
@@ -92,7 +92,7 @@ const templateContent = `
           <div className="columns">
               <div className="column is-three-quarters">
                 <strong className="break-word">{ this.props.data.file } (line { this.props.data.line })</strong>
-                <p>{ this.props.data.details }</p>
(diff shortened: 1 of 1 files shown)


report/html/template.go
```diff
@@ -92,7 +92,7 @@ const templateContent = `
           <div className="columns">
               <div className="column is-three-quarters">
                 <strong className="break-word">{ this.props.data.file } (line { this.props.data.line })</strong>
-                <p>{ this.props.data.details }</p>
+                <p>{this.props.data.rule_id} (CWE-{this.props.data.cwe.id}): { this.props.data.details }</p>
```
mmorel-35, line 95: This shall do the trick
```suggestion
                <p>{this.props.data.rule_id} ({this.props.data.cwe.SprintID()}): { this.props.data.details }</p>
```

bakito: SprintID() is not a function of the JavaScript data object

mmorel-35: You're right ! I thought this was a template with direct access to go functions

ccojocar (approved):

Issue #343, opened 2019-08-22
Issue: Gosec doesn't use context besides the function body


MVrachev:
### Summary 
The problem I encountered is when I am scanning this code with Gosec: 
```
package main

import (
	"log"
	"os/exec"
)

func RunCmd(command string) {
	cmd := exec.Command(command, "5")
	err := cmd.Start()
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Waiting for command to finish...")
	err = cmd.Wait()
}

func main() {
	RunCmd("sleep")
}
```
Gosec thinks there is command injection and I launched the process with a variable but the command variable in the function `RunCmd` is nothing but a hardcoded string. 

I get an error from Gosec: 

```
G204: Subprocess launched with variable (Confidence: HIGH, Severity: MEDIUM)
  > exec.Command(command, "5")
```

### gosec version
2.0.0
### Go version (output of 'go version')
`go version go1.12.6 darwin/amd64`
### Operating system / Environment
macOS Mojave
### Expected behavior
The problem I noticed here is that Gosec doesn't traverse the three back to the "RunCmd" function call but instead tries to resolve the command variable only in the context of the function.

### Actual behavior
It would be great if Gosec understands better the context not only of the function but the function call with its arguments.

nanikjava:
@MVrachev @ccojocar correct me if I'm wrong:

When there is a call such as this

`cmd := exec.Command(command, "5")`

`gosec` should flag this as valid because `command` is part of the parameter pass when calling function `RunCmd` ?

ccojocar:
`gosec` should not flag that the `command` variable is an insecure input injected by the user because is hardcoded to `sleep` when traversing back the call graph. The main issue is that `gosec` doesn't have the ability to track back to the source the parameters of a function through the call graph.

PR #677, opened 2021-08-05
fix: add more rules for G204

nanikjava:
Fix : #343 

This PR fix is to allow `exec.Command` to use parameter pass as part of function.


nanikjava:
@ccojocar Still WIP but interested to know your thoughts about this fix.

testutils/source.go
```diff
@@ -1369,7 +1369,7 @@ func RunCmd(command string) {
 
 func main() {
 	RunCmd("sleep")
-}`}, 1, gosec.NewConfig()},
+}`}, 0, gosec.NewConfig()},
```
ccojocar, line 1372: Could you please add another test?

nanikjava: @ccojocar added extra test

ccojocar (approved):

Issue #535, opened 2020-10-09
`nosec` doesn't work


kylidboy:
### Summary 
"gosec -nosec=true ./..."  does not work,
"gosec -exclude -nosec=true ./..."  works,

### gosec version
Installed by "go get"

### Go version (output of 'go version')
go version go1.15 linux/amd64

### Operating system / Environment
No LSB modules are available.
Distributor ID: Linuxmint
Description:    Linux Mint 20
Release:        20
Codename:       ulyana

### Expected behavior
"gosec -nosec=true ./..."  should work

### Actual behavior
"gosec -nosec=true ./..."  does not work,
"gosec -exclude -nosec=true ./..."  works,

ccojocar:
This looks like a bug to me which need a look.

gcmurphy:
I think I need to clarify the "doesn't work" part. As per the help:

```
  -nosec
    	Ignores #nosec comments when set
```

So if you supply the -nosec flag, or -nosec=true gosec will ignore any `#nosec` annotations and report any errors.
If you supply the arguments in this order: 

`-exclude -nosec` you will see the following log message: 

```
[gosec] 2020/11/02 20:53:49 Excluding rules: -nosec
```
Which means that the `-nosec` flag is being considered as a rule to exclude. This obviously won't exclude any actual rules. I'm not sure if this is a common issue with Go's command line argument parsing or something we've introduced..

@ccojocar thoughts?

ccojocar:
@gcmurphy Thanks for digging into it. I am not sure either 😄. It seems that something is not quite right with the command line arguments parsing.

nanikjava:
@kylidboy do you have sample to reproduce this issue ? want to test it out and see if I can put in a fix.

nanikjava:
@ccojocar what fix need to be put for this ? happy to help out in putting in the fix.

ccojocar:
@nanikjava I would try to have a look at the argument parsing to see what's going on when there are the combinations mentioned above:
- `-exclude -nosec=true` => should return a warning when `exclude` option has no arguments and `nosec` option should be set appropriately
-   `-nosec=true` => should work (nosec flag should be set)

nanikjava:
@ccojocar Looked at the code for `-exclude`. According to the doc https://pkg.go.dev/flag the following is considered as valid

`-exclude=Gxxx`
`-exclude -nosec=true`
`-exclude=`

[...]

PR #692, opened 2021-08-24
fix: create a separate type for flag that has validation

nanikjava:
Fix: #535 

Create a separate type to allow CLI validation using the `flag` package from the go library.

### DIFF
--- a/cmd/vflag/flag.go
+++ b/cmd/vflag/flag.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 )
 
-//ValidateFlag cli string type
+// ValidateFlag cli string type
 type ValidateFlag string
 
 func (f *ValidateFlag) String() string {
@@ -20,4 +20,3 @@ func (f *ValidateFlag) Set(value string) error {
 	}
 	return nil
 }
-
--- a/cmd/vflag/flag.go
+++ b/cmd/vflag/flag.go
@@ -2,21 +2,24 @@ package vflag
 
 import (
 	"errors"
-	"fmt"
 	"strings"
 )
 
-// ValidateFlag cli string type
-type ValidateFlag string
+// ValidatedFlag cli string type
+type ValidatedFlag struct {
+	Value string
+}
 
-func (f *ValidateFlag) String() string {
-	return fmt.Sprint(*f)
+func (f *ValidatedFlag) String() string {
+	return f.Value
 }
 
 // Set will be called for flag that is of validateFlag type
-func (f *ValidateFlag) Set(value string) error {
+func (f *ValidatedFlag) Set(value string) error {
 	if strings.Contains(value, "-") {
-		return errors.New("")
+		return errors.New("flag value cannot start with -")
 	}
+
+	f.Value = value
 	return nil
 }
--- a/cmd/vflag/flag.go
+++ b/cmd/vflag/flag.go
@@ -1,4 +1,4 @@
-package validate
+package vflag
 
 import (
 	"errors"
@@ -6,15 +6,15 @@ import (
 	"strings"
 )
 
-// ValidateFlag used for flag cli string type
-type ValidatedFlag string
+//ValidateFlag cli string type
+type ValidateFlag string
 
-func (f *ValidatedFlag) String() string {
+func (f *ValidateFlag) String() string {
 	return fmt.Sprint(*f)
 }
 
 // Set will be called for flag that is of validateFlag type
-func (f *ValidatedFlag) Set(value string) error {
+func (f *ValidateFlag) Set(value string) error {
 	if strings.Contains(value, "-") {
 		return errors.New("")
 	}
--- a/flag_test.go
+++ b/flag_test.go
@@ -1,0 +1,42 @@
+package gosec_test
+
+import (
+	"flag"
+	"os"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/securego/gosec/v2/cmd/vflag"
+)
+
+var _ = Describe("Cli", func() {
+	Context("vflag test", func() {
+		It("value must be empty as parameter value contains invalid character", func() {
+			os.Args = []string{"gosec", "-test1=-incorrect"}
+			f := vflag.ValidatedFlag{}
+			flag.Var(&f, "test1", "")
+			flag.CommandLine.Init("test1", flag.ContinueOnError)
+			flag.Parse()
+			Expect(flag.Parsed()).Should(Equal(true))
+			Expect(f.Value).Should(Equal(``))
+		})
+		It("value must be empty as parameter value contains invalid character without equal sign", func() {
+			os.Args = []string{"gosec", "-test2= -incorrect"}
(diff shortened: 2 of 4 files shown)


nanikjava:
@ccojocar let me know you think.

cmd/vflag/flag.go
```diff
@@ -0,0 +1,22 @@
...
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// ValidateFlag cli string type
+type ValidateFlag string
+
+func (f *ValidateFlag) String() string {
+	return fmt.Sprint(*f)
```
ccojocar, line 13: `*f` is already a string, `fmt.Sprint` call is not required.

ccojocar: Please could you remove it?

cmd/vflag/flag.go
```diff
@@ -0,0 +1,22 @@
+package vflag
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+)
+
+// ValidateFlag cli string type
+type ValidateFlag string
```
ccojocar, line 10: I would call this type `ValidatedFlag`.

cmd/vflag/flag.go
```diff
@@ -0,0 +1,22 @@
...
+
+func (f *ValidateFlag) String() string {
+	return fmt.Sprint(*f)
+}
+
+// Set will be called for flag that is of validateFlag type
+func (f *ValidateFlag) Set(value string) error {
+	if strings.Contains(value, "-") {
+		return errors.New("")
```
ccojocar, line 19: I would add some message to this error. e.g. `Flag value cannot start with -`.

cmd/vflag/flag.go
```diff
@@ -0,0 +1,22 @@
...
+	}
```
ccojocar, line 20: Why is not the `value` stored in the `f` receiver such that later can be retrieved by the `String` method?

ccojocar (changes requested):
Could you add please a small unit test for ValidateFlag's methods? Thanks

cmd/vflag/flag.go
```diff
@@ -0,0 +1,26 @@
...
+}
+
+func (f *ValidatedFlag) String() string {
+	return f.Value
+}
+
+// Set will be called for flag that is of validateFlag type
+func (f *ValidatedFlag) Set(value string) error {
+	if strings.Contains(value, "-") {
+		return errors.New(f.Error)
```
ccojocar, line 21: I am not sure if this error message needs to be configurable since the check which verifies that it doesn't start with `-`  is always the same.

ccojocar: I would just create an error here with the message `Flag value cannot start with -`

flag_test.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+var _ = Describe("Cli", func() {
+	Context("vflag test", func() {
+		It("value must be empty as parameter value contains invalid character", func() {
+			os.Args = []string{"gosec", "-test1=-incorrect"}
+			f := vflag.ValidatedFlag{Error: "Flag value cannot start with -."}
+			flag.Var(&f, "test1", "")
+			flag.CommandLine.Init("test1", flag.ContinueOnError)
+			flag.Parse()
```
ccojocar, line 19: We should check if the `flag.Parse` returns an error.

flag_test.go
```diff
@@ -0,0 +1,40 @@
...
+			Expect(f.Value).Should(Equal(``))
+		})
+		It("value must be empty as parameter value contains invalid character without equal sign", func() {
+			os.Args = []string{"gosec", "-test2= -incorrect"}
+			f := vflag.ValidatedFlag{Error: "Flag value cannot start with -."}
+			flag.Var(&f, "test2", "")
+			flag.CommandLine.Init("test2", flag.ContinueOnError)
+			flag.Parse()
```
ccojocar, line 27: The same here.

flag_test.go
```diff
@@ -0,0 +1,40 @@
...
+			Expect(f.Value).Should(Equal(``))
+			Expect(f.Error).Should(Equal(`Flag value cannot start with -.`))
+		})
+		It("value must not be empty as parameter value contains valid character", func() {
+			os.Args = []string{"gosec", "-test3=incorrect"}
+			f := vflag.ValidatedFlag{Error: "Flag value cannot start with -."}
+			flag.Var(&f, "test3", "")
+			flag.CommandLine.Init("test3", flag.ContinueOnError)
+			flag.Parse()
```
ccojocar, line 36: We should check that `flag.Parse` doesn't return an error.

flag_test.go
```diff
@@ -0,0 +1,40 @@
...
+			os.Args = []string{"gosec", "-test2= -incorrect"}
+			f := vflag.ValidatedFlag{Error: "Flag value cannot start with -."}
+			flag.Var(&f, "test2", "")
+			flag.CommandLine.Init("test2", flag.ContinueOnError)
+			flag.Parse()
+			Expect(f.Value).Should(Equal(``))
+			Expect(f.Error).Should(Equal(`Flag value cannot start with -.`))
+		})
+		It("value must not be empty as parameter value contains valid character", func() {
+			os.Args = []string{"gosec", "-test3=incorrect"}
```
ccojocar, line 32: I would use a consistent value for `test3` flag which reflects the test case purpose. For instance you can change `-test3=incorrect` with `-test3=correct`.

flag_test.go
```diff
@@ -0,0 +1,40 @@
...
+			Expect(f.Value).Should(Equal(``))
+		})
+		It("value must be empty as parameter value contains invalid character without equal sign", func() {
+			os.Args = []string{"gosec", "-test2= -incorrect"}
+			f := vflag.ValidatedFlag{Error: "Flag value cannot start with -."}
+			flag.Var(&f, "test2", "")
+			flag.CommandLine.Init("test2", flag.ContinueOnError)
+			flag.Parse()
+			Expect(f.Value).Should(Equal(``))
+			Expect(f.Error).Should(Equal(`Flag value cannot start with -.`))
```
ccojocar, line 29: There is no point to check here the `f.Error`.

ccojocar (changes requested):

nanikjava:
@ccojocar PR updated. Thanks.

ccojocar (commented):
@nanikjava Thanks for fixing the review comments. Just a few small things to fix before merging, especially I would like to see that we check the error returned by flag.Parse in the tests. Thanks

flag_test.go
nanikjava: From the API the `flag.Parse()` does not return anything https://cs.opensource.google/go/go/+/refs/tags/go1.16:src/flag/flag.go;l=1020 or am I looking it the wrong one ?

flag_test.go
nanikjava: Done

flag_test.go
ccojocar: You are right. I looked at https://pkg.go.dev/flag#FlagSet.Parse.

ccojocar (approved):

PR #715, opened 2021-10-18
Use a better naming for the variable

rleungx:
This PR uses `value` to replace `severity` in `convertToScore` since both severity and confidence will use this function.


rleungx:
@ccojocar PTAL, thanks.

PR #719, opened 2021-11-03
add openssh to docker image

jaffee:
I'm trying to scan a project which has dependencies which are private projects. When Go tries to fetch the dependencies it normally uses HTTPS, but that doesn't work if they're private (terminal prompts disabled, can't enter username/password). So you do this little trick with git configuration to get Go to fetch dependencies over ssh: `GIT_CONFIG_PARAMETERS=url.ssh://git@github.com/.insteadOf=https://github.com/`
unfortunately the docker image doesn't have ssh installed so this doesn't work :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN go mod download
 RUN make build-linux
 
 FROM golang:${GO_VERSION}-alpine 
-RUN apk add --update --no-cache ca-certificates bash git gcc libc-dev
+RUN apk add --update --no-cache ca-certificates bash git gcc libc-dev openssh
 ENV GO111MODULE on
 COPY --from=builder /build/gosec /bin/gosec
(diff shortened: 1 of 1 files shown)


ccojocar:
@jaffee How much does the image size increases when adding openssh?

jaffee:
@ccojocar if I'm doing this right (`make image`) it's about a 1% increase (5MB)

```
securego/gosec                                                     v2.9.1-3-g52de346   30f30e1248ff   9 seconds ago    451MB
securego/gosec                                                     v2.9.1-6-g55c6cea   39bdb43ba13d   34 minutes ago   446MB
```

jaffee:
just rebased on master, but image size didn't change

ccojocar:
@jaffee Thanks for clarification. It looks good to me. I'll merge it in.

PR #735, opened 2021-12-06
Suppression support

### LINKED ISSUE #723: Add suppression information in the output of GoSec
### Summary 

Violations can be suppressed by inline comments and external configurations in GoSec. We propose to add a `-track-suppressions=true` CLI option for adding suppression information, including its kind and justification, in the output of GoSec.

The code below shows the example in the format of SARIF:

```json
"results": [
  {
    "level": "error",
    "locations": [
      {
        "physicalLocation": {
          "artifactLocation": {
            "uri": "D:\\gotest\\hello.go"
          },
          "region": {
            "endColumn": 2,
            "endLine": 7,
            "snippet": {
              "text": "6: \tusername := \"admin\"\n7: \tpassword := \"f62e5bcda4fae4f82370da0c6f20697b8f8447ef\"\n8: \tfmt.Println(\"Doing something with: \", username, password)\n"
            },
            "sourceLanguage": "go",
            "startColumn": 2,
            "startLine": 7
          }
        }
      }
    ],
    "message": {
      "text": "Potential hardcoded credentials"
    },
    "ruleId": "G101",
    "suppressions": [
      {
        "kind": "inSource",
        "justification": "Sample justification message."
      }
    ],
  }
]
```

### Motivation

GoSec is to inspect source code for security problems by scanning the Go AST. When codes violate one or more defined rules, GoSec will report errors.

In our scenarios, when violations are suppressed, Security Development Lifecycle (SDL) tools, such as GoSec, are expected to export suppression information, including its kind and justification. We could take advantage of this info to generate corresponding signals for auditing purposes.

### Proposed Solution

1. Keep the current behavior of GoSec, i.e., issues will not be shown on the command line if the related rules are suppressed, unless `-nosec=true` is used.
2. When the new CLI option `-trakc-suppressions=true` is used, the violations (or issues in GoSec) that suppressed by inline directive comments will not be ignored, and the rules that suppressed externally will run to generate issues (but will not report errors in CLI).
3. Add suppression info in the output of GoSec
    1. For inline suppressions, we generate suppression info where `kind` is `inSource` and `justification` is the text after two or more consecutive dashes in the directive comments.
    For example:
        ```golang
        /* #nosec G101 -- Justification example */
        password := "f62e5bcda4fae4f82370da0c6f20697b8f8447ef"
        ```
        ```go
        password := "f62e5bcda4fae4f82370da0c6f20697b8f8447ef" // #nosec -- Justification example
        ```
        Output:
        ```json
        {"kind": "inSource", "justification": "Justification example"}
        ```
    2. For external suppressions, we generate suppression info where `kind` is `external` and `justification` is a certain sentence such as “Globally suppressed”.
    For example:
        ```
        gosec -exclude=G101 -track-suppressions=true ./..
        ```
        Output:
        ```json
        {"kind": "external", "justification": "Globally suppressed"}
        ```

We are very glad to contribute to GoSec. Let me know if you have any ideas.

Yiwei-Ding:
Add a new feature to track suppression info of GoSec. Details described in #723 

fixes #723

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -84,7 +84,6 @@ type Analyzer struct {
 	excludeGenerated  bool
 	showIgnored       bool
 	trackSuppressions bool
-	ruleSuppressed    map[string]bool
 }
 
 // SuppressionInfo object is to record the kind and the justification that used
@@ -95,7 +94,7 @@ type SuppressionInfo struct {
 }
 
 // NewAnalyzer builds a new analyzer.
-func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressions bool, ruleSuppressedList map[string]bool, logger *log.Logger) *Analyzer {
+func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressions bool, logger *log.Logger) *Analyzer {
 	ignoreNoSec := false
 	if enabled, err := conf.IsGlobalEnabled(Nosec); err == nil {
 		ignoreNoSec = enabled
@@ -110,7 +109,7 @@ func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressio
 	return &Analyzer{
 		ignoreNosec:       ignoreNoSec,
 		showIgnored:       showIgnored,
-		ruleset:           make(RuleSet),
+		ruleset:           NewRuleSet(),
 		context:           &Context{},
 		config:            conf,
 		logger:            logger,
@@ -120,7 +119,6 @@ func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressio
 		tests:             tests,
 		excludeGenerated:  excludeGenerated,
 		trackSuppressions: trackSuppressions,
-		ruleSuppressed:    ruleSuppressedList,
 	}
 }
 
@@ -136,10 +134,10 @@ func (gosec *Analyzer) Config() Config {
 
 // LoadRules instantiates all the rules to be used when analyzing source
 // packages
-func (gosec *Analyzer) LoadRules(ruleDefinitions map[string]RuleBuilder) {
+func (gosec *Analyzer) LoadRules(ruleDefinitions map[string]RuleBuilder, ruleSuppressed map[string]bool) {
 	for id, def := range ruleDefinitions {
 		r, nodes := def(id, gosec.config)
-		gosec.ruleset.Register(r, nodes...)
+		gosec.ruleset.Register(r, ruleSuppressed[id], nodes...)
 	}
 }
 
@@ -395,7 +393,7 @@ func (gosec *Analyzer) Visit(n ast.Node) ast.Visitor {
 			suppressions, ignored = ignores[rule.ID()]
 		}
 		// Track external suppressions.
-		if gosec.ruleSuppressed[rule.ID()] {
+		if gosec.ruleset.IsRuleSuppressed(rule.ID()) {
 			ignored = true
 			suppressions = append(suppressions, SuppressionInfo{
 				Kind:          "external",
--- a/analyzer.go
+++ b/analyzer.go
@@ -84,6 +84,7 @@ type Analyzer struct {
 	excludeGenerated  bool
 	showIgnored       bool
 	trackSuppressions bool
+	ruleSuppressed    map[string]bool
 }
 
 // SuppressionInfo object is to record the kind and the justification that used
@@ -94,7 +95,7 @@ type SuppressionInfo struct {
 }
 
 // NewAnalyzer builds a new analyzer.
-func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressions bool, logger *log.Logger) *Analyzer {
+func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressions bool, ruleSuppressedList map[string]bool, logger *log.Logger) *Analyzer {
 	ignoreNoSec := false
 	if enabled, err := conf.IsGlobalEnabled(Nosec); err == nil {
 		ignoreNoSec = enabled
@@ -119,6 +120,7 @@ func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressio
 		tests:             tests,
 		excludeGenerated:  excludeGenerated,
 		trackSuppressions: trackSuppressions,
+		ruleSuppressed:    ruleSuppressedList,
 	}
 }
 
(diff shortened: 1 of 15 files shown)


ccojocar (approved):
Thanks for this great contribution! It looks good. Please could you add a section in the README with some documentation for this feature. Thanks again!

report/sarif/formatter.go
```diff
@@ -199,3 +199,15 @@ func getSarifLevel(s string) Level {
 		return Note
 	}
 }
+
+func getSarifSuppressions(issue *gosec.Issue) []*Suppression {
```
mmorel-35, line 203: Please use `issue.Suppressions` as function parameter as nothing else is used

report/sarif/formatter.go
```diff
@@ -199,3 +199,15 @@ func getSarifLevel(s string) Level {
...
+	var suppressions []*Suppression
+	for _, s := range issue.Suppressions {
+		sarifSuppression := &Suppression{
```
mmorel-35, line 206: Please define a dedicated NewSuppression func in SARIF builder

analyzer.go
```diff
@@ -346,31 +369,39 @@ func (gosec *Analyzer) Visit(n ast.Node) ast.Visitor {
...
 	// Push the new set onto the stack.
-	gosec.context.Ignores = append([]map[string]bool{ignores}, gosec.context.Ignores...)
+	gosec.context.Ignores = append([]map[string][]SuppressionInfo{ignores}, gosec.context.Ignores...)
 
 	// Track aliased and initialization imports
 	gosec.context.Imports.TrackImport(n)
 
 	for _, rule := range gosec.ruleset.RegisteredFor(n) {
-		_, ignored := ignores[rule.ID()]
+		suppressions, ignored := ignores["*"]
```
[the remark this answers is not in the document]
Yiwei-Ding: Added comments for `*`. It works as the id of "all rules" and is transparent to users.

report/sarif/builder.go
```diff
@@ -1,5 +1,7 @@
 package sarif
 
+import "github.com/securego/gosec/v2"
```
mmorel-35, line 3: Please don't bring any gosec dependency in here

Yiwei-Ding: Totally agree. So did you mean I should put `NewSuppression` in `sarif/builder.go` but should not import gosec?

analyzer.go
```diff
@@ -307,31 +318,44 @@ func (gosec *Analyzer) ignore(n ast.Node) ([]string, bool) {
...
 				// Pull out the specific rules that are listed to be ignored.
 				re := regexp.MustCompile(`(G\d{3})`)
-				matches := re.FindAllStringSubmatch(group.Text(), -1)
+				matches := re.FindAllStringSubmatch(directive, -1)
 
-				// If no specific rules were given, ignore everything.
-				if len(matches) == 0 {
-					return nil, true
-				}
+				suppression := new(SuppressionInfo)
```
ccojocar, line 339: nit refactoring:
```Go
suppression := SuppressionInfo{
   Kind: "inSource",
   Justification: justification,
}
```
And then at line 346 and 352, you don't need to use a pointer

analyzer.go
```diff
@@ -307,31 +318,44 @@ func (gosec *Analyzer) ignore(n ast.Node) ([]string, bool) {
...
+				ignores := make(map[string]SuppressionInfo)
 				for _, v := range matches {
-					ignores = append(ignores, v[1])
+					ignores[v[1]] = *suppression
+				}
+
+				// If no specific rules were given, ignore everything.
+				// Use "*" as ignoring all rules.
+				if len(matches) == 0 {
+					ignores["*"] = *suppression
```
ccojocar, line 352: nit: extract "*" into a const with a meaningful name

issue.go
```diff
@@ -200,3 +201,9 @@ func NewIssue(ctx *Context, node ast.Node, ruleID, desc string, severity Score,
 		Cwe:        GetCweByRule(ruleID),
 	}
 }
+
+// WithSuppressions set the suppressions of the issue
+func (i *Issue) WithSuppressions(suppressions []SuppressionInfo) *Issue {
```
ccojocar, line 206: nit: here you don't need to return an issue pointer back since the pointer is already mutable unless you want to have this kind of style in the API when chaining multiple methods calls. What is the intend here?

report/formatter.go
```diff
@@ -77,3 +80,13 @@ func CreateReport(w io.Writer, format string, enableColor bool, rootPaths []stri
 	}
 	return err
 }
+
+func filterOutSuppressedIssues(issues []*gosec.Issue) []*gosec.Issue {
+	nonSuppressedIssues := make([]*gosec.Issue, 0)
```
ccojocar, line 85: nit refactoring: 
```
nonSuppressedIssues := []*gosec.Issue{}
```

report/sarif/builder.go
ccojocar: The SupressionInfo type is not logically related to the format (e.g. Sarif) in this case. In the same time, we tried to avoid importing the gosec package in here. All the types of the variable in the method are primitive for a reason. 

I would rather create a duplicated type in the formatter (e.g. Sarif) and populate it at the time when the method gets called by coping the data from `gosec.SuppressionInfo`.

cmd/gosec/main.go
```diff
@@ -147,14 +150,14 @@ func usage() {
 	fmt.Fprint(os.Stderr, "\n\nRULES:\n\n")
 
 	// sorted rule list for ease of reading
-	rl := rules.Generate()
-	keys := make([]string, 0, len(rl))
-	for key := range rl {
+	rl := rules.Generate(false)
```
ccojocar, line 153: I think here you need to use the `flagTrackSuppressions` instead of hardcoding `false`.

ccojocar (changes requested):

ccojocar:
@Yiwei-Ding Thanks for adding the section in the README file. I left a few review comments which it would be nice to have a them.

report/sarif/builder.go
mmorel-35: I mean that 
```go
func NewSuppression(kind string, justification string) *Suppression
```
goes in `sarif/builder.go`
while
```go
func buildSarifSuppressions(suppressions []gosec.SuppressionInfo) []*Suppression
```
stays in `sarif/formatter.go`

issue.go
Yiwei-Ding: The original idea was to follow the code style of gosec. I saw all `With-` methods return pointers.

ccojocar (approved):
Thanks for fixing the review comments!

mcauto:
when I update gosec.
then nosec is not working.

PR #738, opened 2021-12-13
Fix #736

### LINKED ISSUE #736: Some of the inline code annotation stopped working
### Summary 
The code annotations in the format:
```
/* #nosec G306 */
// Avoid nosec "Expect WriteFile permissions to be 0600 or less" error
some code
```
or annotations without justification stopped working

```
// #nosec G306 -- Avoid nosec "Expect WriteFile permissions to be 0600 or less" error
some code
```
works as expected.

run gosec with the following code

```
func main() {
	/* #nosec G306 */
	// Avoid nosec "Expect WriteFile permissions to be 0600 or less" error
	err := ioutil.WriteFile("test", []byte("someData"), 0644)
	if err != nil {}
}
```
### gosec version
Just installed the latest version, 
```
gosec  -version                                                                                                                                                                   
Version: dev
Git tag: 
Build date: 
```

go version go1.16.3 linux/amd64

Linux alexey-VB2 5.4.0-91-generic #102-Ubuntu SMP Fri Nov 5 16:31:28 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux

### Expected behavior
Do not print security warnings

### Actual behavior
Warnings are printed

Yiwei-Ding:
This PR is to fix #736.

### Root cause

Block comments will have an extra whitespace before the comment content that gotten from `group.Text()`. In the following table, `▯` indicates to a whitespace.

|comment|`group.Text()`|
|---|---|
|`// #nosec`|`/* #nosec */`|
|`#nosec`|`▯#nosec`|

In PR #735, `#nosec` or other `noSecAlternativeTag` is treated as a prefix of the annotation. There were not a test case that covered block comments and resulted in #736.

### Solution in this PR

Remove the whitespaces before comments with `strings.Trim()`.

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -319,16 +319,16 @@ func (gosec *Analyzer) ignore(n ast.Node) map[string]SuppressionInfo {
 		}
 
 		for _, group := range groups {
-
-			foundDefaultTag := strings.HasPrefix(group.Text(), noSecDefaultTag)
-			foundAlternativeTag := strings.HasPrefix(group.Text(), noSecAlternativeTag)
+			comment := strings.TrimSpace(group.Text())
+			foundDefaultTag := strings.HasPrefix(comment, noSecDefaultTag)
+			foundAlternativeTag := strings.HasPrefix(comment, noSecAlternativeTag)
 
 			if foundDefaultTag || foundAlternativeTag {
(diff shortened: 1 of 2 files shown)


ccojocar (approved):
Thanks for fixing so quick the issue!

PR #739, opened 2021-12-14
Remove space between `//` and `#nosec` in examples and internal use

scop:
Comments intended for machines to read do not have the space by convention.

I'm unable to find an "authoritative" reference at the moment, but it is mentioned here and there, and is in my experience the convention indeed. Anyway one example mention is at end of https://golangci-lint.run/usage/false-positives/#nolint

### DIFF
--- a/README.md
+++ b/README.md
@@ -277,7 +277,7 @@ AST so can apply to a whole block or more granularly to a single expression.
 
 ```go
 
-import "md5" // #nosec
+import "md5" //#nosec
 
 
 func main(){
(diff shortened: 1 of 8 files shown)


ccojocar:
@scop What's the reason for removing the space? What are you trying to solve? It should normally work with a space but it might be currently an issue due to a recent feature merged which broke some functionality. It is still under investigation. See https://github.com/securego/gosec/issues/736#issuecomment-993697667

scop:
As said, the reason is that it's a convention for comments for machines to be formatted that way in Go source. No other reason, does not fix a real problem that I'm aware, and both spaces and without should continue to be accepted. This is just to follow the convention and promote the variant without spaces.

ccojocar:
Where is this convention documented? The link from description looks to me more like a convention in golangcli than a general convention.

scop:
Still not what I was looking for, but `//go:` specific from https://dave.cheney.net/2018/01/08/gos-hidden-pragmas

> It’s also important to note that there is no space between the // and the go keyword. This is partly an accident of history, but it also makes it less likely to conflict with a regular comment.

ccojocar:
Thanks for clarifications! I will merge the pull request but I think that gosec should not be opinionated when it comes to the comments style. Each user should decide what fits her best.

ccojocar (approved):

scop:
Agree, thanks!

PR #752, opened 2022-01-01
Find more tempdirs

scop:
See individual commits for details.

### DIFF
--- a/rules/tempfiles.go
+++ b/rules/tempfiles.go
@@ -23,19 +23,41 @@ import (
 
 type badTempFile struct {
 	gosec.MetaData
-	calls gosec.CallList
-	args  *regexp.Regexp
+	calls       gosec.CallList
+	args        *regexp.Regexp
+	argCalls    gosec.CallList
+	nestedCalls gosec.CallList
 }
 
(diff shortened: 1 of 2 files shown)


ccojocar:
@scop Please could you have a look, the build is failing?

ccojocar (approved):

ccojocar:
@scop Please could you rebase your branch? The test should run now after upgrading ginkgo.

ccojocar (approved):

scop:
Rebased.

ccojocar:
@scop There is a test failing when the unit tests are executed second time for generating the coverage. Please could you have a look? It might be related to the existing temp files which doesn't get cleaned up after first tests execution.

scop:
At least the latter of the two commits (which I didn't originally intend to include here but decided eventually otherwise and hastily ported over from another branch) added a test case but failed to bump the number of expected hits, that's fixed now. Not sure why it'd manifest just as a coverage failure, as it's a clear bug. Anyway should be taken care of now.

ccojocar (approved):

PR #755, opened 2022-01-04
Fix false negatives for SQL injection in multi-line queries

### LINKED ISSUE #711: False negatives for SQL injection in multi-line queries
### Summary 

SQL injection vulnerabilities in multi-line queries are not detected if the WHERE clause and the vulnerable condition are on different lines. If they're on the same line, the vulnerability is spotted correctly.

Here G201 is not detected:
```
query := fmt.Sprintf(`
	SELECT
		name
	FROM
		users
	WHERE
		user_id = '%s' LIMIT 1`, "abcd")
```

But when I put the condition on the same line as the WHERE clause, G201 is correctly detected:
```
query := fmt.Sprintf(`
	SELECT
		name
	FROM
		users
```

### gosec version

v2.9.0

### Expected behavior

G201 is detected

### Actual behavior

G201 is not detected

kaiili:
Problem
fixes #711

Details :
    can be seen in #711.
    G201 regexp only match  space after the keyword.
    If there is a SQLStrFormat like the following, there will be problems with matching.
"""
SELECT * FROM t1 WHERE\n\r'%s';
"""
  
Solution
   Change regexp to match \n \r


ccojocar (approved):

ccojocar (changes requested):
@kaiili Please could you add a unit test? Also you can revert the changes from go.mod and go.sum files because are not related to the fix. Thanks

ccojocar (approved):

PR #757, opened 2022-01-09
Fix false negative for SQL injection when using DB.QueryRow.Scan()

### LINKED ISSUE #713: False negative for SQL injection when using DB.QueryRow.Scan()
### Summary 

If the call to Row.Scan() is in the same expression as DB.QueryRow() SQL injection vulnerability is not detected.

Works (G201 reported):
```
var name string
query := fmt.Sprintf(`
	SELECT
		name
	FROM
		users
row := db.QueryRow(query)
err := row.Scan(&name)
```

Does not work (nothing reported):
```
var name string
query := fmt.Sprintf(`
	SELECT
		name
	FROM
		users
err := db.QueryRow(query).Scan(&name)
```

### gosec version

v2.9.0

### Expected behavior

G201 is detected

### Actual behavior

G201 is not detected

kaiili:
### Problem
fixes #713

### Details :
can be seen in #713.
G201 regexp only match callexpr once.
If there is a sqlquery  like the following, there will be problems with matching.
```
func main() {
	var name string
	db, err := sql.Open("sqlite3", ":memory:")
	if err != nil {
		panic(err)
	}
	q := fmt.Sprintf("SELECT name FROM users where id = '%s'", os.Args[1])
	err = db.QueryRow(q).Scan(&name)
	if err != nil {
		panic(err)
	}
	defer db.Close()
}

```
### Solution
 Add  an extra check befor checkQuery.


ccojocar:
@kaiili Can you please rebase? Thanks

kaiili:
I rebase my fork repo , and  pull request on https://github.com/securego/gosec/pull/759.Sorry for wasting some time with this pr, as I'm not very proficient with git.

PR #759, opened 2022-01-11
Fix false negative for SQL injection when using DB.QueryRow.Scan()

### LINKED ISSUE #713: False negative for SQL injection when using DB.QueryRow.Scan()

kaiili:
Problem
fixes #713

Details :
can be seen in #713.
G201 regexp only match callexpr once.
If there is a sqlquery like the following, there will be problems with matching.

func main() {
	var name string
	db, err := sql.Open("sqlite3", ":memory:")
	if err != nil {
		panic(err)
	}
	q := fmt.Sprintf("SELECT name FROM users where id = '%s'", os.Args[1])
	err = db.QueryRow(q).Scan(&name)
	if err != nil {
		panic(err)
	}
	defer db.Close()
}

Solution
Add an extra check befor checkQuery.

### DIFF
--- a/rules/sql.go
+++ b/rules/sql.go
@@ -261,14 +261,12 @@ func (s *sqlStrFormat) Match(n ast.Node, ctx *gosec.Context) (*gosec.Issue, erro
 	switch stmt := n.(type) {
 	case *ast.AssignStmt:
 		for _, expr := range stmt.Rhs {
-			// when expr is CallExpr and Call.Fun.X is CallExpr ,check Call.Fun.X
-			if Call, ok := expr.(*ast.CallExpr); ok {
-				Selector, ok := Call.Fun.(*ast.SelectorExpr)
+			if call, ok := expr.(*ast.CallExpr); ok {
+				selector, ok := call.Fun.(*ast.SelectorExpr)
 				if !ok {
 					continue
 				}
-				sqlQueryCall, ok := Selector.X.(*ast.CallExpr)
-				// checkQuery when success get sqlQuery CallExpr and Selector contains Callexpr
+				sqlQueryCall, ok := selector.X.(*ast.CallExpr)
 				if ok && s.ContainsCallExpr(sqlQueryCall, ctx) != nil {
 					issue, err := s.checkQuery(sqlQueryCall, ctx)
 					if err == nil && issue != nil {
--- a/rules/sql.go
+++ b/rules/sql.go
@@ -261,6 +261,21 @@ func (s *sqlStrFormat) Match(n ast.Node, ctx *gosec.Context) (*gosec.Issue, erro
 	switch stmt := n.(type) {
 	case *ast.AssignStmt:
 		for _, expr := range stmt.Rhs {
+			// when expr is CallExpr and Call.Fun.X is CallExpr ,check Call.Fun.X
+			if Call, ok := expr.(*ast.CallExpr); ok {
+				Selector, ok := Call.Fun.(*ast.SelectorExpr)
+				if !ok {
+					continue
+				}
+				sqlQueryCall, ok := Selector.X.(*ast.CallExpr)
+				// checkQuery when success get sqlQuery CallExpr and Selector contains Callexpr
+				if ok && s.ContainsCallExpr(sqlQueryCall, ctx) != nil {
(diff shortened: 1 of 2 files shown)


rules/sql.go
```diff
@@ -261,6 +261,21 @@ func (s *sqlStrFormat) Match(n ast.Node, ctx *gosec.Context) (*gosec.Issue, erro
 	switch stmt := n.(type) {
 	case *ast.AssignStmt:
 		for _, expr := range stmt.Rhs {
+			// when expr is CallExpr and Call.Fun.X is CallExpr ,check Call.Fun.X
+			if Call, ok := expr.(*ast.CallExpr); ok {
```
ccojocar, line 265: please could you rename the `Call` to `call`, normally the local variables are not capitalised

rules/sql.go
```diff
@@ -261,6 +261,21 @@ func (s *sqlStrFormat) Match(n ast.Node, ctx *gosec.Context) (*gosec.Issue, erro
...
+				Selector, ok := Call.Fun.(*ast.SelectorExpr)
```
ccojocar, line 266: `Selector` => rename to `selector`

testutils/source.go
```diff
@@ -1189,6 +1189,75 @@ func main(){
 		panic(err)
 	}
 	defer rows.Close()
+<<<<<<< HEAD
```
ccojocar, line 1192: This is a left over from git rebase. Please could you remove it?

testutils/source.go
```diff
@@ -1189,6 +1189,75 @@ func main(){
...
+	if err != nil {
+		panic(err)
+	}
+	q := fmt.Sprintf("SELECT name FROM users where id = '%s'", os.Args[1])
+	err = db.QueryRow(q).Scan(&name)
+	if err != nil {
+		panic(err)
+	}
+	defer db.Close()
+=======
```
ccojocar, line 1259: left over from git rebase

testutils/source.go
```diff
@@ -1189,6 +1189,75 @@ func main(){
...
+>>>>>>> 58058af0c8275f11249a71f18fc548bbd7b97ccc
```
ccojocar, line 1260: The same this one is from git rebase.

rules/sql.go
```diff
@@ -261,6 +261,21 @@ func (s *sqlStrFormat) Match(n ast.Node, ctx *gosec.Context) (*gosec.Issue, erro
 	switch stmt := n.(type) {
 	case *ast.AssignStmt:
 		for _, expr := range stmt.Rhs {
+			// when expr is CallExpr and Call.Fun.X is CallExpr ,check Call.Fun.X
```
ccojocar, line 264: Please remove this comment since this information can be read directly from the code

rules/sql.go
```diff
@@ -261,6 +261,21 @@ func (s *sqlStrFormat) Match(n ast.Node, ctx *gosec.Context) (*gosec.Issue, erro
...
+			if Call, ok := expr.(*ast.CallExpr); ok {
+				Selector, ok := Call.Fun.(*ast.SelectorExpr)
+				if !ok {
+					continue
+				}
+				sqlQueryCall, ok := Selector.X.(*ast.CallExpr)
+				// checkQuery when success get sqlQuery CallExpr and Selector contains Callexpr
```
ccojocar, line 271: Please remove this comment, this info can be read directly from the code

ccojocar (changes requested):
Thank for rebasing the pull request. I left a few review comments. It would be great if you could have a look at them. Please could you also check the tests, there is one test failing. Thanks again!

PR #778, opened 2022-02-12
feat: add concurrency option to parallelize package loading

kruskall:
This PR adds support for running package loading in parallel due to `gosec.load` being a major bottleneck for big projects. 

The concurrency option defaults to the number of available CPUs and tests are using a concurrency value of 1.

## Benchmarks:

Benchmarks of a couple of GitLab projects using [hyperfine](https://github.com/sharkdp/hyperfine/):
___

#### before:

```
Benchmark 1: /path/to/gosec -exclude-generated ./...
Time (mean ± σ):     11.384 s ±  0.227 s    [User: 28.443 s, System: 8.731 s]
Range (min … max):   10.974 s … 11.793 s    10 runs

Warning: Ignoring non-zero exit code.
```

#### after:

```
Benchmark 1: /path/to/gosec -exclude-generated ./...
Time (mean ± σ):      2.292 s ±  0.078 s    [User: 23.965 s, System: 8.732 s]
Range (min … max):    2.230 s …  2.469 s    10 runs

Warning: Ignoring non-zero exit code.
```

#### before:

```
Benchmark 1: /path/to/gosec -exclude-generated ./...
Time (mean ± σ):     27.442 s ±  0.710 s    [User: 75.289 s, System: 20.270 s]
Range (min … max):   26.601 s … 28.497 s    10 runs

Warning: Ignoring non-zero exit code.
```

#### after:

```
Benchmark 1: /path/to/gosec -exclude-generated ./...
Time (mean ± σ):      5.953 s ±  1.047 s    [User: 67.921 s, System: 20.703 s]
Range (min … max):    5.237 s …  8.658 s    10 runs

Warning: Ignoring non-zero exit code.
```

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -29,6 +29,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"sync"
 
 	"golang.org/x/tools/go/packages"
 )
@@ -88,6 +89,7 @@ type Analyzer struct {
 	excludeGenerated  bool
 	showIgnored       bool
 	trackSuppressions bool
+	concurrency       int
 }
 
 // SuppressionInfo object is to record the kind and the justification that used
@@ -98,7 +100,7 @@ type SuppressionInfo struct {
 }
 
 // NewAnalyzer builds a new analyzer.
-func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressions bool, logger *log.Logger) *Analyzer {
+func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressions bool, concurrency int, logger *log.Logger) *Analyzer {
 	ignoreNoSec := false
 	if enabled, err := conf.IsGlobalEnabled(Nosec); err == nil {
 		ignoreNoSec = enabled
@@ -121,6 +123,7 @@ func NewAnalyzer(conf Config, tests bool, excludeGenerated bool, trackSuppressio
 		stats:             &Metrics{},
 		errors:            make(map[string][]Error),
 		tests:             tests,
+		concurrency:       concurrency,
 		excludeGenerated:  excludeGenerated,
 		trackSuppressions: trackSuppressions,
 	}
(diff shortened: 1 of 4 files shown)


ccojocar:
@kruskall Thanks for this improvement. I'll have a closer look at it soon. Please could you add some unit tests where the concurrency is more than `1`. I'll like to see that the concurrent case is covered by test.

analyzer.go
```diff
@@ -153,12 +156,48 @@ func (gosec *Analyzer) Process(buildTags []string, packagePaths ...string) error
...
+		pkgPath string
+		pkgs    []*packages.Package
+		err     error
+	}
+
+	results := make(chan result)
+	jobs := make(chan string)
+
+	var wg sync.WaitGroup
+	wg.Add(gosec.concurrency)
```
ccojocar, line 169: Normally the wait group is incremented in the loop where the workers are started.

analyzer.go
```diff
@@ -153,12 +156,48 @@ func (gosec *Analyzer) Process(buildTags []string, packagePaths ...string) error
...
+
+	worker := func(j chan string, r chan result) {
+		for s := range j {
+			packages, err := gosec.load(s, config)
```
ccojocar, line 173: It is only the load made concurrent. I would also try to make the `Check` concurrent after loading, in the same goroutine.

analyzer.go
```diff
@@ -153,12 +156,48 @@ func (gosec *Analyzer) Process(buildTags []string, packagePaths ...string) error
...
+			r <- result{pkgPath: s, pkgs: packages, err: err}
+		}
+		wg.Done()
+	}
+
+	for i := 0; i < gosec.concurrency; i++ {
+		go worker(jobs, results)
+	}
+
+	go func() {
```
ccojocar, line 183: If you move the Check in the goroutine, You won't need to spin up other go routine here.

analyzer.go
```diff
@@ -153,12 +156,48 @@ func (gosec *Analyzer) Process(buildTags []string, packagePaths ...string) error
...
+		for _, pkgPath := range packagePaths {
+			jobs <- pkgPath
+		}
+
+		close(jobs)
+	}()
+
+	go func() {
```
ccojocar, line 191: You can get rid of this goroutine if the Check is also concurrent and move the wg.Wait() at the end.

ccojocar (changes requested):

ccojocar:
`Check` cannot be made concurrent due to the shared state in the analyser.

analyzer.go
```diff
@@ -153,12 +156,48 @@ func (gosec *Analyzer) Process(buildTags []string, packagePaths ...string) error
...
+
+	worker := func(j chan string, r chan result) {
+		for s := range j {
+			packages, err := gosec.load(s, config)
+			r <- result{pkgPath: s, pkgs: packages, err: err}
+		}
+		wg.Done()
+	}
+
+	for i := 0; i < gosec.concurrency; i++ {
```
ccojocar, line 179: Please move here the wg.Add

analyzer.go
```diff
@@ -153,12 +156,48 @@ func (gosec *Analyzer) Process(buildTags []string, packagePaths ...string) error
...
+	}()
+
+	for r := range results {
+		if r.err != nil {
+			gosec.AppendError(r.pkgPath, r.err)
 		}
-		for _, pkg := range pkgs {
+		for _, pkg := range r.pkgs {
 			if pkg.Name != "" {
 				err := gosec.ParseErrors(pkg)
```
ccojocar, line 202: When a parse error is encountered, all the workers needs to be stopped in a clean way, also the goroutine waiting for the worker to finish. Currently they get sopped by the program exit and not before.

analyzer.go
kruskall: Done!

analyzer.go
kruskall: Good catch! I've added a quit channel to signal the goroutine to stop so this should be handled

kruskall:
> @kruskall Thanks for this improvement. I'll have a closer look at it soon. Please could you add some unit tests where the concurrency is more than `1`. I'll like to see that the concurrent case is covered by test.

@ccojocar Thank you for the feedback! I've added a test with a concurrency value of 32, a few more comments and made sure a goroutine would stop if there are no jobs left. Back to you :ping_pong:

ccojocar (approved):
Thanks for fixing the review comments. It looks good now!

PR #788, opened 2022-03-06
added recursive flag -r to skip specifying ./... path

### LINKED ISSUE #591: Add -recursive parameter
### Summary 
In order to recursively some directory user must append `./...` to target directory.   
It would be useful for automation if there's an optional `-recursive` parameter (like in bandit [example](https://github.com/PyCQA/bandit/blob/bcebfc8f5397b5f52d10a66f9bf3c894e5060d6a/bandit/cli/main.py#L148)) that could automatically append target dir with `./...`.

ArnPellesGit:
Fixes #591

### DIFF
--- a/cmd/gosec/main.go
+++ b/cmd/gosec/main.go
@@ -384,21 +384,16 @@ func main() {
 	excludedDirs := gosec.ExcludedDirsRegExp(flagDirsExclude)
 	var packages []string
 
-	if len(flag.Args()) == 0 {
-		path := "./..."
+	paths := flag.Args()
+	if len(paths) == 0 {
+		paths = append(paths, "./...")
+	}
+	for _, path := range paths {
 		pcks, err := gosec.PackagePaths(path, excludedDirs)
 		if err != nil {
 			logger.Fatal(err)
 		}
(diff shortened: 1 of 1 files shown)


cmd/gosec/main.go
```diff
@@ -380,13 +383,24 @@ func main() {
...
+
+	if len(flag.Args()) == 0 {
+		path := "./..."
 		pcks, err := gosec.PackagePaths(path, excludedDirs)
 		if err != nil {
 			logger.Fatal(err)
 		}
 		packages = append(packages, pcks...)
+	} else {
+		for _, path := range flag.Args() {
```
ccojocar, line 395: I would avoid this code duplication here. You can do something like:
```Go
paths := flag.Args()
if len(paths) == 0 {
  paths = append(paths, "./...")
}
for _, path := range paths {
  ...
}
```

ccojocar (changes requested):

cmd/gosec/main.go
ArnPellesGit: You're right, I'll fix that :)

ccojocar (approved):

PR #798, opened 2022-03-28
Fix G201 and G202 error messages 

rubensayshi:
This fixes G201 and G202 error messages not point to the same line as where you need to apply the nosec comment.

it changes

```
[file.go:267-277] - G201 (CWE-89): SQL string formatting (Confidence: HIGH, Severity: MEDIUM)
    266: 
  > 267: 	qry := fmt.Sprintf(`SELECT "blabla" FROM %s`, tableName)
```

to 

```
[file.go] - G201 (CWE-89): SQL string formatting (Confidence: HIGH, Severity: MEDIUM)
    278: 
  > 279: 	rows, err := db.QueryContext(ctx, qry)
```

But I'm not so sure this is a good idea ...  
It does point you in the right direction where you should put a `// #nosec` line, but it's less informative about the actual problem.

I think the best solution would be to have `NewIssue` accept a list of `ast.Node` and concat the code snippets together?  -> https://github.com/securego/gosec/pull/799

### DIFF
--- a/rules/sql.go
+++ b/rules/sql.go
@@ -95,7 +95,7 @@ func (s *sqlStrConcat) checkQuery(call *ast.CallExpr, ctx *gosec.Context) (*gose
 				if op, ok := op.(*ast.Ident); ok && s.checkObject(op, ctx) {
 					continue
 				}
-				return gosec.NewIssue(ctx, be, s.ID(), s.What, s.Severity, s.Confidence), nil
+				return gosec.NewIssue(ctx, query, s.ID(), s.What, s.Severity, s.Confidence), nil
 			}
 		}
 	}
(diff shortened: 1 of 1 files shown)


rules/sql.go
```diff
@@ -95,7 +95,7 @@ func (s *sqlStrConcat) checkQuery(call *ast.CallExpr, ctx *gosec.Context) (*gose
 				if op, ok := op.(*ast.Ident); ok && s.checkObject(op, ctx) {
 					continue
 				}
-				return gosec.NewIssue(ctx, be, s.ID(), s.What, s.Severity, s.Confidence), nil
+				return gosec.NewIssue(ctx, query, s.ID(), s.What, s.Severity, s.Confidence), nil
```
ccojocar, line 98: I am not sure if this is right. I will point to ` db.QueryContext` but the issue might be in the string.

ccojocar (commented):
What issues are you trying to solve. Please could you provide some examples? Thanks

ccojocar:
@rubensayshi Thanks for contributing this but I am going to close it for now.

rubensayshi:
hey @ccojocar , no worries ;)

PR #799, opened 2022-03-28
Add option to specify multiple nodes for Issue

rubensayshi:
builds on top of https://github.com/securego/gosec/pull/798 because that's where the need for specifying multiple nodes comes from.

If the maintainers like this change then I'll add some tests, but considering I have to figure out how to use the test utils I'd like a :+1: before investing that time.

I don't like the `NewIssueMultipleNodes` naming, but can't think of something better, except maybe:
```
func NewIssue(ctx *Context, node ast.Node, ruleID, desc string, severity Score, confidence Score, extraNodes ...ast.Node) *Issue {
```

but I think it varies heavily per person if you like that approach or not ;)

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -23,6 +23,7 @@ import (
 	"go/token"
 	"os"
 	"strconv"
+	"strings"
 
 	"github.com/securego/gosec/v2/cwe"
 )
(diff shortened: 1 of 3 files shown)


ccojocar (commented):
Why is this change needed? An issue should not point into two different places. I don't think this is a good idea.

ccojocar:
@rubensayshi Thanks for contributing this but I am going to close it for now.

PR #808, opened 2022-04-27
Add new checks, new possibilities

hackallcode:
It seems that this pull request is not ready, it can serve as a code for inspiration, new ideas.

These checks work, but controversial fixes were required to the modules loading algorithm.
However, checking for incorrect usage of switch works without this controversial solution.

New checks:

- usage of deprecated functions
- usage of switch without default block
- usage of Rat.SetString in math/big with an overflow error (CVE-2022-23772)

New possibilities:

- get declarations AST not of the current package;
- add possibility to grep files to analyze.


ccojocar:
Thanks for this contribution! 

> usage of deprecated functions
> usage of switch without default block

What are the potential security issues detected by this two new rules? They look to me a bit generic. gosec is mainly focused on security vulnerabilities.

> usage of Rat.SetString in math/big with an overflow error (https://github.com/advisories/GHSA-q99m-p7hq-5v4f)

Could you please extract this rule in a separate pull request?

hackallcode:
> Could you please extract this rule in a separate pull request?

Yes, I created #810 pull request

ccojocar:
Thanks for contributing this but I am going to close this pull request since the remaining changes here are not in scope of gosec.

PR #810, opened 2022-04-30
Add check for usage of Rat.SetString in math/big with an overflow error

hackallcode:
Rat.SetString in math/big in Go before 1.16.14 and 1.17.x before 1.17.7
has an overflow that can lead to Uncontrolled Memory Consumption.

It is the CVE-2022-23772.

### DIFF
--- a/README.md
+++ b/README.md
@@ -166,6 +166,7 @@ directory you can supply `./...` as the input argument.
 - G504: Import blocklist: net/http/cgi
 - G505: Import blocklist: crypto/sha1
 - G601: Implicit memory aliasing of items from a range statement
+- G602: Usage of Rat.SetString in math/big with an overflow (CVE-2022-23772)
 
 ### Retired rules
 
--- a/README.md
+++ b/README.md
@@ -145,6 +145,7 @@ directory you can supply `./...` as the input argument.
 - G110: Potential DoS vulnerability via decompression bomb
 - G111: Potential directory traversal
 - G112: Potential slowloris attack
+- G113: Usage of Rat.SetString in math/big with an overflow (CVE-2022-23772)
 - G201: SQL query construction using format string
 - G202: SQL query construction using string concatenation
 - G203: Use of unescaped data in HTML templates
@@ -166,7 +167,6 @@ directory you can supply `./...` as the input argument.
 - G504: Import blocklist: net/http/cgi
 - G505: Import blocklist: crypto/sha1
 - G601: Implicit memory aliasing of items from a range statement
-- G602: Usage of Rat.SetString in math/big with an overflow (CVE-2022-23772)
 
 ### Retired rules
 
--- a/analyzer.go
+++ b/analyzer.go
@@ -58,12 +58,14 @@ type Context struct {
 	Comments     ast.CommentMap
 	Info         *types.Info
 	Pkg          *types.Package
+	RawPkg       *packages.Package
 	PkgFiles     []*ast.File
 	Root         *ast.File
 	Config       Config
 	Imports      *ImportTracker
 	Ignores      []map[string][]SuppressionInfo
 	PassedValues map[string]interface{}
+	LoadPackages func(pkgPath string) ([]*packages.Package, error)
 }
 
 // Metrics used when reporting information about a scanning run.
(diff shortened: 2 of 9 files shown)


README.md
```diff
@@ -165,6 +165,7 @@ directory you can supply `./...` as the input argument.
 - G504: Import blocklist: net/http/cgi
 - G505: Import blocklist: crypto/sha1
 - G601: Implicit memory aliasing of items from a range statement
+- G602: Usage of Rat.SetString in math/big with an overflow (CVE-2022-23772)
```
ccojocar, line 168: I would change the ID of this rule to G113. There is already a similar rule for integer overflow.

analyzer.go
```diff
@@ -258,12 +262,32 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
 		}
 	}
 
-	// step 3/3 remove build tags from conf to proceed build correctly.
+	return packageFiles, nil
+}
+
+func (gosec *Analyzer) load(pkgPath string) ([]*packages.Package, error) {
```
ccojocar, line 268: Why do you need all these refactoring in the analyser? I would prefer that you only add the rule without changing the analyser since this is not in scope of this pull request. Please have a look at how is rule G109 implemented.

ccojocar (changes requested):
Thanks contributing this pull request. I would prefer that you add only the rule without refactoring the analyser. Please have a look at the rule G109. Thanks

analyzer.go
hackallcode: The new rule needs to get the source code of the called function, namely the Rat.SetString source code.

Prior to this patch, it was not possible to load the source code of a called function from within a rule.

Now in the context is the `LoadPackages` function, which refers to the load function. It won't work without refactoring, since it's harder to get the context from inside the rule.

README.md
hackallcode: Changed

analyzer.go
hackallcode: Also, if GetPkgAbsPath returned an error, this does not mean that the package is not present, so you can continue executing the code

ccojocar: `RawPkg *packages.Package` introduces a different level of abstraction in the analyser which is not required for this rule. Also this has performance issues in a large code base because you load all packages. 

You just need to check that the type of the variable where `SetString` is called is `big.Rat`. You can find an example on this rule https://github.com/securego/gosec/blob/34d144b3fa5d29af061157c1cd156a762f97f864/rules/slowloris.go#L51. 

I am not planning to merge the refactoring for this rule. Thanks for understanding!

rules/math_big_rat.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, nil
+	}
+
+	funcDecl, funcDeclErr := gosec.FindFuncDecl(callExpr, ctx)
+	if funcDeclErr != nil {
+		return nil, nil
+	}
+
+	for _, elem := range funcDecl.Body.List {
+		firstIf, firstName, err := getXName(elem)
```
ccojocar, line 55: You are introspecting the function and make assumptions about the code. This is completely not scaling as soon as the Go team  will change the function code. gosec doesn't do anything like this in any rule. 

Looking at the patch https://go-review.googlesource.com/c/go/+/379537/. I think you can just simplify the check a lot by only checking the type of the variable and in the warning message just mention the vulnerable Go version and CVE number.

The SetString on big.Rat is really specific method, I don't expect that is used in a lot of projects.

ccojocar:
Closing this due to inactivity. Thanks

rules/math_big_rat.go
hackallcode: Do you propose to issue regardless of the version of Go? Or still check the version using the runtime library?

analyzer.go
hackallcode: This was done in order not to check the runtime, but it's possible always output an issue

rules/math_big_rat.go
hackallcode: Yes, I was talking about the same library

hackallcode: I create new PR #819, because this one is closed, and I can't reopen it

PR #819, opened 2022-06-01
Add check for usage of Rat.SetString in math/big with an overflow error

hackallcode:
Rat.SetString in math/big in Go before 1.16.14 and 1.17.x before 1.17.7
has an overflow that can lead to Uncontrolled Memory Consumption.

It is the CVE-2022-23772.

Fix of #810

### DIFF
--- a/rules/math_big_rat.go
+++ b/rules/math_big_rat.go
@@ -8,6 +8,7 @@ import (
 
 type usingOldMathBig struct {
 	gosec.MetaData
+	calls gosec.CallList
 }
 
 func (r *usingOldMathBig) ID() string {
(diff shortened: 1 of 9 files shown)


rules/math_big_rat.go
```diff
@@ -0,0 +1,50 @@
...
+	if !callExprOk {
+		return nil, nil
+	}
+
+	packageName, callName, callInfoOk := gosec.GetCallInfo(callExpr, ctx)
+	if callInfoOk != nil {
+		return nil, nil
+	}
+
+	if packageName != "math/big.Rat" || callName != "SetString" {
```
ccojocar, line 28: Please could you the existing helper function to check this. Here is an example https://github.com/securego/gosec/blob/fb587c1d106addd2efc9d62c1ebc31a6caa9d746/rules/rsa.go#L35

ccojocar (changes requested):
It looks good, please could you use the helper function instead to check the call as mentioned in the comment. Thanks

rules/math_big_rat.go
hackallcode: I had to fix this function a little, because it could not work with functions belonging to structures

ccojocar (approved):

PR #824, opened 2022-06-17
Pin cosign-installer to `v2`

saschagrunert:
We now have tags available in the cosign-installer, which allows us to pin the latest release via `v2`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,7 @@ jobs:
       GO111MODULE: on
       ACTIONS_ALLOW_UNSECURE_COMMANDS: true
     steps:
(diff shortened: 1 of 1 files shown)


JAORMX (approved):

JAORMX:
@saschagrunert There is https://github.com/sigstore/cosign-installer/releases/tag/v2.4.0 ... wouldn't we want to pin to that instead?

saschagrunert:
> @saschagrunert There is https://github.com/sigstore/cosign-installer/releases/tag/v2.4.0 ... wouldn't we want to pin to that instead?

Yep, done :+1:

JAORMX (approved):

Issue #823, opened 2022-06-16
G112 is invalid if ReadTimeout is set


ainar-g:
### Summary 

G112 should not be triggered if `http.Server.ReadTimeout` is set while `http.Server.ReadHeaderTimeout` isn't.

### Steps to reproduce the behavior

```go
package main

import (
	"fmt"
	"net/http"
	"time"
)

func main() {
	fmt.Println(&http.Server{
		ReadTimeout:  1 * time.Second,
		WriteTimeout: 1 * time.Second,
		IdleTimeout:  1 * time.Second,
	})
}
```

### gosec version

```none
/home/ainar/go/bin/gosec: go1.18.3
        path    github.com/securego/gosec/v2/cmd/gosec
        mod     github.com/securego/gosec/v2    v2.12.0 h1:CQWdW7ATFpvLSohMVsajscfyHJ5rsGmEXmsNcsDNmAg=
        dep     github.com/google/uuid  v1.3.0  h1:t6JiXgmwXMjEs8VusXIJk2BXHsn+wx8BZdTaoZ5fu7I=
        dep     github.com/gookit/color v1.5.1  h1:Vjg2VEcdHpwq+oY63s/ksHrgJYCTo0bwWvmmYWdE9fQ=
        dep     github.com/nbutton23/zxcvbn-go  v0.0.0-20210217022336-fa2cb2858354      h1:4kuARK6Y6FxaNu/BnU2OAaLF86eTVhP2hjTB6iMvItA=
        dep     github.com/xo/terminfo  v0.0.0-20210125001918-ca9a967f8778      h1:QldyIu/L63oPpyvQmHgvgickp1Yw510KJOqX7H24mg8=
        dep     golang.org/x/mod        v0.6.0-dev.0.20220419223038-86c51ed26bb4        h1:6zppjxzCulZykYSLyVDYbneBfbaBIQPYMevg0bEwv2s=
        dep     golang.org/x/sys        v0.0.0-20220319134239-a9b59b0215f8      h1:OH54vjqzRWmbJ62fjuhxy7AxFFgoHN0/DPc/UrL8cAs=
        dep     golang.org/x/tools      v0.1.11 h1:loJ25fNOEhSXfHrpoGj91eCUThwdNX6u24rO1xnNteY=
        dep     gopkg.in/yaml.v2        v2.4.0  h1:D8xgwECY7CYvx+Y2n4sBz93Jn9JRvxdiyyo8CTfuKaY=
        build   -compiler=gc
[...]

PR #825, opened 2022-06-17
Fix ReadTimeout for G112

yunwei37:
fix issue #823:

if `http.Server.ReadTimeout` is set, G112 should not report error.

### DIFF
--- a/rules/slowloris.go
+++ b/rules/slowloris.go
@@ -35,7 +35,7 @@ func containsReadHeaderTimeout(node *ast.CompositeLit) bool {
 	for _, elt := range node.Elts {
 		if kv, ok := elt.(*ast.KeyValueExpr); ok {
 			if ident, ok := kv.Key.(*ast.Ident); ok {
-				if ident.Name == "ReadHeaderTimeout" {
+				if ident.Name == "ReadHeaderTimeout" || ident.Name == "ReadTimeout" {
 					return true
 				}
 			}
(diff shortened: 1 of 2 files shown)


rules/slowloris.go
```diff
@@ -35,7 +35,7 @@ func containsReadHeaderTimeout(node *ast.CompositeLit) bool {
 	for _, elt := range node.Elts {
 		if kv, ok := elt.(*ast.KeyValueExpr); ok {
 			if ident, ok := kv.Key.(*ast.Ident); ok {
-				if ident.Name == "ReadHeaderTimeout" {
+				if ident.Name == "ReadHeaderTimeout" || ident.Name == "ReadTimeout" {
```
ainar-g, line 38: Wouldn't that mean that a configuration where the programmer sets `ReadHeaderTimeout` but not `ReadTimeout` would be considered valid?  Because it seems like that would still allow the remote peer to slow down the server by sending headers quickly and then not sending the body.

That's technically not an error, since the documentation of `ReadHeaderTimeout` implies that that is precisely the intention behind this parameter, but I feel like a tool like `gosec` should at least warn the programmer about that.

ainar-g (commented):
Thanks for doing this so quickly!

rules/slowloris.go
ccojocar: I think that such an attack is technically more difficult to perform that an attack without any timeout. gosec warns the programmer to setup some timeout. I think is fair enough in this case.

ccojocar (approved):

PR #828, opened 2022-06-29
fix: filepaths with git anywhere in them being erroneously excluded

timcosta:
This fix addresses an issue where the default exclude aimed to exclude `.git` folders is erroneously excluding any file paths with `git` in them.

For example, given a path like `src/git-client/main.go` or `client-git/main.go` the current implementation fails to scan that directory. This causes `gosec` to error with `No packages found` when a folder being scanned only contains `.go` files in these erroneously matched paths.

I verified this by using regexpal.com and the strings `src/git-client/main.go` and `client-git/main.go` mentioned above. The existing exclude `.git` compiles to `([\\/])?.git([\\/])?` which causes the problem as the `.` isnt escaped. The fixed `\\.git/` compiles to `([\\/])?\.git\/([\\/])?` which escapes the leading `.` and adds the protection of the required trailing `/` indicating that it's a directory. This prevents file paths like `src/.git-client/main.go` from being excluded while leaving the behavior originally desired in #485 intact.

### DIFF
--- a/cmd/gosec/main.go
+++ b/cmd/gosec/main.go
@@ -306,9 +306,9 @@ func main() {
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "\nError: failed to exclude the %q directory from scan", "vendor")
 	}
-	err = flag.Set("exclude-dir", ".git")
+	err = flag.Set("exclude-dir", "\\.git/")
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "\nError: failed to exclude the %q directory from scan", ".git")
+		fmt.Fprintf(os.Stderr, "\nError: failed to exclude the %q directory from scan", "\\.git/")
 	}
 
 	// set for exclude


timcosta:
it may be worth adding a `(?:^|\/)` to the beginning of the regex as well to enforce that _only_ `.git` directories should be caught by this. The current regex would exclude a folder like `src/something.git/main.go` because it still matches the `\.git`. Curious to hear any thoughts on that as it's an even more restrictive regex but only handles a very specific case whereas the fix as currently implemented handles 9/10 use cases.

yunwei37:
Hi! 

It seems currently gosec is using a helper function to walk through all folders to get all  packages path at a given root directory, see https://github.com/securego/gosec/blob/b0f3e78e07d26ff0156a493fb8232f0de5b8ce15/helpers.go#L395

My question is, why can't we using a simple `packages.Load(cfg, "./...")` to load and return all the Go packages in a given project root? For example:

```go
       config := &packages.Config{
		Mode:        LoadMode,
		BuildFlags: buildTags,
		Tests:          gosec.tests,
                Dir:             root,
	}
        pkgs, err := packages.Load(config, "./...")
	if err != nil {
		return nil, err
	}
```

ccojocar (approved):

PR #834, opened 2022-07-22
Improve SQL check extensibility, add support for pgx

scop:
The SQL checks currently contain a number of hardwired/heuristic assumptions about the index of the argument taking an SQL string. The first commit here generalizes that.

The second one adds support for checking the SQL issues against the popular https://github.com/jackc/pgx family of PostgreSQL drivers and tools.


scop:
Marking as a draft, because the second commit is work in progress; otherwise complete but lacking in test coverage.

Before completing it, I'd like to verify if the pgx support, once complete here, is something you'd welcome in gosec.

ccojocar:
Thanks for this contribution. I would keep the pgx stuff out of gosec since the tool is focused on standard Go package. Thanks for understanding.

scop:
Sure. It's unfortunate though, as there doesn't seem to be anything else one could easily use or extend to craft these checks for non-stdlib packages.

Anyway, I think the refactoring commit https://github.com/securego/gosec/pull/834/commits/a18c3a0edc35b2e3b4ebf1c0dd939e1e27c084e1 still has value on its own. If you agree, I can submit it as another PR.

ccojocar:
@scop Thanks, I left some comments in the commit. I don't see it as a major improvement but rather a bit more complicated and not so obvious. If you can make it a bit more clear, keeping in mind that someone needs maybe at some point extend the method list, you can submit a PR.

scop:
The whole point of that commit is to make it easier for someone at some point to extend the method list. Before my commit, one has to modify two places in the code, and if the function to be added has the query in argument index > 1, or doesn't conform to the "ends with Context => +1" heuristic, the current code wouldn't work for that at all.

After that commit, one just adds the function name and its interesting argument index to sqlCallIdents, and it's done. Note that my second commit actually shows this in action: https://github.com/securego/gosec/pull/834/commits/567d75a58bfb8f10d85acb0ce5ede7ad257157dc -> just addition to the map and it's covered, no matter what the argument indexes are, no other code changes needed.

ccojocar:
Thanks for clarifications! Please send that commit in a different pull request and close this one. Thanks again!

scop:
Done: https://github.com/securego/gosec/pull/841

PR #835, opened 2022-07-22
fix G204 bugs

### LINKED ISSUE #831: Confused about G204
### Summary 

I'm little confused about G204 rules, when I wrote a test file, with following content 

```go
package main

import (
	"log"
	"os"
	"os/exec"
)

func RunCmdWithShortDeclartionsWithInputArgs() {
	cmdstr := os.Args[0]
	cmd := exec.Command(cmdstr)
	err := cmd.Start()
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Waiting for command to finish...")
	err = cmd.Wait()
}

func main() {
        RunCmdWithVariableDeclartionWithInputArgs()
}
```
The gosec tool will complained `Subprocess launched with variable (Confidence: HIGH, Severity: MEDIUM)`. Which I think the output is reasonable.

But when I just change `cmdstr` from short declaration to var key keyword, gosec do not complain anymore.

```go
package main

import (
	"log"
	"os"
	"os/exec"
)

func RunCmdWithVariableDeclartionWithInputArgs() {
	var cmdstr = os.Args[0]
	cmd := exec.Command(cmdstr)
	err := cmd.Start()
	if err != nil {
		log.Fatal(err)
	}
	log.Printf("Waiting for command to finish...")
	err = cmd.Wait()
}

func main() {
        RunCmdWithVariableDeclartionWithInputArgs()
}

```

So I'm confused that since the G204 is intended for making sure there isn't a potential command injection vector here, the way of variable declaration should not matter? 

### gosec version

master

go version go1.18.1 darwin/amd64

macos

### Expected behavior

The way of variable declaration should not influence the gosec's behavior, no matter in short declaration `cmd := os.Args[0]` or in var keyword `var cmd = os.Args[0]`

fatsheep9146:
Signed-off-by: Ziqi Zhao <zhaoziqi9146@gmail.com>

fix #831

### DIFF
--- a/rules/subproc.go
+++ b/rules/subproc.go
@@ -77,6 +77,13 @@ func (r *subprocess) Match(n ast.Node, c *gosec.Context) (*gosec.Issue, error) {
 								return gosec.NewIssue(c, n, r.ID(), "Subprocess launched with variable", gosec.Medium, gosec.High), nil
 							}
 						}
+					case *ast.ValueSpec:
+						_, valueSpec := ident.Obj.Decl.(*ast.ValueSpec)
+						if variable && valueSpec {
+							if !gosec.TryResolve(ident, c) {
+								return gosec.NewIssue(c, n, r.ID(), "Subprocess launched with variable", gosec.Medium, gosec.High), nil
+							}
+						}
 					}
 				}
 			} else if !gosec.TryResolve(arg, c) {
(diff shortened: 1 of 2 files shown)


ccojocar (approved):
lgtm, thanks for fixing this!

PR #859, opened 2022-08-22
fix: add a CWE ID mapping to rule G114

### LINKED ISSUE #858: G114 is not in CWE map
### Summary 
G114 does not show up in the mapping at https://github.com/securego/gosec/blob/master/issue.go#L50

ccojocar:
fixes #858


gb-papercut:
Beaut, thanks. [That CWE](https://cwe.mitre.org/data/definitions/676.html) is very broad. Is the idea of gosec flagging listenAndServe that, without timeouts, it could use up all resources by an attacker tying up connections?

Issue #862, opened 2022-09-04
False positive for G404 if both math/rand and crypto/rand are imported


thaJeztah:
- relates to https://github.com/securego/gosec/issues/820
- relates to https://github.com/securego/gosec/pull/838

### Summary 

If a file imports both `math/rand` and `crypto/rand`, gosec reports a false positive.

### Steps to reproduce the behavior

Create a file that imports both;

```go
package main

import (
	"crypto/rand"
	"math/big"
	rnd "math/rand"
)

func main() {
	_, _ = rand.Int(rand.Reader, big.NewInt(int64(2)))
	_ = rnd.Intn(2)
}
```

### gosec version

Tried both the latest release (v2.3.1) and "master"

```bash
go install github.com/securego/gosec/v2/cmd/gosec@v2.13.1


go install github.com/securego/gosec/v2/cmd/gosec@master
go: downloading github.com/securego/gosec v0.0.0-20200330112059-e030aa4f768b
go: downloading github.com/securego/gosec/v2 v2.13.2-0.20220829082147-ae58325bfea0
go: downloading github.com/nbutton23/zxcvbn-go v0.0.0-20210217022336-fa2cb2858354
go: downloading golang.org/x/tools v0.1.12
go: downloading github.com/google/uuid v1.3.0
go: downloading github.com/gookit/color v1.5.1
go: downloading gopkg.in/yaml.v2 v2.4.0
go: downloading github.com/xo/terminfo v0.0.0-20210125001918-ca9a967f8778
go: downloading golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4
go: downloading golang.org/x/sys v0.0.0-20220722155257-8c9f86f7a55f
```

### Go version (output of 'go version')

```bash
go version
go version go1.19 darwin/amd64
```

### Operating system / Environment

macOS, but reproduces as well on Linux (running inside a `golang:1.19` container

### Actual behavior

```bash
gosec --quiet .
Results:


[/gosec-issue/main.go:10] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    9: func main() {
  > 10: 	_, _ = rand.Int(rand.Reader, big.NewInt(int64(2)))
    11: 	_ = rnd.Intn(2)



Summary:
  Gosec  : dev
  Files  : 1
  Lines  : 12
  Nosec  : 0
  Issues : 1
```

[...]

Issue #820, opened 2022-06-02
gosec does not respect the go package import alias


vishaltak:
### Summary 

If we import `crypto/tls` with an alias( say `cryptotls`), gosec does not recognize it properly and reports issues even though it would not if it `crypto/tls` was imported without an alias.

### Steps to reproduce the behavior
1. Run gosec on the below file using command `gosec .`

```go
package main

import (
	"crypto/tls"
	cryptotls "crypto/tls"
)

func main() {
	// this will not create an issue
	_ = tls.Config{MinVersion: tls.VersionTLS12}
	// this will create an issue
	_ = cryptotls.Config{MinVersion: cryptotls.VersionTLS12}
}
```

Output:
```
~/.asdf/installs/golang/1.17.9/packages/bin/gosec .
[gosec] 2022/06/02 11:37:59 Including rules: default
[gosec] 2022/06/02 11:37:59 Excluding rules: default
[gosec] 2022/06/02 11:37:59 Import directory: /Users/vishaltak/Desktop
[gosec] 2022/06/02 11:38:00 Checking package: main
[gosec] 2022/06/02 11:38:00 Checking file: /Users/vishaltak/Desktop/test.go
Results:


[/Users/vishaltak/Desktop/test.go:12] - G402 (CWE-295): TLS MinVersion too low. (Confidence: HIGH, Severity: HIGH)
    11:         // this will create an issue
  > 12:         _ = cryptotls.Config{MinVersion: cryptotls.VersionTLS12}
    13: }



Summary:
  Gosec  : 2.11.0
  Files  : 1
  Lines  : 13
  Nosec  : 0
  Issues : 1
```

### gosec version
Version: 2.11.0
Git tag: v2.11.0
Build date: 2022-03-21T15:55:53Z

### Go version (output of 'go version')
go version go1.17.9 darwin/arm64

### Operating system / Environment
Output of `uname -a` is

```
Darwin Vishals-MacBook-Pro.local 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:46:32 PDT 2022; root:xnu-8020.101.4~15/RELEASE_ARM64_T6000 arm64
```

### Expected behavior
No issues should be reported by gosec when import alias is used

### Actual behavior
Issue is reported by gosec when import alias is used

TimonOmsk:
@ccojocar Can I fix this? :)

ccojocar:
Sure, go for it! Thanks

PR #863, opened 2022-09-04
Fix false positives for G404 with aliased packages

thaJeztah:
- fixes https://github.com/securego/gosec/issues/862
- relates to https://github.com/securego/gosec/issues/820
- relates to https://github.com/securego/gosec/pull/838

It appears that `GetImportedName` returns _both_ aliased and non-aliased imports.
As a result, a file having both crypto/rand and math/rand (but aliased) would
trigger false positives on G404. Given the following file;

```go
package main

import (
	"crypto/rand"
	"math/big"
	rnd "math/rand"
)

func main() {
	_, _ = rand.Int(rand.Reader, big.NewInt(int64(2)))
	_ = rnd.Intn(2)
}
```

And patching for debugging;

```patch
diff --git a/helpers.go b/helpers.go
index 437d032..80f4233 100644
--- a/helpers.go
+++ b/helpers.go
@@ -250,6 +250,8 @@ func GetBinaryExprOperands(be *ast.BinaryExpr) []ast.Node {
 // GetImportedName returns the name used for the package within the
 // code. It will ignore initialization only imports.
 func GetImportedName(path string, ctx *Context) (string, bool) {
+       fmt.Printf("%+v", ctx.Imports.Imported)
+       os.Exit(1)
        importName, imported := ctx.Imports.Imported[path]
        if !imported {
                return "", false
```

Would show that `math/rand` was included in the list, using it's non-aliased
name (`:rand`).

    gosec -quiet .
    map[crypto/rand:rand math/big:big math/rand:rand]

This patch works around this problem by reversing the order in which imports
are resolved in `MatchCallByPackage()`. Aliased packages are tried first, after
which non-aliased imports are tried.

Given the example application mentioned above:

Before this patch:

```bash
gosec -quiet .
Results:

[/Users/sebastiaan/Projects/test/gosec-issue/main.go:10] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
    9: func main() {
  > 10: 	_, _ = rand.Int(rand.Reader, big.NewInt(int64(2)))
    11: 	_ = rnd.Intn(2)
```

With this patch applied:

```bash
    gosec --quiet .
    Results:

    [/Users/sebastiaan/Projects/test/gosec-issue/main.go:11] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
        10: 	_, _ = rand.Int(rand.Reader, big.NewInt(int64(2)))
      > 11: 	_ = rnd.Intn(2)
        12: }
```

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -37,9 +37,9 @@ import (
 //
 //	node, matched := MatchCallByPackage(n, ctx, "math/rand", "Read")
 func MatchCallByPackage(n ast.Node, c *Context, pkg string, names ...string) (*ast.CallExpr, bool) {
-	importedName, found := GetImportedName(pkg, c)
+	importedName, found := GetAliasedName(pkg, c)
 	if !found {
-		importedName, found = GetAliasedName(pkg, c)
+		importedName, found = GetImportedName(pkg, c)
 		if !found {
 			return nil, false
 		}
--- a/testutils/source.go
+++ b/testutils/source.go
@@ -3180,6 +3180,22 @@ func main() {
 	bad := rand.Intn(10)
 	println(bad)
 }`}, 1, gosec.NewConfig()},
+		{[]string{`
+package main
+
+import (
+	"crypto/rand"
+	"math/big"
+	rnd "math/rand"
+)
+
+func main() {
+	good, _ := rand.Int(rand.Reader, big.NewInt(int64(2)))
+	println(good)
+	bad := rnd.Intn(2)
+	println(bad)
+}
+`}, 1, gosec.NewConfig()},
 	}
 
 	// SampleCodeG501 - Blocklisted import MD5


thaJeztah:
@TimonOmsk @ccojocar ptal

TimonOmsk:
I have a strong feeling that there is a case that can break that logic too. But I've failed to find it :) So, this PR looks good for me 👍

ccojocar (approved):
lgmt, thanks for fixing this issue!

thaJeztah:
> I have a strong feeling that there is a case that can break that logic too.

Yes, I'll be the first to admit that it felt like a "workaround", but also tried to come up with issues in this approach, and couldn't think of any, so kept it simple.

I think there _will_ be issues if a file would have duplicate imports, such as;

```go
import(
   "math/rand"
    foo "math/rand"
    bar "math/rand"
    baz "math/rand"
)
```

Perhaps we need a `GetImportedNames` (plural) instead that returns a slice of import names (non-aliased and aliased combined). That said; having duplicate imports isn't great, and should already make other linters fail.


> thanks for fixing this issue!

Thanks for reviewing (both!). I hadn't coded on linters before, and expected it to be complicated, but after looking a bit at the code involved in this part, it turned out to be not too difficult, so thought I might as well open a PR and fix it 😄

PR #865, opened 2022-09-05
Refactor to support duplicate imports with different aliases

thaJeztah:
The existing code assumed imports to be either imported, or imported with an
alias. Badly formatted files may have duplicate imports for a package, using
different aliases.

This patch refactors the code, and;

Introduces a new `GetImportedNames` function, which returns all name(s) and
aliase(s) for a package, which effectively combines `GetAliasedName` and
`GetImportedName`, but adding support for duplicate imports.

The old `GetAliasedName` and `GetImportedName` functions have been rewritten to
use the new function and marked deprecated, but could be removed if there are no
external consumers.

With this patch, the linter is able to detect issues in files such as;

    package main

    import (
        crand "crypto/rand"
        "math/big"
        "math/rand"
        rand2 "math/rand"
        rand3 "math/rand"
    )

    func main() {
        _, _ = crand.Int(crand.Reader, big.NewInt(int64(2))) // good

        _ = rand.Intn(2) // bad
        _ = rand2.Intn(2)  // bad
        _ = rand3.Intn(2)  // bad
    }

Before this patch, only a single issue would be detected:

    gosec --quiet .

    [main.go:14] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
        13:
      > 14: 	_ = rand.Intn(2) // bad
        15: 	_ = rand2.Intn(2)  // bad

With this patch, all issues are identified:

    gosec --quiet .

    [main.go:16] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
        15: 	_ = rand2.Intn(2)  // bad
      > 16: 	_ = rand3.Intn(2)  // bad
        17: }

    [main.go:15] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
        14: 	_ = rand.Intn(2) // bad
      > 15: 	_ = rand2.Intn(2)  // bad
        16: 	_ = rand3.Intn(2)  // bad

    [main.go:14] - G404 (CWE-338): Use of weak random number generator (math/rand instead of crypto/rand) (Confidence: MEDIUM, Severity: HIGH)
        13:
      > 14: 	_ = rand.Intn(2) // bad
        15: 	_ = rand2.Intn(2)  // bad

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -172,9 +172,9 @@ func (gosec *Analyzer) Process(buildTags []string, packagePaths ...string) error
 		for {
 			select {
 			case s := <-j:
-				packages, err := gosec.load(s, config)
+				pkgs, err := gosec.load(s, config)
 				select {
-				case r <- result{pkgPath: s, pkgs: packages, err: err}:
+				case r <- result{pkgPath: s, pkgs: pkgs, err: err}:
 				case <-quit:
 					// we've been told to stop, probably an error while
 					// processing a previous result.
@@ -296,7 +296,6 @@ func (gosec *Analyzer) Check(pkg *packages.Package) {
 		gosec.context.Pkg = pkg.Types
 		gosec.context.PkgFiles = pkg.Syntax
 		gosec.context.Imports = NewImportTracker()
-		gosec.context.Imports.TrackFile(file)
 		gosec.context.PassedValues = make(map[string]interface{})
 		ast.Walk(gosec, file)
 		gosec.stats.NumFiles++
@@ -434,6 +433,12 @@ func (gosec *Analyzer) Visit(n ast.Node) ast.Visitor {
 		}
 		return gosec
 	}
+	switch i := n.(type) {
+	case *ast.File:
+		// Using ast.File instead of ast.ImportSpec, so that we can track
+		// all imports at once.
+		gosec.context.Imports.TrackFile(i)
+	}
 
 	// Get any new rule exclusions.
 	ignoredRules := gosec.ignore(n)
@@ -452,9 +457,6 @@ func (gosec *Analyzer) Visit(n ast.Node) ast.Visitor {
 
 	// Push the new set onto the stack.
 	gosec.context.Ignores = append([]map[string][]SuppressionInfo{ignores}, gosec.context.Ignores...)
-
-	// Track aliased and initialization imports
-	gosec.context.Imports.TrackImport(n)
 
 	for _, rule := range gosec.ruleset.RegisteredFor(n) {
 		// Check if all rules are ignored.
--- a/helpers.go
+++ b/helpers.go
@@ -37,12 +37,9 @@ import (
 //
 //	node, matched := MatchCallByPackage(n, ctx, "math/rand", "Read")
 func MatchCallByPackage(n ast.Node, c *Context, pkg string, names ...string) (*ast.CallExpr, bool) {
-	importedName, found := GetAliasedName(pkg, c)
+	importedNames, found := GetImportedNames(pkg, c)
 	if !found {
-		importedName, found = GetImportedName(pkg, c)
-		if !found {
-			return nil, false
-		}
+		return nil, false
 	}
 
 	if callExpr, ok := n.(*ast.CallExpr); ok {
@@ -50,7 +47,10 @@ func MatchCallByPackage(n ast.Node, c *Context, pkg string, names ...string) (*a
 		if err != nil {
 			return nil, false
 		}
-		if packageName == importedName {
+		for _, in := range importedNames {
+			if packageName != in {
+				continue
+			}
 			for _, name := range names {
 				if callName == name {
 					return callExpr, true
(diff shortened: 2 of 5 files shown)


thaJeztah:
Alright, this needs some work; have a false positive on 

```go
package main

import (
	"crypto/rand"
	mrand "math/rand"
)

func main() {
	good, _ := rand.Read(nil)
	println(good)
	bad := mrand.Int31()
	println(bad)
}
```

helpers.go
```diff
@@ -247,48 +247,67 @@ func GetBinaryExprOperands(be *ast.BinaryExpr) []ast.Node {
...
+
+	return importNames, true
+}
+
+// GetImportedName returns the name used for the package within the code.
+// If multiple names are found, it returns the first hit. It ignores
+// initialization-only imports.
+//
+// Deprecated: use GetImportedNames
+func GetImportedName(path string, ctx *Context) (string, bool) {
```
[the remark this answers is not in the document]
ccojocar: I think this can be removed, it is not supposed to be used outside of the project.

import_tracker.go
```diff
@@ -22,54 +22,60 @@ import (
 // by a source file. It is able to differentiate between plain imports, aliased
 // imports and init only imports.
 type ImportTracker struct {
-	Imported map[string]string
-	Aliased  map[string]string
+	// Imported is a map of imports with their associated names/aliases.
+	Imported map[string][]string
 	InitOnly map[string]bool
```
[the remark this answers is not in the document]
ccojocar: I think is fine, this is not supposed to be used outside of the project. We didn't do a good job to isolate this as `internal`.

ccojocar (commented):
Thanks for working on this. It looks good so far.

mauriciocm9:
any update?

ccojocar:
Hi @thaJeztah, Do you still plan to finish this pull request? Thanks

thaJeztah:
@ccojocar I think I fixed it, but looks like CI is failing due to older Go versions being installed, and an outdated module; I opened https://github.com/securego/gosec/pull/880 to fix / work around that, and temporarily rebased this PR on top of that (will rebase once the other PR is merged)

import_tracker.go
```diff
@@ -22,54 +22,63 @@ import (
...
 
 // TrackFile track all the imports used by the supplied file
 func (t *ImportTracker) TrackFile(file *ast.File) {
 	for _, imp := range file.Imports {
-		path := strings.Trim(imp.Path.Value, `"`)
-		parts := strings.Split(path, "/")
-		if len(parts) > 0 {
-			name := parts[len(parts)-1]
-			t.Imported[path] = name
+		importPath := strings.Trim(imp.Path.Value, `"`)
```
thaJeztah, line 39: This was part of the problem; the old code needed to manually track the file, to get the "un-aliased" import, and then used the ImportTracker.TrackImport to get the aliased versions. But the `TrackFile` would ignore aliases.

import_tracker_test.go
```diff
@@ -47,7 +47,7 @@ var _ = Describe("Import Tracker", func() {
 			files := pkgs[0].Syntax
 			Expect(files).Should(HaveLen(1))
 			tracker.TrackFile(files[0])
-			Expect(tracker.Imported).Should(Equal(map[string]string{"fmt": "fmt"}))
+			Expect(tracker.Imported).Should(Equal(map[string][]string{"fmt": {"fm"}}))
```
thaJeztah, line 50: Which can be seen here (the file tested has `fm` as alias for `fmt`, but `TrackFile()` doesn't detect that.

analyzer.go
```diff
@@ -296,7 +296,6 @@ func (gosec *Analyzer) Check(pkg *packages.Package) {
 		gosec.context.Pkg = pkg.Types
 		gosec.context.PkgFiles = pkg.Syntax
 		gosec.context.Imports = NewImportTracker()
-		gosec.context.Imports.TrackFile(file)
 		gosec.context.PassedValues = make(map[string]interface{})
 		ast.Walk(gosec, file)
```
thaJeztah, line 300: ...and because `Analyzer.Check()` called _both_ `TrackFile()` and `ast.Walk()`, the incorrect import name would be added (so a non-aliased import would be added, even though the package used an alias), which broke detection.

With the new code, `CheckImports()` finds both, so calling `TrackFile` is no longer needed.

I wasn't sure if that function could be removed altogether (it's slightly easier to use than starting an Analyzer, so perhaps it's useful as a utility still).

helpers.go
```diff
@@ -247,48 +247,23 @@ func GetBinaryExprOperands(be *ast.BinaryExpr) []ast.Node {
...
-// code. It will ignore initialization only imports.
-func GetAliasedName(path string, ctx *Context) (string, bool) {
-	importName, imported := ctx.Imports.Aliased[path]
-	if !imported {
-		return "", false
-	}
-
-	if _, initonly := ctx.Imports.InitOnly[path]; initonly {
-		return "", false
-	}
```
thaJeztah, line 275: The `initOnly` seemed redundant; all tests seem to pass without it, so I removed it, but maybe there was a reason it was there 😅

ccojocar (approved):
lgtm, please could you please rebase it now? Is this still a draft?

analyzer.go
thaJeztah: > I wasn't sure if that function could be removed altogether (it's slightly easier to use than starting an Analyzer, so perhaps it's useful as a utility still).

Actually, perhaps it should be the reverse; use `TrackFile` once, instead of during `Visit()` (we only need it once per file). Could perhaps be as part as `Visit()` though.

thaJeztah:
Working on some last changes; perhaps it's better after all to handle imports at the File level, otherwise we're calling it for every Node; let me push those changes 👍

thaJeztah:
@ccojocar should be good now; PTAL 🤗

ccojocar (approved):
Perfect! Thanks!

thaJeztah:
YW! Was fun digging a bit into how the ast Parsing works 😄

Issue #732, opened 2021-11-23
Issues scanning private GitHub repos added as dependencies for another project


Ravikanth143143:
### Summary 
I am using gosec v2.9.2 in github workflow. 

I have a project X that has all the commons code and project Y that includes the dependencies of X. Without the dependencies `gosec ./...` in Y seems to run fine with no issues. However, when the dependencies are added it seems to break the security scanning. Below is the error:

> [gosec] 2021/11/23 04:58:25 Import directory: /github/workspace/logger
> [gosec] 2021/11/23 04:58:26 Checking package: logger
> [gosec] 2021/11/23 04:58:26 Checking file: /github/workspace/logger/logger.go
> 
> Results:
> Golang errors in file: [/github/workspace/logger/logger.go]:

  > [line 5 : column 2] - could not import github.com/MyRepo/common-golang-lib/config (invalid package name: "")

I also set:
`GO111MODULE: on`

### Steps to reproduce the behavior
Create a private github repo and reference it in your current project on which you need to run gosec

### gosec version
v2.9.2

### Go version (output of 'go version')
1.17

### Operating system / Environment
ubuntu using github workflow

### Expected behavior
There should be a way to scan the private github repos

### Actual behavior
private github repos doesn't get scanned.

ccojocar:
@Ravikanth143143 Can you build successfully the project with `go build` when you include the dependencies? It seems that the go module dependencies cannot be solved. You can also try to run `go mod download` before running the gosec scan.

ccojocar:
I am closing this issue since it seems to me that is related to the go dependencies and not gosec tool. Note that gosec relies on go build infrastructure to resolve the go module dependencies.

Issue #490, opened 2020-06-18
GitHub action error: could not read Username


IratxeGarrido:
### Summary 
My goal is to create a GitHub actions workflow that runs gosec against one of our repositories. The repository can only be accessed using an ssh key and uses dependencies from other private repositories. I was able to run the tool locally (MacOS) without problems but not as a GitHub action. I have created a secret in the repository with a private key that allows me to access the repository. The webfactory/ssh-agent should allow me to extend this functionality to other repositories but gosec is still not working.
### Steps to reproduce the behavior
This is the go.yml file that is running.
```
name: Run Gosec
on:
  push:
    branches:
      - SPR-8551-gosec-in-github-actions
jobs:
  tests:
    runs-on: ubuntu-latest
    env:
      GO111MODULE: on
    steps:
      - name: Checkout Source 
        uses: actions/checkout@v2
        #with:
        #  repo-token: ${{ SSH_PRIVATE_KEY }}
      - name: webfactory/ssh-agent
        uses: webfactory/ssh-agent@v0.3.0
        with:
          ssh-private-key: ${{ secrets.SSH_PRIVATE_KEY }}
      - name: Run Gosec Security Scanner
        uses: securego/gosec@master
        with:
          args: ./...
```
### gosec version
2.263
### Go version (output of 'go version')
go version go1.14.3 darwin/amd64
### Operating system / Environment
The action is running Ubuntu 18.04.4
```
2020-06-17T12:56:42.6972391Z Current runner version: '2.263.0'
2020-06-17T12:56:42.7002024Z ##[group]Operating System
2020-06-17T12:56:42.7003399Z Ubuntu
2020-06-17T12:56:42.7004365Z 18.04.4
2020-06-17T12:56:42.7005357Z LTS
2020-06-17T12:56:42.7006359Z ##[endgroup]
2020-06-17T12:56:42.7007437Z ##[group]Virtual Environment
2020-06-17T12:56:42.7008233Z Environment: ubuntu-18.04
2020-06-17T12:56:42.7009132Z Version: 20200604.1
2020-06-17T12:56:42.7010044Z Included Software: https://github.com/actions/virtual-environments/blob/ubuntu18/20200604.1/images/linux/Ubuntu1804-README.md
```
### Expected behavior
I want to generate a json file with the tool's output like this one:
```
{
	"Golang errors": {},
	"Issues": [
		{
			"severity": "HIGH",
			"confidence": "HIGH",
			"cwe": {
				"ID": "200",
[...]

PR #869, opened 2022-09-23
Add a way to use private repositories on GitHub

elgohr:
Would be great to have something like this inside the Docker image for working with private repositories.
Especially when running local and there's no other way than vendoring at the moment.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -4,4 +4,8 @@
 # provides all arguments concatenated as a single string.
 ARGS=("$@")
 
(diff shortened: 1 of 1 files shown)


ccojocar (approved):

pr0PM:
@ccojocar I think this would help in resolving the credential issues with gosec GitHub action while run on private repositories too.

tldr;
was just trying to figure out how to make it run on a private repo and noticed these
https://github.com/securego/gosec/issues/490
https://github.com/securego/gosec/issues/732

After googling for a while I noticed this 20h old merge!

elgohr:
@pr0PM That was also one thought behind it 😁

PR #871, opened 2022-09-30
flag ALL use of Go xcrypto as bad in rule G506

thorrsson:


thorrsson:
Sorry, was testing something in my own fork and opened this against the wrong upstream

PR #880, opened 2022-10-16
go.mod:  ginkgo/v2 v2.3.1, golang.org/x/text v0.3.8, update go versions

thaJeztah:
### gha: remove go1.17, temporarily force go1.18.7

The  security scanner is flagging the code to have a vulnerability, but it's
detecting that we're running go1.18.6, not "latest" (go1.18.7 at time of writing).

Temporarily pinning to go1.18.7 to force installing the latest version:

    Vulnerability #1: GO-2022-1039
      Programs which compile regular expressions from untrusted
      sources may be vulnerable to memory exhaustion or denial of
      service. The parsed regexp representation is linear in the size
      of the input, but in some cases the constant factor can be as
      high as 40,000, making relatively small regexps consume much
      larger amounts of memory. After fix, each regexp being parsed is
      limited to a 256 MB memory footprint. Regular expressions whose
      representation would use more space than that are rejected.
      Normal use of regular expressions is unaffected.
    
      Call stacks in your code:
      Error:       helpers.go:463:26: github.com/securego/gosec/v2.ExcludedDirsRegExp calls regexp.MustCompile, which eventually calls regexp/syntax.Parse
    
      Found in: regexp/syntax@go1.18.6
      Fixed in: regexp/syntax@go1.19.2
      More info: https://pkg.go.dev/vuln/GO-2022-1039

### go.mod: github.com/onsi/ginkgo/v2 v2.3.1

CI was failing because of a mismatch:

    /home/runner/go/bin/ginkgo -v --fail-fast
    Ginkgo detected a version mismatch between the Ginkgo CLI and the version of Ginkgo imported by your packages:
      Ginkgo CLI Version:
        2.3.1
      Mismatched package versions found:
       2.2.0 used by gosec

### go.mod: golang.org/x/text v0.3.8

to address GO-2022-1059:

    The vulnerabilities below are in packages that you import, but your code
    doesn't appear to call any vulnerable functions. You may not need to take any
    action. See https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck
    for details.
    
    Vulnerability #1: GO-2022-1059
      An attacker may cause a denial of service by crafting an Accept-Language
      header which ParseAcceptLanguage will take significant time to parse.
    
      Found in: golang.org/x/text/language@v0.3.7
      Fixed in: golang.org/x/text/language@v0.3.8
      More info: https://pkg.go.dev/vuln/GO-2022-1059


.github/workflows/ci.yml
```diff
@@ -11,8 +11,7 @@ jobs:
     strategy:
       matrix:
         go_version:
-          - '1.17'
-          - '1.18'
+          - '1.18.7' # TODO: remove this once actions/setup-go@v3 uses latest as latest; see https://github.com/securego/gosec/pull/880
```
thaJeztah, line 14: This shouldn't be needed, but looks like the action installs 1.18.6 otherwise

thaJeztah:
Wow, and now it's installing go1.19.1 ??? Looks like if you don't specify a version, it picks  a _random_ 1.19.x version (perhaps caching??)

```
  Call stacks in your code:
Error:       helpers.go:463:26: github.com/securego/gosec/v2.ExcludedDirsRegExp calls regexp.MustCompile, which eventually calls regexp/syntax.Parse

  Found in: regexp/syntax@go1.19.1
  Fixed in: regexp/syntax@go1.19.2
```

thaJeztah:
@ccojocar ptal 👍

ccojocar (approved):
Thanks for fixing this! I wanted to do the same until the Go version gets updated into the action.

Issue #891, opened 2022-11-13
G104 fails to catch err reassignments


omercnet:
### Summary 
Since `err` is commonly used, it's not uncommon to have it reassigned
If someone forgets to check it before reassigning it, G104 should catch it

### Steps to reproduce the behavior
```go
package main

import (
	"io"
	"log"
	"os"
)

func main() {
	_, _ = io.WriteString(os.Stdout, "Hello World") // # this is ok

	_, err := io.WriteString(os.Stdout, "Hello World")
	if err != nil { // good
		log.Fatal(err)
	}

	_, err = io.WriteString(os.Stdout, "Hello World") // # this err will not be checked
	_, err = io.WriteString(os.Stdout, "Hello World") // # this err will be checked

	if err != nil { // checking the second err but not the first one
		log.Fatal(err)
	}

}
```

### gosec version
2.14.0
### Go version (output of 'go version')
go version go1.19.2
### Operating system / Environment
darwin/arm64
### Expected behavior
Should catch the `err` that wasn't checked
### Actual behavior
Doesn't.

omercnet:
https://staticcheck.io/ does catch this, worth looking into how they implement this check

TimonOmsk:
@ccojocar I'll take this one. Finally have some time to dive deep into the SSA :)

ccojocar:
@TimonOmsk feel free to take on this issue. Thanks

PR #895, opened 2022-11-25
G104 now raises an issue in case of reassignment err variable without checking

TimonOmsk:
This PR is related to https://github.com/securego/gosec/issues/891 issue

Now the G104 also checks for reassignments by checking referrers of err variable. 

An issue will be raised only in case if no other expression related to the err variable exists:
Screenshot 2022-11-25 at 2 31 09

This check was made using SSA(https://pkg.go.dev/golang.org/x/tools/go/ssa)

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -31,10 +31,9 @@ import (
 	"strings"
 	"sync"
 
+	"golang.org/x/tools/go/packages"
 	"golang.org/x/tools/go/ssa"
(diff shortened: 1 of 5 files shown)


TimonOmsk:
Closed because SSA requires complete program to build SSA, but gosec builds package by package and internal imports cannot be resolved. 

Screenshot 2022-11-25 at 23 21 08

PR #929, opened 2023-02-09
Add support for Go analysis framework and SSA code representation

### LINKED ISSUE #280: Rewrite to leverage the go analysis package
We should consider adopting a more standard framework for the gosec internals. 

https://go.googlesource.com/tools/+/master/go/analysis/doc.go

ccojocar:
Fixes #280


sluongng:
I think this paved the foundation to solving https://github.com/securego/gosec/issues/280 but not yet fully fixing it?

The only `analysis.Analyzer` in the MR comes with a big TODO.
Would you consider re-opening https://github.com/securego/gosec/issues/280 until that is fully implemented?

ccojocar:
That's right, I reopened it. 

My goal is anyhow to complement the existing rules with better analysis based on SSA. I think over time, we can move some rules over which can be better implemented in SSA. For instance SSA enables taint analysis but I believe AST still provides other benefits. Both representation can coexist.

PR #971, opened 2023-06-13
Feature: G101 match variable values and names

morgenm:
I updated the HardcodedCredentials code to apply regex against the actual values themselves in assignment statements, valuespecs, and equalities.  The intention is to search for secrets, such as API keys, by looking for the values. For example if we have:

`key = "3490b39bc66f7e7730e8fa68041da2e4f7c0a6a8cb119466c3ddc5245bda1781"`

this would previously not be caught by gosec, because gosec only looks at the variable names. 

Now, with these changes, the code matches values too using a new regex pattern that will look for secrets such as this. I added a new config variable called "patternValue," with a default regex which searches for VirusTotal API keys (or other keys with the same format).

### DIFF
--- a/rules/hardcoded_credentials.go
+++ b/rules/hardcoded_credentials.go
@@ -29,6 +29,7 @@ import (
 type credentials struct {
 	issue.MetaData
 	pattern          *regexp.Regexp
+	patternValue	 *regexp.Regexp // Pattern for matching string values (LHS on assign statements)
 	entropyThreshold float64
 	perCharThreshold float64
 	truncate         int
@@ -70,9 +71,21 @@ func (r *credentials) Match(n ast.Node, ctx *gosec.Context) (*issue.Issue, error
 func (r *credentials) matchAssign(assign *ast.AssignStmt, ctx *gosec.Context) (*issue.Issue, error) {
 	for _, i := range assign.Lhs {
 		if ident, ok := i.(*ast.Ident); ok {
+			// Match the variable name
 			if r.pattern.MatchString(ident.Name) {
 				for _, e := range assign.Rhs {
 					if val, err := gosec.GetString(e); err == nil {
+						if r.ignoreEntropy || (!r.ignoreEntropy && r.isHighEntropyString(val)) {
+							return ctx.NewIssue(assign, r.ID(), r.What, r.Severity, r.Confidence), nil
+						}
+					}
+				}
+			}
+
+			// Match the value assigned
+			for _, e := range assign.Rhs {
+				if val, err := gosec.GetString(e); err == nil {
+					if r.patternValue.MatchString(val) {
 						if r.ignoreEntropy || (!r.ignoreEntropy && r.isHighEntropyString(val)) {
 							return ctx.NewIssue(assign, r.ID(), r.What, r.Severity, r.Confidence), nil
 						}
@@ -127,6 +140,7 @@ func (r *credentials) matchEqualityCheck(binaryExpr *ast.BinaryExpr, ctx *gosec.
 // assigned to variables that appear to be related to credentials.
 func NewHardcodedCredentials(id string, conf gosec.Config) (gosec.Rule, []ast.Node) {
 	pattern := `(?i)passwd|pass|password|pwd|secret|token|pw|apiKey|bearer|cred`
+	patternValue := "(?i)[a-f0-9]{64}"
 	entropyThreshold := 80.0
 	perCharThreshold := 3.0
 	ignoreEntropy := false
@@ -138,6 +152,13 @@ func NewHardcodedCredentials(id string, conf gosec.Config) (gosec.Rule, []ast.No
 				pattern = cfgPattern
 			}
 		}
+
+		if configPatternValue, ok := conf["patternValue"]; ok {
+			if cfgPatternValue, ok := configPatternValue.(string); ok {
+				patternValue = cfgPatternValue
+			}
+		}
+
 		if configIgnoreEntropy, ok := conf["ignore_entropy"]; ok {
 			if cfgIgnoreEntropy, ok := configIgnoreEntropy.(bool); ok {
 				ignoreEntropy = cfgIgnoreEntropy
@@ -168,6 +189,7 @@ func NewHardcodedCredentials(id string, conf gosec.Config) (gosec.Rule, []ast.No
 
 	return &credentials{
 		pattern:          regexp.MustCompile(pattern),
+		patternValue:	  regexp.MustCompile(patternValue),
 		entropyThreshold: entropyThreshold,
 		perCharThreshold: perCharThreshold,
 		ignoreEntropy:    ignoreEntropy,
--- a/rules/hardcoded_credentials.go
+++ b/rules/hardcoded_credentials.go
@@ -116,6 +116,12 @@ func (r *credentials) matchValueSpec(valueSpec *ast.ValueSpec, ctx *gosec.Contex
 
 func (r *credentials) matchEqualityCheck(binaryExpr *ast.BinaryExpr, ctx *gosec.Context) (*issue.Issue, error) {
 	if binaryExpr.Op == token.EQL || binaryExpr.Op == token.NEQ {
+		identStrConst, okLit := binaryExpr.X.(*ast.BasicLit)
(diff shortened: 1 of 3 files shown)


rules/hardcoded_credentials.go
```diff
@@ -79,12 +81,24 @@ func (r *credentials) matchAssign(assign *ast.AssignStmt, ctx *gosec.Context) (*
 					}
 				}
 			}
+
+			// Match the value assigned
+			for _, e := range assign.Rhs {
+				if val, err := gosec.GetString(e); err == nil {
```
audunmo, line 87: These deep nestings can be quite hard to follow. Can this be flattened? Maybe just `continue` on `err != nil`, then put the rest of the logic under? Currently, this `err` seems to be unhandled. If we truly don't care about it, you can `_` it, though you'll likely want to log it or return it

rules/hardcoded_credentials.go
```diff
@@ -79,12 +81,24 @@ func (r *credentials) matchAssign(assign *ast.AssignStmt, ctx *gosec.Context) (*
...
+					if r.patternValue.MatchString(val) {
+						if r.ignoreEntropy || (!r.ignoreEntropy && r.isHighEntropyString(val)) {
```
audunmo, line 89: `!r.ignoreEntropy` will always be `true` here, since it'll only get evaluated if `r.ignoreEntropy` is false. The expression can be simplified to `r.ignoreEntropy || r.isHighEntropyString(val)`

rules/hardcoded_credentials.go
```diff
@@ -98,11 +112,29 @@ func (r *credentials) matchValueSpec(valueSpec *ast.ValueSpec, ctx *gosec.Contex
 			}
 		}
 	}
+
+	// Match values
+	for _, ident := range valueSpec.Values {
+		if val, err := gosec.GetString(ident); err == nil {
+			if r.patternValue.MatchString(val) {
+				if r.ignoreEntropy || (!r.ignoreEntropy && r.isHighEntropyString(val)) {
```
audunmo, line 120: `!r.ignoreEntropy` is always true

rules/hardcoded_credentials.go
```diff
@@ -119,6 +151,14 @@ func (r *credentials) matchEqualityCheck(binaryExpr *ast.BinaryExpr, ctx *gosec.
 				}
 			}
 		}
+		if okLit { // Match for literals
+			s, err := gosec.GetString(identStrConst)
```
audunmo, line 155: `err` is unhandled. If it shouldn't propagate the error up, it should probably be logged

rules/hardcoded_credentials.go
```diff
@@ -119,6 +151,14 @@ func (r *credentials) matchEqualityCheck(binaryExpr *ast.BinaryExpr, ctx *gosec.
...
+			if err == nil &&  r.patternValue.MatchString(s) {
+				if r.ignoreEntropy || (!r.ignoreEntropy && r.isHighEntropyString(s)) {
```
audunmo, line 157: `!r.ignoreEntropy` is always true

rules/hardcoded_credentials.go
```diff
@@ -98,11 +112,29 @@ func (r *credentials) matchValueSpec(valueSpec *ast.ValueSpec, ctx *gosec.Contex
...
+			}
+		}
+	}
+
 	return nil, nil
 }
 
 func (r *credentials) matchEqualityCheck(binaryExpr *ast.BinaryExpr, ctx *gosec.Context) (*issue.Issue, error) {
 	if binaryExpr.Op == token.EQL || binaryExpr.Op == token.NEQ {
+		identStrConst, okLit := binaryExpr.X.(*ast.BasicLit)
```
audunmo, line 132: Looks like the rest of the code base uses `ok`. Unless there's a good reason to specify this in particular, it'd be good to match the codebase

rules/hardcoded_credentials.go
```diff
@@ -79,12 +81,24 @@ func (r *credentials) matchAssign(assign *ast.AssignStmt, ctx *gosec.Context) (*
...
+					}
+				}
+			}
 		}
 	}
 	return nil, nil
 }
 
 func (r *credentials) matchValueSpec(valueSpec *ast.ValueSpec, ctx *gosec.Context) (*issue.Issue, error) {
+	// Match name
```
audunmo, line 101: Comments should describe why something is the way it is. The what comes from reading the code

rules/hardcoded_credentials.go
```diff
@@ -127,6 +167,7 @@ func (r *credentials) matchEqualityCheck(binaryExpr *ast.BinaryExpr, ctx *gosec.
 // assigned to variables that appear to be related to credentials.
 func NewHardcodedCredentials(id string, conf gosec.Config) (gosec.Rule, []ast.Node) {
 	pattern := `(?i)passwd|pass|password|pwd|secret|token|pw|apiKey|bearer|cred`
+	patternValue := "^(?i)[a-f0-9]{64}$"
```
audunmo, line 170: Many API keys also use `-_` in their syntax. Also, the length of 64 seems arbitrary to me. [Talisman](https://github.com/thoughtworks/talisman) is a secret scanner written in go, they may have a good solution that can be mimicked here

ccojocar (approved):
Thanks for this contribution. It looks good. Is this already covered by existing rule tests? If not, please could you add some test cases? Thanks again

ccojocar:
There are some lint issues. Please could you fix them?

audunmo (commented):
I have some quick thoughts on the code here. As an aside, you could also install [Talisman](https://github.com/thoughtworks/talisman) to solve this issue for yourself today.

Also, would probably be good to have a test case in this PR

audunmo:
Don't mean to butt in here, just really excited about `gosec` 😄

rules/hardcoded_credentials.go
morgenm: Yes you are right, it was arbitrary. I just put one in to specifically get VirusTotal API keys. I am taking a look at the repo you linked. I will do some digging and testing to figure out a good solution.

morgenm: Yes you are right, it was arbitrary. I just put one in to specifically get VirusTotal API keys. I am taking a look at the repo you linked. I will do some digging and testing to figure out a good solution.

rules/hardcoded_credentials.go
morgenm: Nice catch! I resolved these short-circuit evaluation suggestions.

rules/hardcoded_credentials.go
morgenm: I did this because there were two ok variables, but I noticed I could rearrange the code to resolve this.

rules/hardcoded_credentials.go
morgenm: I wrote better comments to resolve this.

rules/hardcoded_credentials.go
morgenm: I am gonna go ahead and resolve this comment, but if there are any questions are anything wrong with my solution, you can reopen it on the latest commit.

rules/hardcoded_credentials.go
morgenm: Talisman is using various [regexes for their detector](https://github.com/thoughtworks/talisman/blob/e8b8eba8718b137d723007877f3d105498193a81/detector/pattern/pattern_detector.go#L5]); I am not sure if there is a more general solution that would catch a wide variety of API keys.  So I found some regex patterns for a couple of other keys [here](https://github.com/odomojuli/RegExAPI). I added regex for Google API keys, GitHub Fine-Grained PATs, and Google OAuth keys, as well as tests for each. The tests are run against some fake API keys I randomly generated based on the pattern.

morgenm:
> Don't mean to butt in here, just really excited about `gosec` smile

Thanks for the suggestions on my code changes. Please me know if there are any further suggestions or anything.

> There are some lint issues. Please could you fix them?

I think I fixed them, I do not see any more issues on when running golangci-lint locally. I have no further changes unless anything else comes up.

ccojocar:
@morgenm Thanks for addressing all review comments. The change looks good now.

Issue #954, opened 2023-04-29
No issue reported for slice bounds out of range which causes panic


domsolutions:
### Summary 

No issue reported for resclicing a slice whose length is smaller than the reslice upper limit. Could lead to DoS attacks. Staticcheck also doesn't seem to alert on this.

### Steps to reproduce the behavior

Run against 

```golang
package main

import "fmt"

func main() {

	bb := make([]byte, 0)

	fmt.Println(bb[:3])

}
```

Reports no errors

```shell
dominic@dominic-XPS-15-9550:~/go/src/github.com/securego/gosec(master)$ ./gosec /home/dominic/go/src/github.com/securego/gosec/cmd/buggy
[gosec] 2023/04/29 09:45:39 Including rules: default
[gosec] 2023/04/29 09:45:39 Excluding rules: default
[gosec] 2023/04/29 09:45:39 Import directory: /home/dominic/go/src/github.com/securego/gosec/cmd/buggy
[gosec] 2023/04/29 09:45:39 Checking package: main
[gosec] 2023/04/29 09:45:39 Checking file: /home/dominic/go/src/github.com/securego/gosec/cmd/buggy/my-buggy-code.go
Results:


Summary:
  Gosec  : dev
  Files  : 1
  Lines  : 11
  Nosec  : 0
  Issues : 0

dominic@dominic-XPS-15-
```

### gosec version

Latest from master

### Go version (output of 'go version')

go version go1.19 linux/amd64

### Operating system / Environment

Ubuntu 18.04

### Expected behavior

Maybe report an error length has not been verified so may cause a panic? 

### Actual behavior

No issue reported.

ccojocar:
Thanks for reporting this. This looks like a good enhancement but probably can be detected only a constant slice bound.

morgenm:
#973

morgenm:
I created this PR for this: #973

ccojocar:
fixed by https://github.com/securego/gosec/pull/973

PR #973, opened 2023-06-17
Feature: G602 Slice Bound Checking

morgenm:
This PR addresses this issue: #954 

This adds a new rule which checks slice bounds access (reslicing and indexing) to ensure that slices are not accessed out of bounds. This only works for slice bounds defined by a `make([]..., len)` or `make([]..., len, capacity)` where len/capacity is defined as a literal integer. For example it works for:
```go
s := make([]float, 10, 20)
```
But not for:
```go
s := make([]float, 10*3)
```
Since we don't evaluate expressions. It also will calculate and validate slice capacities for new slices made by reslicing when possible, such as:
```go
s := make([]float, 10)
x := s[5:]
```
The capacity for the new slice, `x`, will be validated in any subsequent usage. In addition, slices passed to functions are validated when possible by storing function-call-specific capacities. For example:
```go
func main() {
	s := make([]int, 0, 30)
	doStuff(s)
	x := make([]int, 20)
	y := x[10:]
	doStuff(y)
	z := y[5:]
	doStuff(z)
}

func doStuff(x []int) {
	newSlice := x[:10]
	fmt.Println(newSlice)
	newSlice2 := x[:6]
	fmt.Println(newSlice2)
}`
```
will cause an error on creation of `newSlice` and `newSlice2`, since when `z` is passed to it, it is accessed out of bounds.

### DIFF
--- a/rules/slice_bounds.go
+++ b/rules/slice_bounds.go
@@ -42,7 +42,7 @@ func (s *sliceOutOfBounds) Match(node ast.Node, ctx *gosec.Context) (*issue.Issu
 		return s.matchIndexExpr(node, ctx)
 	case *ast.FuncDecl:
 		s.currentFuncName = node.Name.Name
-		s.loadArgCaps(node, ctx)
+		s.loadArgCaps(node)
 	case *ast.CallExpr:
 		sliceMap := make(map[string]*int64)
 		s.sliceCaps[node] = sliceMap
@@ -61,15 +61,15 @@ func (s *sliceOutOfBounds) updateSliceCaps(varName string, caps map[*ast.CallExp
 func (s *sliceOutOfBounds) getAllCalls(funcName string, ctx *gosec.Context) []*ast.CallExpr {
 	calls := []*ast.CallExpr{}
 
-	for callExpr, _ := range s.sliceCaps {
+	for callExpr := range s.sliceCaps {
 		if callExpr != nil {
 			// Compare the names of the function the code is scanning with the current call we are iterating over
-			_, funcName, err := gosec.GetCallInfo(callExpr, ctx)
+			_, callFuncName, err := gosec.GetCallInfo(callExpr, ctx)
 			if err != nil {
 				continue
 			}
 
-			if funcName == s.currentFuncName {
+			if callFuncName == funcName {
 				calls = append(calls, callExpr)
 			}
 		} else {
@@ -111,7 +111,7 @@ func (s *sliceOutOfBounds) setupCallArgCaps(callExpr *ast.CallExpr, ctx *gosec.C
 			// to a function call would catch the most issues, but would require a data structure like a stack and a
 			// reworking of the code for scanning itself. Use the lowest capacity, as this would be more likely to
 			// raise an issue for being out of bounds.
-			var lowestCap *int64 = nil
+			var lowestCap *int64
 			for _, cap := range caps {
 				if lowestCap == nil {
 					lowestCap = cap
@@ -130,7 +130,7 @@ func (s *sliceOutOfBounds) setupCallArgCaps(callExpr *ast.CallExpr, ctx *gosec.C
 			ident := arg.(*ast.Ident)
 			caps := s.getSliceCapsForFunc(funcName, ident.Name, ctx)
 
-			var lowestCap *int64 = nil
+			var lowestCap *int64
 			for _, cap := range caps {
 				if cap == nil {
 					continue
@@ -156,7 +156,7 @@ func (s *sliceOutOfBounds) setupCallArgCaps(callExpr *ast.CallExpr, ctx *gosec.C
 }
 
 // Load caps that were saved for a call to this function
-func (s *sliceOutOfBounds) loadArgCaps(funcDecl *ast.FuncDecl, ctx *gosec.Context) {
+func (s *sliceOutOfBounds) loadArgCaps(funcDecl *ast.FuncDecl) {
 	sliceMap := make(map[string]*int64)
 	funcName := funcDecl.Name.Name
 
@@ -172,8 +172,8 @@ func (s *sliceOutOfBounds) loadArgCaps(funcDecl *ast.FuncDecl, ctx *gosec.Contex
 	}
 
 	for it := range params {
-		cap := argCaps[it]
-		if cap == nil {
+		capacity := argCaps[it]
+		if capacity == nil {
 			continue
 		}
 
@@ -182,7 +182,7 @@ func (s *sliceOutOfBounds) loadArgCaps(funcDecl *ast.FuncDecl, ctx *gosec.Contex
 		}
 
 		paramName := params[it].Names[0].Name
-		sliceMap[paramName] = cap
+		sliceMap[paramName] = capacity
 	}
 
 	s.sliceCaps[nil] = sliceMap
@@ -195,7 +195,7 @@ func (s *sliceOutOfBounds) matchSliceMake(funcCall *ast.CallExpr, sliceName stri
 		return nil, nil
 	}
 
-	capacityArg := 1
+	var capacityArg int
 	if len(funcCall.Args) < 2 {
 		return nil, nil // No size passed
 	} else if len(funcCall.Args) == 2 {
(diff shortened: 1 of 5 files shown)


morgenm:
Note: This solution is complex since it stores the constant capacities of all the slices made with `make()`. I am thinking of adding in a simple check for non-constant capacities (or capacities evaluated from expressions such as `make([]int, 5*5)`) by making sure the user adds in a bounds check such as `if index < cap(slice)`. If it is better to simplify the whole check by only validating that the user has checked bounds, without storing the actual constant capacities in the map as I did in this PR, I can change the code.

Let me know your thoughts.

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+)
+
+type sliceOutOfBounds struct {
+	sliceCaps       map[*ast.CallExpr]map[string]*int64 // Capacities of slices. Maps function call -> var name -> value
+	currentScope    *types.Scope                        // Current scope. Map is cleared when scope changes.
+	currentFuncName string                              // Current function
+	funcCallArgs    map[string][]*int64                 // Caps to load once a func declaration is scanned
+	issue.MetaData
+}
+
```
ccojocar, line 18: Please could you add a comment to this publicly exposed method?

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+func (s *sliceOutOfBounds) ID() string {
+	return s.MetaData.ID
+}
+
```
ccojocar, line 22: ditto

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+	return nil, nil
+}
+
+func (s *sliceOutOfBounds) updateSliceCaps(varName string, caps map[*ast.CallExpr]*int64) {
+	for callExpr, cap := range caps {
+		s.sliceCaps[callExpr][varName] = cap
+	}
+}
+
+// Get all calls for the current function
```
ccojocar, line 60: The godoc format prefixes the comment with the method/function name.

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+			_, callFuncName, err := gosec.GetCallInfo(callExpr, ctx)
+			if err != nil {
+				continue
+			}
+
+			if callFuncName == funcName {
+				calls = append(calls, callExpr)
+			}
+		} else {
+			calls = append(calls, callExpr)
```
ccojocar, line 76: Why do you still append a nil value here?

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+
+			// Simplifying assumption: use the lowest capacity. Storing all possibly capacities for slices passed
+			// to a function call would catch the most issues, but would require a data structure like a stack and a
+			// reworking of the code for scanning itself. Use the lowest capacity, as this would be more likely to
+			// raise an issue for being out of bounds.
+			var lowestCap *int64
+			for _, cap := range caps {
+				if lowestCap == nil {
+					lowestCap = cap
+				} else if *lowestCap > *cap {
```
ccojocar, line 118: Is missing a nil check here? Can cap be nil?

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+			funcCallArgs = append(funcCallArgs, nil)
+		}
+	}
+	s.funcCallArgs[funcName] = funcCallArgs
+}
+
+// Load caps that were saved for a call to this function
+func (s *sliceOutOfBounds) loadArgCaps(funcDecl *ast.FuncDecl) {
+	sliceMap := make(map[string]*int64)
+	funcName := funcDecl.Name.Name
```
ccojocar, line 161: Can the name be nil? e.g. for an anonymous function? I would check it here at least.

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+		capacity := argCaps[it]
+		if capacity == nil {
+			continue
+		}
+
+		if len(params[it].Names) == 0 {
+			continue
+		}
+
+		paramName := params[it].Names[0].Name
```
ccojocar, line 184: I would check `Name[0]` that is not nil before accessing the `Name` filed.

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+		sliceMap[paramName] = capacity
+	}
+
+	s.sliceCaps[nil] = sliceMap
+}
+
+// Matches calls to make() and stores the capacity of the new slice in the map to compare against future slice usage
```
ccojocar, line 191: method name prefix in the comment

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+			caps[callExpr] = &newCap
+		} else if lowIdent == nil { // If no lower bound, capacity will be same
+			continue
+		}
+	}
+
+	return caps
+}
+
+// Matches slice assignments, calculates capacity of slice if possible to store it in map
```
ccojocar, line 257: function name prefix missing in the comment

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+func (s *sliceOutOfBounds) matchSliceAssignment(node *ast.SliceExpr, sliceName string, ctx *gosec.Context) (*issue.Issue, error) {
+	// First do the normal match that verifies the slice expr is not out of bounds
+	if i, err := s.matchSliceExpr(node, ctx); err != nil {
+		return i, err
```
ccojocar, line 261: I would wrap the err with some more context before returning.

rules/slice_bounds.go
```diff
@@ -0,0 +1,373 @@
...
+			if index >= *cap {
+				return ctx.NewIssue(node, s.ID(), s.What, s.Severity, s.Confidence), nil
+			}
+		}
+	}
+
+	return nil, nil
+}
+
+// NewSliceBoundsCheck attempts to find any slices being accessed out of bounds
```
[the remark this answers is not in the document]
ccojocar: Could you please fix the function name in the comment?

ccojocar (commented):
Thank you for this great contribution. I left a few comments. It would be great if you could fix them before we merge this rule. Thanks again!

rules/slice_bounds.go
morgenm: Yes it can; I just wrote some tests and it caused runtime errors because of this. Good catch! I am fixing this now.

ccojocar:
> Note: This solution is complex since it stores the constant capacities of all the slices made with make(). I am thinking of adding in a simple check for non-constant capacities (or capacities evaluated from expressions such as make([]int, 5*5)) by making sure the user adds in a bounds check such as if index < cap(slice). If it is better to simplify the whole check by only validating that the user has checked bounds, without storing the actual constant capacities in the map as I did in this PR, I can change the code.

I think the current solution which checks the constant bounds looks fine to me. It is non intrusive comparing with a bound check which is more enforced on user. 

My impression is, that one will really check the bounds at runtime if she wants to be defensive but most of people won't do it. This will cause a lot of warnings which most likely will get ignored.

rules/slice_bounds.go
morgenm: I was storing functions that were never called by another function (main()) as "nil", then every time a new function declaration was scanned, the current function would become the new nil value to store info about the current call. I now have replaced this with creating dummy calls when loading the arguments of a function declaration, so no need for nil value and it will be less error prone.

rules/slice_bounds.go
morgenm: Anonymous functions would be type FuncLit, so it should never be nil to my understanding, as I noted in an above comment.

rules/slice_bounds.go
morgenm: Okay, I added that check.

rules/slice_bounds.go
morgenm: Okay I wrapped it using `fmt.Errorf`.

morgenm:
I believe I addressed all the reviews. 

> I think the current solution which checks the constant bounds looks fine to me. It is non intrusive comparing with a bound check which is more enforced on user.
> 
> My impression is, that one will really check the bounds at runtime if she wants to be defensive but most of people won't do it. This will cause a lot of warnings which most likely will get ignored.

Yes, thinking on it some more I agree with you.

razor-1:
It would be nice if this didn't fire when it was protected by something like `if len(slice) >= x`

franchb:
Yes, this rule fires a lot of false positives on code protected with something like `if len(slice) >= x`

Issue #885, opened 2022-10-26
No issues reported for secDevLabs (vulnerable apps)


aleisalem:
### Summary 
I am new to using gosec. So, I was trying it against a set of vulnerable go apps in the [secDevLabs](https://github.com/globocom/secDevLabs). The tools reports absolutely no issues although the apps are designed to be vulnerable with some of the vulnerabilities being really obvious (e.g., concatenated SQL statements for SQLi). Am I doing something wrong? Thanks a lot in advance :)

### Steps to reproduce the behavior

1. Clone, for example, the [copy-n-paste](https://github.com/globocom/secDevLabs/blob/master/owasp-top10-2021-apps/a3/copy-n-paste/) app which is vulnerable to SQL injection attacks.
2. Go to the `app/` directory (not sure if necessary but it avoids "failed to import" errors).
3. Run `gosec ./... --verbose` (just to check the vulnerable files are scanned).

### gosec version
v2.14.0

### Go version (output of 'go version')
go1.18.7 linux/amd64

### Operating system / Environment
Ubuntu Linux 22.04

### Expected behavior
I expect to see some security issues like in line [49](https://github.com/globocom/secDevLabs/blob/904a2fb969cb9fb0db30b50709072da8f9d57002/owasp-top10-2021-apps/a3/copy-n-paste/app/util/db.go#L49) according to the rule **G202: SQL query construction using string concatenation**. 

### Actual behavior
The tools reports zero issues:
Summary:
  Gosec  : v2.14.0
  Files  : 6
  Lines  : 391
  Nosec  : 0
  Issues : 0

ccojocar:
I think on a first look that we should detect this issue, but it seems that somehow the SQL rule doesn't. This looks like a bug to me.

audunmo:
Might be that some of these apps from secDevLabs could be good to benchmark the performance of gosec?

audunmo:
I have a hypothesis of why this is happening, will try to implement a fix

audunmo:
So I've been looking a bit at why `gosec` doesn't flag the SQL injections in secDevLabs. The hypothesis I've been working from is that the failure arises because of how `gosec` handles SQLi via format calls seperately from how it handles SQLi via string concatenations. I still haven't properly confirmed this hypothesis, nor have I found a good fix to the problem.

[...]

PR #987, opened 2023-07-21
fix: correctly identify infixed concats as potential SQL injections

audunmo:
This doesn't have an issue, popped up while I was investigating #885 

In some cases, gosec fails to fully evaluate statements where string concatenations contain variable components. This leads to it missing potential SQL injection vulnerabilities that are "infixed", i.e. `"INSERT INTO foo VALUES (name) ('" + os.Args[0] + "')"`

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -114,21 +114,17 @@ func GetStringRecursive(n ast.Node) (string, error) {
 	}
 
 	if expr, ok := n.(*ast.BinaryExpr); ok {
-		var err error
-		x, xerr := GetStringRecursive(expr.X)
-		// TODO: replace this error formatting strangeness with errors.Join when gosec is bumped to go 1.20
-		if xerr != nil {
-			err = fmt.Errorf("Error on X branch in recursion: %w", xerr)
+		x, err := GetStringRecursive(expr.X)
+		if err != nil {
+			return "", err
 		}
 
-		y, yerr := GetStringRecursive(expr.Y)
-		if yerr != nil {
-			// Linter dislikes using non-%w formatting verbs for errors, but also doesn't like formatting with two %w statements.
-			// So we just add the string representation of the underlying error
-			err = fmt.Errorf("%w Error on Y branch in recursion: %v", err, yerr.Error())
+		y, err := GetStringRecursive(expr.Y)
+		if err != nil {
+			return "", err
 		}
 
-		return x + y, err
+		return x + y, nil
 	}
 
 	return "", nil


rules/rulelist.go
```diff
@@ -79,8 +79,8 @@ func Generate(trackSuppressions bool, filters ...RuleFilter) RuleList {
 		{"G114", "Use of net/http serve function that has no support for setting timeouts", NewHTTPServeWithoutTimeouts},
 
 		// injection
-		{"G201", "SQL query construction using format string", NewSQLStrFormat},
 		{"G202", "SQL query construction using string concatenation", NewSQLStrConcat},
+		{"G201", "SQL query construction using format string", NewSQLStrFormat},
```
audunmo, line 83: Woops, this is left over from an experiment. Will fix

testutils/source.go
```diff
@@ -1807,7 +1838,8 @@ func main() {
 		panic(err)
 	}
 	fmt.Println(result)
-}`}, 1, gosec.NewConfig()}, {[]string{`
+}`}, 1, gosec.NewConfig()},
```
audunmo, line 1841: These are results of `gofumpting` the file. Am I missing some config, or did `gofumpt` get more litigious?

helpers.go
```diff
@@ -96,11 +96,48 @@ func GetChar(n ast.Node) (byte, error) {
...
+func GetStringRecursive(n ast.Node) (string, error) {
+	if node, ok := n.(*ast.BasicLit); ok && node.Kind == token.STRING {
+		return strconv.Unquote(node.Value)
+	}
+
+	if expr, ok := n.(*ast.BinaryExpr); ok {
+		var err error
+		x, xerr := GetStringRecursive(expr.X)
+		if xerr != nil {
+			err = fmt.Errorf("%w Error on X branch in recursion: %v", err, xerr)
```
ccojocar, line 121: This error wrapping looks strange since `err` has no value initialised. Please use `errors.New` instead.

ccojocar: If an error occurs on the left side why do not stop recurring since in the end you will always return an error?

helpers.go
```diff
@@ -96,11 +96,48 @@ func GetChar(n ast.Node) (byte, error) {
...
+		}
+
+		y, yerr := GetStringRecursive(expr.Y)
+		if yerr != nil {
+			err = fmt.Errorf("%w Error on Y branch in recursion: %v", err, err)
```
ccojocar, line 126: here you need the `yerr` instead of twice `err`

ccojocar (changes requested):

ccojocar:
Also please fix the lint warnings. You can run the golanci-lint locally.

audunmo:
@ccojocar thanks for the review, will fix

helpers.go
audunmo: Good catch! Makes sense to just return on a left-side err

ccojocar (approved):

PR #1020, opened 2023-09-25
Add a new rule which detects when a file is created with os.Create but the configured permissions are less than 0666

### LINKED ISSUE #1019: G306 can be easily bypassed by os.Create
os.Create creates file with permission 0666.
https://pkg.go.dev/os#Create

os.Create is usually preffered to create & truncate a file so it can accidentally bypass G306 even if developers want to enforce stricter permissions.

I'm not sure what can be a reasonable behavior. Reporting os.Create if allowed permission is stricter than 0666 can be an instant one.

ccojocar:
It seems that the os.Create will create by default a file with 0666 permissions.

This should be detected when the configured permissions are less than 0666. By default will not detect this case
unless the more restrictive mode is configured.

fixes #1019

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,7 @@ directory you can supply `./...` as the input argument.
 - G304: File path provided as taint input
 - G305: File traversal when extracting zip/tar archive
 - G306: Poor file permissions used when writing to a new file
+- G307: Poor file permissions used when crating a file with os.Create
 - G401: Detect the usage of DES, RC4, MD5 or SHA1
 - G402: Look for bad TLS connection settings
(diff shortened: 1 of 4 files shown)


ldez:
Hello @ccojocar,

Is it expected to reuse a retire rule code (G307)?

https://github.com/securego/gosec/blob/3952187ea76579f7b405e90336a90a56114a4119/README.md?plain=1#L160

https://github.com/securego/gosec/blob/3952187ea76579f7b405e90336a90a56114a4119/README.md?plain=1#L176

Issue #1032, opened 2023-10-10
Latest version 2.18.0 (master branch) is getting panic


patels30:
### Summary 
We are getting a panic issue on 2.18.0 version. 2.17.0 is working fine.

```
[gosec] 2023/10/10 04:57:21 Including rules: default
[gosec] 2023/10/10 04:57:21 Excluding rules: default
[gosec] 2023/[10](https://github.com/***/****/actions/runs/*******/job/*******?pr=1649#step:6:11)/10 
[gosec] 2023/10/10 04:57:35 Import directory: /github/workspace/internal/temporal/activity
panic: interface conversion: types.Type is *types.Basic, not *types.Signature
```

### gosec version
2.18.0

### Go version (output of 'go version')
go1.21.1

### Operating system / Environment
ubuntu22-6

ccojocar:
Could you please provide a code sample with which I can reproduce the issue? Thanks

ilbambino:
we are also getting panic.

`go1.21.1 linux/amd64` and `securego/gosec@master` running from a GitHub action 
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x38 pc=0x62c55d]

goroutine 1 [running]:
golang.org/x/tools/go/ssa.(*Program).packageLevelMember(...)
	/go/pkg/mod/golang.org/x/tools@v0.14.0/go/ssa/source.go:186
golang.org/x/tools/go/ssa.(*builder).expr0(0xc000bb7648, 0xc0006a5b00, {0x8e54b0?, 0xc0000d1320}, {0x0, {0x0, 0x0}, {0x0, 0x0}})
	/go/pkg/mod/golang.org/x/tools@v0.14.0/go/ssa/builder.go:796 +0x123d
golang.org/x/tools/go/ssa.(*builder).expr(0x7ccac0?, 0xc0006a5b00, {0x8e54b0, 0xc0000d1320})
	/go/pkg/mod/golang.org/x/tools@v0.14.0/go/ssa/builder.go:623 +0x175
golang.org/x/tools/go/ssa.(*builder).expr0(0xc000bb7648, 0xc0006a5b00, {0x8e5540?, 0xc000416cc0}, {0x0, {0x0, 0x0}, {0x0, 0x0}})
	/go/pkg/mod/golang.org/x/tools@v0.14.0/go/ssa/builder.go:818 +0x18d1
[...]

PR #1034, opened 2023-10-10
add ssa enable check to avoid panic in analyzer

mmsqe:
update https://github.com/securego/gosec/issues/1032


ccojocar:
Thanks for this but this is not a fix. 

This flag was used to disable all analysers as long they were in the development but now they  are enabled by default. 

Please could you open an issue with the panic trace and with the code sample which generates the panic? Thanks

Issue #622, opened 2021-05-20
invalid package name "" when importing tview


SamWhited:
### Summary 

When running gosec against a package on alpine/edge (but not on Feora 34) I get the error:

```
  > [line 26 : column 2] - could not import github.com/rivo/tview (invalid package name: "")
```

### Steps to reproduce the behavior

```
go get -u github.com/securego/gosec/cmd/gosec
cd project-that-imports-tview/
gosec ./...
```

### gosec version

Always pulled in CI using `go get` as above. Last run I see was:

```
go: downloading github.com/securego/gosec v0.0.0-20200401082031-e946c8c39989
```

### Go version (output of 'go version')

```
+ go version
go version go1.16.4 linux/amd64
```

### Operating system / Environment

alpine/edge

```
+ go env
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/build/.cache/go-build"
GOENV="/home/build/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/build/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/build/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/lib/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.16.4"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/dev/null"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build254833280=/tmp/go-build -gno-record-gcc-switches"
```

### Expected behavior

It should run like it used to before this project imported tview.

### Actual behavior

```
  > [line 26 : column 2] - could not import github.com/rivo/tview (invalid package name: "")
```

This is similar to #580 except not in the standard library so I'm opening this separately.

[...]

Issue #501, opened 2020-07-04
hitting error (invalid package name: "") on repo have multiple go.mod


WLun001:
### Summary 
Hitting error `invalid package name: ""` on repo with multiple `go.mod`

### Steps to reproduce the behavior

Repo strucutre
```
.root
  |- cli 
  |   | - cli-name 
  |       |- go.mod
  | - go.mod
```
Run `gosec` on root
```sh
gosec ./...
```

after awhile

```sh
Golang errors in file: [some-path/cmd/generate.go]:

  > [line 21 : column 2] - could not import github.com/spf13/cobra (invalid package name: "")

```

### gosec version
Not sure what is the version, it prints nothing
```sh
gosec -version

Version:
Git tag:
Build date:
```

### Go version (output of 'go version')
```sh
go version go1.14.3 darwin/amd64
```
### Operating system / Environment
MacOS 10.15.4

### Expected behavior
Should able to resolve packages

### Actual behavior
Unable to resolve packages on repo with multiple `go.mod`

ccojocar:
Hi @WLun001, This is not an idiomatic go project layout. You need to have one module per repository. See this docs for more details https://github.com/golang-standards/project-layout. 

Also note that gosec relies on go standard library packages to load the modules.  So if it fails, it means that your project doesn't follow the standards.

I am going to close this because gosec is not going to support this project structure. Thanks!

WLun001:
@ccojocar thanks for the suggestion, I have refactor my project layout and it worked!

nitinvalake:
@ccojocar Facing same issue with private repos.

In my project we have used private repositories and while running gosec command it is giving "unknown revision" error for those private repos.

lootek:
fd

> Hi @WLun001, This is not an idiomatic go project layout. You need to have one module per repository. See this docs for more details https://github.com/golang-standards/project-layout.
> 
> Also note that gosec relies on go standard library packages to load the modules. So if it fails, it means that your project doesn't follow the standards.
> 
> I am going to close this because gosec is not going to support this project structure. Thanks!

The repo you're referring to is NOT an official Go project structure - see https://github.com/golang-standards/project-layout/issues/117

Also, it's perfectly valid to have one git repo with many Go modules - see the OFFICIAL docs https://golang.org/doc/modules/managing-source#tools

Emptyless:
for anyone facing this issue, #1100 fixes the aforementioned issue, the PR is closed but I'll keep the branch open and will try to keep it in sync

PR #1100, opened 2024-02-09
support multi module project

Emptyless:
support multi module project by using the build system to locate the go.mod if present. Related to #501 #622

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -337,8 +337,14 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
 	var goModDir string
 	if os.Getenv("DISABLE_MULTI_MODULE_MODE") != "true" {
 		if modPkgs, err := packages.Load(&packages.Config{Mode: packages.NeedModule, Dir: abspath}, abspath); err == nil && len(modPkgs) == 1 {
-			goModFile = modPkgs[0].Module.GoMod
-			goModDir = path.Dir(goModFile)
+			if len(modPkgs[0].Errors) != 0 {
+				for _, modPkgErr := range modPkgs[0].Errors {
+					gosec.logger.Printf("Skipping multi module mode for '%s': %s", pkgPath, modPkgErr)
+				}
+			} else {
+				goModFile = modPkgs[0].Module.GoMod
+				goModDir = path.Dir(goModFile)
+			}
 		}
 	}
 
--- a/analyzer.go
+++ b/analyzer.go
@@ -333,12 +333,25 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
 		return []*packages.Package{}, fmt.Errorf("importing dir %q: %w", pkgPath, err)
 	}
 
-	var packageFiles []string
-	for _, filename := range basePackage.GoFiles {
-		packageFiles = append(packageFiles, path.Join(pkgPath, filename))
+	var goModFile string
+	goModDir := abspath[0 : len(abspath)-len(pkgPath)] // get root dir
+	if modPkgs, err := packages.Load(&packages.Config{Mode: packages.NeedModule, Dir: abspath}, abspath); err == nil && len(modPkgs) == 1 {
+		goModFile = modPkgs[0].Module.GoMod
+		goModDir = path.Dir(goModFile)
 	}
-	for _, filename := range basePackage.CgoFiles {
-		packageFiles = append(packageFiles, path.Join(pkgPath, filename))
+
+	var packageFiles []string
+	for _, filename := range append(basePackage.GoFiles, basePackage.CgoFiles...) {
+		if goModDir == "" {
+			packageFiles = append(packageFiles, path.Join(pkgPath, filename))
+		} else {
+			filePath := path.Join(abspath, filename)
+			relPath, err := filepath.Rel(goModDir, filePath)
+			if err != nil {
+				return []*packages.Package{}, fmt.Errorf("get relative path between %q and %q: %w", goModDir, filePath, err)
+			}
+			packageFiles = append(packageFiles, relPath)
+		}
 	}
 
 	if gosec.tests {
@@ -354,6 +367,7 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
 	gosec.mu.Lock()
 	conf.BuildFlags = nil
 	defer gosec.mu.Unlock()
+	conf.Dir = goModDir
 	pkgs, err := packages.Load(conf, packageFiles...)
 	if err != nil {
 		return []*packages.Package{}, fmt.Errorf("loading files from package %q: %w", pkgPath, err)
--- a/analyzer.go
+++ b/analyzer.go
@@ -376,9 +376,16 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
 	gosec.mu.Lock()
 	conf.BuildFlags = nil
 	defer gosec.mu.Unlock()
+
+	// set conf.Dir and reset back to original value
+	confDir := conf.Dir
 	if goModDir != "" {
 		conf.Dir = goModDir
+		defer func() {
+			conf.Dir = confDir
+		}()
 	}
+
 	pkgs, err := packages.Load(conf, packageFiles...)
 	if err != nil {
 		return []*packages.Package{}, fmt.Errorf("loading files from package %q: %w", pkgPath, err)
--- a/analyzer.go
+++ b/analyzer.go
@@ -335,9 +335,11 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
 
 	var goModFile string
 	goModDir := abspath[0 : len(abspath)-len(pkgPath)] // get root dir
-	if modPkgs, err := packages.Load(&packages.Config{Mode: packages.NeedModule, Dir: abspath}, abspath); err == nil && len(modPkgs) == 1 {
-		goModFile = modPkgs[0].Module.GoMod
-		goModDir = path.Dir(goModFile)
+	if os.Getenv("DISABLE_MULTI_MODULE_MODE") != "true" {
+		if modPkgs, err := packages.Load(&packages.Config{Mode: packages.NeedModule, Dir: abspath}, abspath); err == nil && len(modPkgs) == 1 {
+			goModFile = modPkgs[0].Module.GoMod
+			goModDir = path.Dir(goModFile)
+		}
 	}
 
 	var packageFiles []string
@@ -348,7 +350,8 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
 			filePath := path.Join(abspath, filename)
 			relPath, err := filepath.Rel(goModDir, filePath)
 			if err != nil {
-				return []*packages.Package{}, fmt.Errorf("get relative path between %q and %q: %w", goModDir, filePath, err)
+				gosec.logger.Printf("Skipping: %s. Cannot get relative path between %q and %q: %s", filename, goModDir, filePath, err)
+				continue
 			}
 			packageFiles = append(packageFiles, relPath)
 		}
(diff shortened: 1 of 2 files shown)


analyzer.go
```diff
@@ -333,12 +333,25 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
 		return []*packages.Package{}, fmt.Errorf("importing dir %q: %w", pkgPath, err)
 	}
 
-	var packageFiles []string
-	for _, filename := range basePackage.GoFiles {
-		packageFiles = append(packageFiles, path.Join(pkgPath, filename))
+	var goModFile string
+	goModDir := abspath[0 : len(abspath)-len(pkgPath)] // get root dir
+	if modPkgs, err := packages.Load(&packages.Config{Mode: packages.NeedModule, Dir: abspath}, abspath); err == nil && len(modPkgs) == 1 {
```
ccojocar, line 338: Is this only loading the module files, or does parse all the package as well? Please could you double check? I want to avoid to introduce performance issues?

analyzer.go
```diff
@@ -333,12 +333,25 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
...
+		goModFile = modPkgs[0].Module.GoMod
+		goModDir = path.Dir(goModFile)
 	}
-	for _, filename := range basePackage.CgoFiles {
-		packageFiles = append(packageFiles, path.Join(pkgPath, filename))
+
+	var packageFiles []string
+	for _, filename := range append(basePackage.GoFiles, basePackage.CgoFiles...) {
+		if goModDir == "" {
```
ccojocar, line 345: Is this ever going to be empty since it has a default value assigned at line 337?

analyzer.go
```diff
@@ -333,12 +333,25 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
...
+			packageFiles = append(packageFiles, path.Join(pkgPath, filename))
+		} else {
+			filePath := path.Join(abspath, filename)
+			relPath, err := filepath.Rel(goModDir, filePath)
+			if err != nil {
+				return []*packages.Package{}, fmt.Errorf("get relative path between %q and %q: %w", goModDir, filePath, err)
```
ccojocar, line 351: This will fail the whole scanning in case the relative path of one file cannot be looked-up. This is impractical in very large projects. I would log the error, and just skip the file if it cannot find the relative path.

analyzer.go
ccojocar: Why is this `1`? What if there are more than 1 module files?

ccojocar:
Please try to add some tests as well in the analyzer_test.go for a multi-module project.

analyzer.go
Emptyless: I'll double check on performance but my assumption was with only NeedModule is that it skips the remaining parsing , the check is for `1` given that it should only find a single go.mod for the given source file - or there are go files with different package identifiers in the same directory (which is not valid).

analyzer.go
Emptyless: the default value could ben empty if the pkgPath and abspath are equal

Emptyless:
> Please try to add some tests as well in the analyzer_test.go for a multi-module project.

will add them tonight or tomorrow

analyzer.go
Emptyless: Fixed in latest commit

analyzer.go
Emptyless: Added benchmarks, see below

analyzer.go
```diff
@@ -333,12 +333,28 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
 		return []*packages.Package{}, fmt.Errorf("importing dir %q: %w", pkgPath, err)
 	}
 
-	var packageFiles []string
-	for _, filename := range basePackage.GoFiles {
-		packageFiles = append(packageFiles, path.Join(pkgPath, filename))
+	var goModFile string
+	goModDir := abspath[0 : len(abspath)-len(pkgPath)] // get root dir
+	if os.Getenv("DISABLE_MULTI_MODULE_MODE") != "true" {
```
ccojocar, line 338: Doesn't the multi-module project requires a workspace file? https://go.dev/ref/mod#workspaces

The switch from single module to multi module project is controlled by GOWORK environment variable and this should be already supported inside of Load function.

analyzer.go
```diff
@@ -333,12 +333,28 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
...
+		if modPkgs, err := packages.Load(&packages.Config{Mode: packages.NeedModule, Dir: abspath}, abspath); err == nil && len(modPkgs) == 1 {
```
ccojocar, line 339: What if you run the scan in the workspace folder? There are multiple modules, each module with multiple packages. As I understand here, there will be loaded just one of this modules? Is the right module of the package?

analyzer_test.go
```diff
@@ -124,6 +126,36 @@ var _ = Describe("Analyzer", func() {
...
+				func main(){
+				}`)
+			pkg2.AddFile("go.mod", `
+				module github.com/securego/gosec/v2/test2
+			`)
+			pkg2.AddFile("bar.go", `
+				package main
+				func bar(){
+				}`)
+			err := pkg1.Build()
```
ccojocar, line 149: This doesn't look like a multi-module project. You are missing the `go.work` file. See the `Workspaces` for details https://go.dev/ref/mod#workspaces.

analyzer.go
Emptyless: It is possible to use the replace directive in a `go.mod`. See: https://go.dev/ref/mod#go-mod-file-replace

`go.work` was introduced later in go1.18 to simplify multi-module projects but is not mandated iirc

Emptyless:
> The NeedModule is already into the configuration. See
> 
> https://github.com/securego/gosec/blob/c824a5d308340f4aa39cd05fc6da76549d7fd11a/analyzer.go#L243
> 
> There is no need to load it separately. The loader will go back into the folder looking for it. You can easily verify this by lunching a scan inside of a package without a go.mod file.

the problem is that the package files are not referenced relative to the go.mod so an additional pass is needed in order to load them correctly. The reason your example works is that never becomes an issue but it does with nested go.mod files.

```
.
├── go.mod
├── manual.go
└── some
   ├── go.mod
   ├── generated.go
   └── nested
       ├── main.go
       └── go.mod

3 directories, 6 files
```

This will cause errors as the files of `nested` are loaded as `./some/nested/main.go` for package `nested` instead of `main.go`. This will cause the packages tool to fail and report (e.g. for some random imported package): 

```
Golang errors in file: [gosec/testmultimodule/some/nested/main.go]:

  > [line 4 : column 2] - could not import github.com/joho/godotenv (invalid package name: "")
```

I've attached the directory structure above if you want to reproduce this behavior, just put it in the root of the gosec project
[testmultimodule.zip](https://github.com/securego/gosec/files/14310849/testmultimodule.zip)

it works with `gosec -stdout -verbose=text ./...`, it doesnt work with `DISABLE_MULTI_MODULE_MODE=true gosec -stdout -verbose=text ./...`

analyzer.go
Emptyless: I have added a message that if run in a `go.work` environment, the message will be logged and it will fall back to default behaviour. Since then we rely on the go.work the `conf.Dir` has to be reset while it still has the lock. This will support `go.work` behaviour even when not alle modules are used in the `go.work` file.

analyzer_test.go
Emptyless: the test was intended to mock a multi-module project without a go.work file, this relies on the published version of a module rather than local filesystem version. This could be the case if within a project it is decided not to commit the `go.work` version such that CI picks up on versioning issues.

ccojocar:
> the problem is that the package files are not referenced relative to the go.mod so an additional pass is needed in order to load them correctly. The reason your example works is that never becomes an issue but it does with nested go.mod files.


This doesn't look like a correct multi-module project in Go. Please have a look at Workspaces in the documentation https://go.dev/ref/mod#workspaces. 

If you set up correctly the workspace folder, it should work fine:

```
.
├── go.mod
├── go.work
├── manual.go
└── some
    ├── generated.go
    ├── go.mod
    └── nested
        ├── go.mod
        ├── go.sum
        └── main.go

3 directories, 8 file
```

go.work file:
```
go 1.22.0

use (
        .
        ./some
        ./some/nested
)
```

And the scan is successful:

```
[gosec] 2024/02/16 14:24:21 Including rules: default
[gosec] 2024/02/16 14:24:21 Excluding rules: default
[gosec] 2024/02/16 14:24:21 Import directory: /Users/gcojocar/Downloads/testmultimodule/some
[gosec] 2024/02/16 14:24:21 Import directory: /Users/gcojocar/Downloads/testmultimodule/some/nested
[gosec] 2024/02/16 14:24:21 Import directory: /Users/gcojocar/Downloads/testmultimodule
[gosec] 2024/02/16 14:24:21 Checking package: some
[gosec] 2024/02/16 14:24:21 Checking file: /Users/gcojocar/Downloads/testmultimodule/some/generated.go
[gosec] 2024/02/16 14:24:21 Checking package: testmultimodule
[gosec] 2024/02/16 14:24:21 Checking file: /Users/gcojocar/Downloads/testmultimodule/manual.go
[gosec] 2024/02/16 14:24:22 Checking package: main
[gosec] 2024/02/16 14:24:22 Checking file: /Users/gcojocar/Downloads/testmultimodule/some/nested/main.go
Results:


Summary:
  Gosec  : dev
  Files  : 3
  Lines  : 15
  Nosec  : 0
  Issues : 0
  ```

It seems that these changes proposed here are not required. Thank you

Emptyless:
That workflow will fail in a CI environment which is an example of where we run `gosec`. See https://github.com/golang/go/issues/62179 on why the `go.work` file should not be committed. I'd prefer to support both flows, local (where e.g. a `go.work` file is present) and vcs (without `go.work`)

ccojocar:
If you don't want to check it in, you can either generate it on the fly or scan each module folder separately. 

Thanks for contributing it but at this time I don't plan to merge it because it doesn't seem canonical to me and brings a performance penalty since most of the projects doesn't need this feature.

Emptyless:
thanks for your time and review but I think it is shortsighted to close based on a performance penalty of 0,03ms and imho pretty canonical lines and to **force a project structure which is advocated against in the Go repository**. I could make it opt-in behaviour based and invert the env flag if you are truly worried about this. 

There have been two issues already where you tried to enforce the same directory layout which was also downvoted by the community (see e.g. https://github.com/securego/gosec/issues/501#issuecomment-653725474).

ccojocar:
It doesn't seems to be consensus about this in the Go community. When there will be a clear solution, gosec will definitely adopt it. 

If you scan a monorepo, it makes perfect sense to check it in. See the discussion in https://github.com/golang/go/issues/53502#issuecomment-1204134618.

Emptyless:
I've replaced the strong wording, a better wording would be "generally recommended against". 

> It doesn't seem to be consensus about this in the Go community. When there will be a clear solution, gosec will definitely adopt it.
> 
> If you scan a monorepo, it makes perfect sense to check it in. See the discussion in [golang/go#53502 (comment)](https://github.com/golang/go/issues/53502#issuecomment-1204134618).

I think there is a misunderstanding here, in a monorepo is it mentioned that this is fine; later in the thread even mentioned that this should _definitely_ be checked in (https://github.com/golang/go/issues/53502#issuecomment-1939324628). But this is an exception to the general recommendation, the full quote (you linked) is:

> **go.work files should generally not be checked into VCS**, as their purpose is to develop multiple modules at the same time, and each module tends to have its own VCS repository. However, there are certain scenarios where it can make sense to check in go.work files, such as a monorepo holding multiple modules which depend on each other.

would you want `gosec` to **only** be able to scan the exceptions to the general recommendation rather than both situations?

Emptyless:
Given that there is now consensus in the Go community ("It is generally inadvisable to commit go.work files into version control systems", see golang/website@80af5f5f42a708fad470699f7f3fe7eb1d1e6851), please reconsider the PR

Issue #1090, opened 2023-12-21
G601 should not be raised when GOEXPERIMENT=loopvar is enabled


Hades32:
### Summary 
The underlying issue will be fixed in Go 1.22 and is available in Go 1.21 with the `GOEXPERIMENT=loopvar` flag enabled

### Steps to reproduce the behavior
```Go
		for _, item := range items {
			cpItem = &item
			break
		}
```

### gosec version
whatever is included in the latest golang-ci

### Go version (output of 'go version')
go version go1.21.0 linux/amd64

elisarver:
This issue also applies to go 1.22.0

PR #1108, opened 2024-02-21
Ignore 'implicit memory aliasing' rule for Go 1.22+

VirrageS:
After the recent changes in language in Go 1.22 we no longer need to check for "implicit memory aliasing" rule. Doing:
```
for _, item := range items {
	cpItem = &item
	break
}
```
is now valid because `item` is a new variable in memory for every iteration (and not as it was before the same variable in memory what caused all the problems).

Closes: https://github.com/securego/gosec/issues/1090


ccojocar (approved):

SuperSandro2000:
Can you do a release, so that this gets updated in golangci-lint?

PR #1114, opened 2024-03-07
Use standard function to check if a file is generated

nobishino:
As of Go1.21, we can utilize standard package function [`ast.IsGenerated`](https://pkg.go.dev/go/ast#IsGenerated) to check if a file is generated or not. This PR replace `isGeneratedFile` with that new standard function.

Second commit updates test source. 
Standard `ast.IsGenerated` returns `false` for original test sources, because it does not comply with
https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source

> This line must appear before the first non-comment, non-blank text in the file.

Original test source does not satisfy this condition, so original `isGeneratedFile` function returns `true` when it should return `false`.

I would assume we should update the test cases, but this change can be breaking for some users. I wonder if we should do that.

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -55,8 +55,6 @@ const externalSuppressionJustification = "Globally suppressed."
 
 const aliasOfAllRules = "*"
 
-var generatedCodePattern = regexp.MustCompile(`^// Code generated .* DO NOT EDIT\.$`)
-
 type ignore struct {
 	start        int
 	end          int
--- a/analyzer.go
+++ b/analyzer.go
@@ -31,11 +31,12 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/securego/gosec/v2/analyzers"
-	"github.com/securego/gosec/v2/issue"
 	"golang.org/x/tools/go/analysis"
 	"golang.org/x/tools/go/analysis/passes/buildssa"
 	"golang.org/x/tools/go/packages"
+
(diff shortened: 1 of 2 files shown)


nobishino:
I noticed that Ci is failing, I will look into it later.

ccojocar (approved):
Thanks for this update. Please could you fix the lint warning?

nobishino:
@ccojocar I've tried to fix it, but I don't understand the cause of [the lint error](https://github.com/securego/gosec/actions/runs/8189381874/job/22394017030?pr=1114):

```
  Error: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/securego) (gci)
  Error: File is not `gci`-ed with --skip-generated -s standard -s default -s prefix(github.com/securego) (gci)
  
  Error: issues found
```

I would appreciate any advice.

ccojocar:
I think you need to run the [gci](https://github.com/daixiang0/gci) tool to format the two files. Something like this:
```
 gci write --skip-generated -s standard -s default -s prefix(github.com/securego)  analyzer.go
```
You can check if the warnings are fixed by running locally the golangci-lint tool.

ccojocar:
@nobishino I did the formatting. Let's see if now passes the lint.

nobishino:
@ccojocar Thanks!

PR #1149, opened 2024-05-27
Add a new rule to detect integer overflow when converting between integer types

### LINKED ISSUE #1130: Add detection of overflow during integer conversion
### Summary 

G109 provides detection of overflow during strconv.Atoi, but there seems to be no good way to detect the more common conversions between integers, such as uint32 to int32.

ccojocar:
fixes #1130

### DIFF
--- a/analyzers/conversion_overflow.go
+++ b/analyzers/conversion_overflow.go
@@ -1,0 +1,134 @@
+// (c) Copyright gosec's authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package analyzers
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+
+	"github.com/securego/gosec/v2/issue"
+	"golang.org/x/tools/go/analysis"
+	"golang.org/x/tools/go/analysis/passes/buildssa"
+	"golang.org/x/tools/go/ssa"
+)
+
+func newConversionOverflowAnalyzer(id string, description string) *analysis.Analyzer {
+	return &analysis.Analyzer{
+		Name:     id,
+		Doc:      description,
+		Run:      runConversionOverflow,
+		Requires: []*analysis.Analyzer{buildssa.Analyzer},
+	}
+}
+
+func runConversionOverflow(pass *analysis.Pass) (interface{}, error) {
+	ssaResult, err := getSSAResult(pass)
+	if err != nil {
+		return nil, fmt.Errorf("building ssa representation: %w", err)
+	}
+
+	issues := []*issue.Issue{}
+	for _, mcall := range ssaResult.SSA.SrcFuncs {
+		for _, block := range mcall.DomPreorder() {
+			for _, instr := range block.Instrs {
+				switch instr := instr.(type) {
+				case *ssa.Convert:
+					src := instr.X.Type().String()
+					dst := instr.Type().String()
+					if isIntOverflow(src, dst) {
+						issue := newIssue(pass.Analyzer.Name,
+							fmt.Sprintf("integer overflow conversion %s -> %s", src, dst),
+							pass.Fset,
+							instr.Pos(),
+							issue.High,
+							issue.Medium,
+						)
+						issues = append(issues, issue)
+					}
+				}
+			}
+		}
+	}
+
+	if len(issues) > 0 {
+		return issues, nil
+	}
+	return nil, nil
+}
+
+type integer struct {
+	signed bool
+	size   int
+}
+
+func parseIntType(intType string) (integer, error) {
+	re := regexp.MustCompile(`(?P<type>u?int)(?P<size>\d{2})?`)
+	matches := re.FindStringSubmatch(intType)
+	if matches == nil {
+		return integer{}, fmt.Errorf("no integer type match found for %s", intType)
+	}
+
+	it := matches[re.SubexpIndex("type")]
+	is := matches[re.SubexpIndex("size")]
+
+	signed := false
+	if it == "int" {
+		signed = true
+	}
+
+	// use default system int type in case size is not present in the type
+	intSize := strconv.IntSize
+	if is != "" {
+		var err error
+		intSize, err = strconv.Atoi(is)
+		if err != nil {
+			return integer{}, fmt.Errorf("failed to parse the integer type size: %w", err)
+		}
+	}
+
+	return integer{signed: signed, size: intSize}, nil
+}
+
+func isIntOverflow(src string, dst string) bool {
+	srcInt, err := parseIntType(src)
+	if err != nil {
+		return false
+	}
+
+	dstInt, err := parseIntType(dst)
+	if err != nil {
+		return false
+	}
+
+	// converting uint to int of the same size or smaller might lead to overflow
+	if !srcInt.signed && dstInt.signed && dstInt.size <= srcInt.size {
+		return true
+	}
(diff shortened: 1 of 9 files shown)


remyleone:
Could you add documentation about how to fix it and make the linter pass? 🙏

analyzers/conversion_overflow.go
```diff
@@ -0,0 +1,135 @@
...
+	for _, mcall := range ssaResult.SSA.SrcFuncs {
+		for _, block := range mcall.DomPreorder() {
+			for _, instr := range block.Instrs {
+				switch instr := instr.(type) {
+				case *ssa.Convert:
+					src := instr.X.Type().String()
+					dst := instr.Type().String()
+					if isIntOverflow(src, dst) {
+						issue := newIssue(pass.Analyzer.Name,
+							fmt.Sprintf("integer overflow conversion %s -> %s", src, dst),
```
ccoVeille, line 54: I might have missed something about the implementation, but I would like to suggest this


```suggestion
							fmt.Sprintf("possible integer overflow conversion %s -> %s", src, dst),
```

It might not happen depending on people code.

It's better to report a possible error, than affirming there is one. People would argue.

analyzers/conversion_overflow.go
```diff
@@ -0,0 +1,135 @@
...
+	return nil, nil
+}
+
+type integer struct {
+	signed bool
+	size   int
+}
+
+func parseIntType(intType string) (integer, error) {
+	re := regexp.MustCompile(`(?P<type>u?int)(?P<size>\d{2})?`)
```
ccoVeille, line 79: This regexp will be computed each time parseInt will be called, it should be moved out the function

analyzers/conversion_overflow.go
```diff
@@ -0,0 +1,135 @@
...
+	matches := re.FindStringSubmatch(intType)
+	if matches == nil {
+		return integer{}, fmt.Errorf("no integer type match found for %s", intType)
+	}
+
+	it := matches[re.SubexpIndex("type")]
+	is := matches[re.SubexpIndex("size")]
```
ccoVeille, line 86: Please add something like this

```suggestion
	// int64 => int 64, uint8 => uint 8
	it := matches[re.SubexpIndex("type")]
	is := matches[re.SubexpIndex("size")]
```

analyzers/conversion_overflow.go
```diff
@@ -0,0 +1,135 @@
...
+	// converting uint to unit of a smaller size might lead to overflow
+	if !srcInt.signed && !dstInt.signed && dstInt.size < srcInt.size {
+		return true
+	}
+	// converting int to int of a smaller size might lead to overflow
+	if srcInt.signed && dstInt.signed && dstInt.size < srcInt.size {
+		return true
+	}
+	// converting int to uint of a smaller size might lead to overflow
+	if srcInt.signed && !dstInt.signed && dstInt.size < srcInt.size && srcInt.size-dstInt.size > 8 {
```
ccoVeille, line 130: This 8 is magical number here

analyzers/conversion_overflow.go
```diff
@@ -0,0 +1,135 @@
...
+		intSize, err = strconv.Atoi(is)
+		if err != nil {
+			return integer{}, fmt.Errorf("failed to parse the integer type size: %w", err)
+		}
+	}
+
+	return integer{signed: signed, size: intSize}, nil
+}
+
+func isIntOverflow(src string, dst string) bool {
```
ccoVeille, line 106: Maybe this

```suggestion
func couldIntOverflow(src string, dst string) bool {
```

testutils/g115_samples.go
```diff
@@ -0,0 +1,118 @@
+package testutils
+
+import "github.com/securego/gosec/v2"
+
+var SampleCodeG115 = []CodeSample{
```
ccoVeille, line 5: I'm a bit worried these tests are only focusing on things that could be reported.

I would expect to be added to cover conversion of things that should pass.

They are likely the ones to report false positive if something is invalid in the current implementation, or if a later refactoring screw everything

ccoVeille (changes requested):

FairlySadPanda:
This check has rolled out and ergo suddenly comments on it!

This check seems over-prescriptive. In the builtin functions, len returns type int, despite only returning zero or positive ints. Casting out of len into uint32 is thus more declarative. This is just one example where this change enforces a //nolint flag.

The act of casting is an assertion that the programmer knows what they are doing.

ccojocar:
> There is perfectly valid code that can be guaranteed not to overflow that this check catches. For example;

The type of v is still `int` which on a 64 bit system is 64 bits. Essentially the rule sees a conversion from int64 to int32. It is not so smart to detect the conversion done before using `strconv.ParseInt`.  It think adding these extension can be useful but for now you can make the size explicit by selecting the `int32` type for v instead of `int`. This will make the size explicit into the type.

See more discussions in #1185.

Also if this rule doesn't work for your use case, feel free to exclude it but in some cases integer  overflow can lead to security issues and there isn't a protection for this in the go runtime at the moment. 

```
gosec -exclude G115 ./...
```

moitias:
> @moitias The type of v is still `int` which on a 64 bit system is 64 bits.

Well, if we're being pedantic, no, the type of `v` is not `int` but `int64` because ParseInt and ParseUInt return explicit int64s/uint64s but yeah, this is just nitpicking.

> It think adding these extension can be useful but for now you can make the size explicit by selecting the `int32` type for v instead of `int`. This will make the size explicit into the type.

Without commit access to Go stdlib (and considering Go 1 compatibility promise), I don't think I can do that. 

> Essentially the rule sees a conversion from int64 to int32. It is not so smart to detect the conversion done before using `strconv.ParseInt`.

And that, indeed, is the problem. There _are_ safe ways of doing the conversion, but this checker does not allow for them either. 

> Also if this rule doesn't work for your use case, feel free to exclude it but in some cases integer overflow can lead to security issues and there isn't a protection for this in the go runtime at the moment.

Yeah, but this checker does not allow the developer to do it safely either, so they will need to disable the checker (locally or globally), making this checker essentially useless for any case that actually needs to do conversions and of course, nothing but lost CPU time for any case that doesn't do conversions. So, I'd argue that the usefulness of this check is dependent on it being 'smart'. If it is not, people will disable it and overflows might then creep in.

ccojocar:
@moitias I am looking forward to your contributions to improve this rule. I think is more productive to do this instead of just complaining and using an OSS tool maintain by volunteers ;-).

moitias:
Look, I appreciate the work you are putting into this (and helping me out) but I'm trying to point out - to help _you_ out in turn - that this rule in its current form seems to be a bit prematurely turned on because it seems to have no use cases at all. This is because if one has it enabled, there is no way of doing the conversion securely while in fact there obviously are. 

Could you please, to clarify the reasoning for this rule, at least describe one way of doing a conversion to a smaller bit size safely or are you of the opinion that it is never acceptable/secure to do? I would hope we could at least agree that linter warnings should have a way of fixing them, right?

ccoVeille:
> @moitias @FairlySadPanda @ccoVeille I am looking forward to your contributions to improve this rule. I think is more productive to do this instead of just complaining and using an OSS tool maintained by volunteers ;-). 

I can understand your reaction. I could say _Touché_. But I'm more likely to reply you positively than being affected. I'm a smiling person.

Be sure I have nothing against you or the idea you got. I wish the issues we are facing now where found before it was merge, but there were not.

You can understand people are reacting when their CI is now a Christmas tree 🎄

I wish I had reviewed your PR earlier. I wish others may have reviewed earlier.
Someone may have raised the consequences of this PR.

There problems was not with the idea or the code, but with the review of your changes suggestions.

That said, sometimes shit happens. Let's try to find a solution.

co60ca:
I came to see if this issue had be reported yet.

But we are getting
G115: integer overflow conversion uint8 -> int (gosec)
and according to the spec uint8 always fits in an int without overflow.

analyzers/conversion_overflow.go
ccoVeille: @ccojocar what do you think about the changes I suggested?

Do you agree with them?

Would you change it? Or do you want me to open a PR for them?

ccoVeille:
For anyone interested, the discussion continued here

https://github.com/securego/gosec/issues/1185

PR #1167, opened 2024-07-11
Fix multifile analyzer ignores

gartnera:
On `master`, `#nosec G115` will not work if your package has multiple files. This appears to be because the ignores populated in `CheckRules` are cleared after every file. These ignores are needed in the `CheckAnalyzers` call that happen after `CheckRules`.

Fix this by not clearing the ignores every loop. All tests cases appear to still pass:

```
ginkgo run
<snip>
Ran 144 of 144 Specs in 6.095 seconds
SUCCESS! -- 144 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -390,7 +390,6 @@ func (gosec *Analyzer) CheckRules(pkg *packages.Package) {
 		gosec.context.PkgFiles = pkg.Syntax
 		gosec.context.Imports = NewImportTracker()
 		gosec.context.PassedValues = make(map[string]interface{})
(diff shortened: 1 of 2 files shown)


ccojocar (approved):

ccojocar:
@gartnera Thanks for this fix. Please could have a look at the lint warnings. I think you need reformat the go115_samples.go file by running the command `gofumpt -w testutils/go115_samples.go`.

ccojocar (approved):

PR #1180, opened 2024-08-16
Allow excluding analyzers globally

### LINKED ISSUE #1175: CheckAnalyzers doesn't seem to respect exclude or nosec (e.g. G602)
I'm getting false positives with rule G602 and v2.20.0. I can't seem to find a way of supressing them. Adding to `-exclude` on the cmdline doesn't work, nor does adding `// #nosec G602` comments.

As an aside, the code is fine... it's like

```go
if len(x) == 1 {
  ...
} else if len(x) == 2 {
  if x[0].foo() == bar {
    return Pair{x[0], x[1]}
  } else if x[0].baz() == bar {
    return Pair{x[1], x[0]}
  }
} else {
  ...
}
```

and it generates errors for each of the 4 lines inside the `len(x) == 2` block. But the fact that it's not excludable is most worrying.

Downgrading to v2.17.0 is not an option as I now get a panic when it is built with Go 1.22 (upgrading from Go 1.20).

Rgvs:
* This change does not exclude analyzers for inline comment
* Changed the expected issues count for G103, G109. Previously G115 has been included in those issues count.
* Show analyzers IDs(G115, G602) in gosec usage help
* See #1175

fixes #1175

### DIFF
--- a/analyzers/analyzerslist.go
+++ b/analyzers/analyzerslist.go
@@ -1,0 +1,94 @@
+// (c) Copyright gosec's authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package analyzers
+
+import (
+	"golang.org/x/tools/go/analysis"
+)
+
+// AnalyzerDefinition contains the description of an analyzer and a mechanism to
+// create it.
+type AnalyzerDefinition struct {
+	ID          string
+	Description string
+	Create      AnalyzerBuilder
+}
+
+// AnalyzerBuilder is used to register an analyzer definition with the analyzer
+type AnalyzerBuilder func(id string, description string) *analysis.Analyzer
+
+// AnalyzerList contains a mapping of analyzer ID's to analyzer definitions and a mapping
+// of analyzer ID's to whether analyzers are suppressed.
+type AnalyzerList struct {
+	Analyzers          map[string]AnalyzerDefinition
+	AnalyzerSuppressed map[string]bool
+}
+
+// AnalyzersInfo returns all the create methods and the analyzer suppressed map for a
+// given list
+func (al *AnalyzerList) AnalyzersInfo() (map[string]AnalyzerDefinition, map[string]bool) {
+	builders := make(map[string]AnalyzerDefinition)
+	for _, def := range al.Analyzers {
+		builders[def.ID] = def
+	}
+	return builders, al.AnalyzerSuppressed
+}
+
+// AnalyzerFilter can be used to include or exclude an analyzer depending on the return
+// value of the function
+type AnalyzerFilter func(string) bool
+
+// NewAnalyzerFilter is a closure that will include/exclude the analyzer ID's based on
+// the supplied boolean value.
+func NewAnalyzerFilter(action bool, analyzerIDs ...string) AnalyzerFilter {
+	analyzerlist := make(map[string]bool)
+	for _, analyzer := range analyzerIDs {
+		analyzerlist[analyzer] = true
+	}
+	return func(analyzer string) bool {
+		if _, found := analyzerlist[analyzer]; found {
+			return action
+		}
+		return !action
+	}
+}
+
+var defaultAnalyzers = []AnalyzerDefinition{
+	{"G115", "Type conversion which leads to integer overflow", newConversionOverflowAnalyzer},
+	{"G602", "Possible slice bounds out of range", newSliceBoundsAnalyzer},
+}
+
+// Generate the list of analyzers to use
+func Generate(trackSuppressions bool, filters ...AnalyzerFilter) *AnalyzerList {
+	analyzerMap := make(map[string]AnalyzerDefinition)
+	analyzerSuppressedMap := make(map[string]bool)
+
+	for _, analyzer := range defaultAnalyzers {
+		analyzerSuppressedMap[analyzer.ID] = false
+		addToAnalyzerList := true
+		for _, filter := range filters {
+			if filter(analyzer.ID) {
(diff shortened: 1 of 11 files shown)


analyzers/analyzerslist.go
```diff
@@ -0,0 +1,89 @@
...
+
+// AnalyzerDefinition contains the description of an analyzer and a mechanism to
+// create it.
+type AnalyzerDefinition struct {
+	ID          string
+	Description string
+	Create      AnalyzerBuilder
+}
+
+type AnalyzerBuilder func(id string, description string) *analysis.Analyzer
```
ccojocar, line 27: nit: add a godoc comment to this since is a public type

analyzers/analyzerslist.go
```diff
@@ -0,0 +1,89 @@
...
+// AnalyzerList contains a mapping of analyzer ID's to analyzer definitions and a mapping
+// of analyzer ID's to whether analyzers are suppressed.
+type AnalyzerList struct {
+	Analyzers          map[string]AnalyzerDefinition
+	AnalyzerSuppressed map[string]bool
+}
+
+// AnalyzersInfo returns all the create methods and the analyzer suppressed map for a
+// given list
+func (rl AnalyzerList) AnalyzersInfo() (map[string]AnalyzerDefinition, map[string]bool) {
```
ccojocar, line 38: rename `rl` to `al` and use a pointer receiver

analyzers/analyzerslist.go
```diff
@@ -0,0 +1,89 @@
...
+		if _, found := analyzerlist[analyzer]; found {
+			return action
+		}
+		return !action
+	}
+}
+
+// Generate the list of analyzers to use
+func Generate(trackSuppressions bool, filters ...AnalyzerFilter) AnalyzerList {
+	analyzers := []AnalyzerDefinition{
```
ccojocar, line 67: Please move this into a global var at the package level, and I would name it `defaultAnalyzers`.

analyzers/analyzerslist.go
```diff
@@ -0,0 +1,89 @@
...
+	analyzerSuppressedMap := make(map[string]bool)
+
+ANALYZERS:
+	for _, analyzer := range analyzers {
+		analyzerSuppressedMap[analyzer.ID] = false
+		for _, filter := range filters {
+			if filter(analyzer.ID) {
+				analyzerSuppressedMap[analyzer.ID] = true
+				if !trackSuppressions {
+					continue ANALYZERS
```
ccojocar, line 82: Why do you need to go out of the loop here, and start again? Do you need to skip adding to the map if both are true `filter(analyzer.ID) && `trackSuppressions`?

rule.go
```diff
@@ -70,3 +71,25 @@ func (r RuleSet) RegisteredFor(n ast.Node) []Rule {
 func (r RuleSet) IsRuleSuppressed(ruleID string) bool {
 	return r.RuleSuppressedMap[ruleID]
 }
+
+// TODO: move to new file? If so what should be the name of the file?
+type AnalyzerSet struct {
```
ccojocar, line 76: Yeah, maybe move it into `analyzers_set.go` in the `analyzers` package.

testutils/g103_samples.go
```diff
@@ -27,7 +27,7 @@ func main() {
 	intPtr = (*int)(unsafe.Pointer(addressHolder))
 	fmt.Printf("\nintPtr=%p, *intPtr=%d.\n\n", intPtr, *intPtr)
 }
-`}, 3, gosec.NewConfig()},
+`}, 2, gosec.NewConfig()},
```
ccojocar, line 30: Why did these need to be changed? The logic of the analyzer/rule didn't change.

testutils/g109_samples.go
```diff
@@ -20,7 +20,7 @@ func main() {
 	value := int32(bigValue)
 	fmt.Println(value)
 }
-`}, 2, gosec.NewConfig()},
+`}, 1, gosec.NewConfig()},
```
ccojocar, line 23: ditto

testutils/g109_samples.go
```diff
@@ -38,7 +38,7 @@ func main() {
 		fmt.Println(bigValue)
 	}
 }
-`}, 2, gosec.NewConfig()},
+`}, 1, gosec.NewConfig()},
```
ccojocar, line 41: ditto

ccojocar (commented):
Thank you for this contribution! I left a few comments which it would be great if you could address them. Thanks again!

ccojocar:
I would also recommend to rebase this since there are some updates in the main. Thanks

Rgvs:
Thanks for quick feedback.
The analyzerslist.go code i wrote is technically a copy of rulelist.go. Some of your comments would deviate from rulelist.go. In view of that, would you still like me to go ahead and make those changes?

ccojocar:
> The analyzerslist.go code i wrote is technically a copy of rulelist.go. Some of your comments would deviate from rulelist.go. In view of that, would you still like me to go ahead and make those changes?

Please go ahead. It's fine they don't need to be identical. There are anyhow small things. Thanks for addressing my comments!

ccojocar:
Is this fixing #1175 issue?

Rgvs:
Yes, but partly.
Allows to exclude analyzers using `-exclude` option in cmd line.
This does not fix skipping code by comments like `// #nosec G602`

analyzers/analyzerslist.go
Rgvs: Done

analyzers/analyzerslist.go
Rgvs: Yes. Modified to improve legibility.

rule.go
Rgvs: Done

testutils/g103_samples.go
Rgvs: The 3 here represent two - G103 Rules and one - G115(analyzer - Type conversion which leads to integer overflow).
With the analyzers no longer being included. G115 is now excluded and hence the correct value is "2".

testutils/g109_samples.go
Rgvs: same reason as above.
1 issue of G109 and other of G115.

testutils/g109_samples.go
Rgvs: same reason as above.
1 issue of G109 and other of G115.

analyzers/analyzerslist.go
Rgvs: Done

ccojocar (approved):

PR #1189, opened 2024-08-23
Fix conversion overflow false positives

### LINKED ISSUE #1187: G115 ignores bounds checks
### Summary 

G115 reports issues even if we do proper bounds checks. This is similar in spirit to #1185, but would require the linter to be smarter.

```go
var x []string

if len(x) <= math.MaxUint32 {
    y := uint32(len(x))
    fmt.Println(y)
}
```

This reports `integer overflow conversion int -> uint32 (gosec)`.

### gosec version

I am running via golangci-lint v1.62.0.

n/a

n/a

### Expected behavior

The linter should see that there is a bounds check and thus be able to prove to itself that the overflow is impossible.

### Actual behavior

The linter does not consider anything about prior bounds checks, leading to false positives that need to be ignored, diminishing the utility of the check.

czechbol:
This pull request hopefully fixes the issue of false positives in the overflow analysis whenever bounds checks are performed or the size is already pre-determined.

**Disclaimer:**
This is my first time contributing to any linter and I don't consider myself being an AST wizard so please be more cautious when reviewing this code. 

fixes #1187

### DIFF
--- a/analyzers/conversion_overflow.go
+++ b/analyzers/conversion_overflow.go
@@ -16,6 +16,7 @@ package analyzers
 
 import (
 	"fmt"
+	"go/token"
 	"regexp"
 	"strconv"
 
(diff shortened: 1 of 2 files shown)


ccojocar (approved):
Thank you for this great contribution!

ccojocar:
@czechbol It seems that there some more use cases to handle https://github.com/securego/gosec/issues/1187#issuecomment-2311308184. I would be great if you could also check the bounds. Thanks

PR #1194, opened 2024-08-27
Improvement the int conversion overflow logic to handle bound checks

### LINKED ISSUE #1187: G115 ignores bounds checks

### LINKED ISSUE #1202: G115: false positive for uintptr -> unsafe.Pointer
### Summary 

When converting a uintptr to an unsafe.Pointer I get the error `G115: integer overflow conversion uintptr -> unsafe.Pointer`, even though these can be safely converted (see https://pkg.go.dev/unsafe#Pointer "A uintptr can be converted to a Pointer.").

Code sample: 
```go
package main

func main() {
	i := uintptr(123)
	p := unsafe.Pointer(i)
	_ = p
}
```

### gosec version

v2.20.1-0.20240822074752-ab3f6c1c83a0

go version go1.23.0 darwin/amd64

MacOS 14.6.1

### Expected behavior

Passes without errors.

### Actual behavior

Throws error `G115: integer overflow conversion uintptr -> unsafe.Pointer`

czechbol:
As per discussion in #1187, the implementation was lacking. This PR extends the bounds check logic to validate that the integer size is checked within the bounds of the destination type. 

I'd like to hear @rittneje's opinion on the changes as they reported the issues with it and @ben-krieger's too since he is already looking into it as well.

fixes #1187 
fixes #1202

### DIFF
--- a/analyzers/conversion_overflow.go
+++ b/analyzers/conversion_overflow.go
@@ -20,6 +20,7 @@ import (
 	"math"
 	"regexp"
 	"strconv"
+	"strings"
 
 	"golang.org/x/exp/constraints"
 	"golang.org/x/tools/go/analysis"
@@ -37,16 +38,20 @@ type integer struct {
 }
 
 type rangeResult struct {
-	MinValue     int
-	MaxValue     uint
-	IsRangeCheck bool
-	ConvertFound bool
+	MinValue             int
+	MaxValue             uint
+	ExplixitPositiveVals []uint
+	ExplicitNegativeVals []int
+	IsRangeCheck         bool
+	ConvertFound         bool
 }
 
-type branchBounds struct {
-	MinValue     *int
-	MaxValue     *uint
-	ConvertFound bool
+type branchResults struct {
+	MinValue             *int
+	MaxValue             *uint
+	ExplixitPositiveVals []uint
+	ExplicitNegativeVals []int
+	ConvertFound         bool
 }
 
 func newConversionOverflowAnalyzer(id string, description string) *analysis.Analyzer {
@@ -151,11 +156,6 @@ func parseIntType(intType string) (integer, error) {
 		min = -1 << (intSize - 1)
 
 	} else {
-		// Perform a bounds check
-		if intSize < 0 {
-			return integer{}, fmt.Errorf("invalid bit size: %d", intSize)
-		}
-
 		max = (1 << uint(intSize)) - 1
 		min = 0
 	}
@@ -255,6 +255,8 @@ func hasExplicitRangeCheck(instr *ssa.Convert, dstType string) bool {
 
 	minValue := srcInt.min
 	maxValue := srcInt.max
+	explicitPositiveVals := []uint{}
+	explicitNegativeVals := []int{}
 
 	if minValue > dstInt.min && maxValue < dstInt.max {
 		return true
@@ -269,6 +271,8 @@ func hasExplicitRangeCheck(instr *ssa.Convert, dstType string) bool {
 				if result.IsRangeCheck {
 					minValue = max(minValue, &result.MinValue)
 					maxValue = min(maxValue, &result.MaxValue)
+					explicitPositiveVals = append(explicitPositiveVals, result.ExplixitPositiveVals...)
+					explicitNegativeVals = append(explicitNegativeVals, result.ExplicitNegativeVals...)
 				}
 			case *ssa.Call:
 				// these function return an int of a guaranteed size
@@ -287,7 +291,9 @@ func hasExplicitRangeCheck(instr *ssa.Convert, dstType string) bool {
 				}
 			}
 
-			if minValue >= dstInt.min && maxValue <= dstInt.max {
+			if explicitValsInRange(explicitPositiveVals, explicitNegativeVals, dstInt) {
+				return true
+			} else if minValue >= dstInt.min && maxValue <= dstInt.max {
 				return true
 			}
 		}
@@ -350,6 +356,29 @@ func updateResultFromBinOp(result *rangeResult, binOp *ssa.BinOp, instr *ssa.Con
 		updateMinMaxForLessOrEqual(result, constVal, binOp.Op, operandsFlipped, successPathConvert)
 	case token.GEQ, token.GTR:
 		updateMinMaxForGreaterOrEqual(result, constVal, binOp.Op, operandsFlipped, successPathConvert)
+	case token.EQL:
+		if !successPathConvert {
+			break
+		}
+
+		// determine if the constant value is positive or negative
+		if strings.Contains(constVal.String(), "-") {
+			result.ExplicitNegativeVals = append(result.ExplicitNegativeVals, int(constVal.Int64()))
+		} else {
+			result.ExplixitPositiveVals = append(result.ExplixitPositiveVals, uint(constVal.Uint64()))
+		}
+
+	case token.NEQ:
+		if successPathConvert {
+			break
+		}
+
+		// determine if the constant value is positive or negative
+		if strings.Contains(constVal.String(), "-") {
+			result.ExplicitNegativeVals = append(result.ExplicitNegativeVals, int(constVal.Int64()))
+		} else {
+			result.ExplixitPositiveVals = append(result.ExplixitPositiveVals, uint(constVal.Uint64()))
+		}
 	}
 }
 
@@ -384,8 +413,8 @@ func updateMinMaxForGreaterOrEqual(result *rangeResult, constVal *ssa.Const, op 
 }
 
 // walkBranchForConvert walks the branch of the if statement to find the range of the variable and where the conversion is
-func walkBranchForConvert(block *ssa.BasicBlock, instr *ssa.Convert, visitedIfs map[*ssa.If]bool) branchBounds {
-	bounds := branchBounds{}
+func walkBranchForConvert(block *ssa.BasicBlock, instr *ssa.Convert, visitedIfs map[*ssa.If]bool) branchResults {
+	bounds := branchResults{}
 
 	for _, blockInstr := range block.Instrs {
 		switch v := blockInstr.(type) {
@@ -417,10 +446,33 @@ func walkBranchForConvert(block *ssa.BasicBlock, instr *ssa.Convert, visitedIfs 
 func isRangeCheck(v ssa.Value, x ssa.Value) bool {
 	switch op := v.(type) {
 	case *ssa.BinOp:
-		return (op.X == x || op.Y == x) &&
-			(op.Op == token.LSS || op.Op == token.LEQ || op.Op == token.GTR || op.Op == token.GEQ)
+		switch op.Op {
+		case token.LSS, token.LEQ, token.GTR, token.GEQ,
+			token.EQL, token.NEQ:
+			return op.X == x || op.Y == x
+		}
 	}
 	return false
+}
+
+func explicitValsInRange(explicitPosVals []uint, explicitNegVals []int, dstInt integer) bool {
+	if len(explicitPosVals) == 0 && len(explicitNegVals) == 0 {
+		return false
+	}
+
+	for _, val := range explicitPosVals {
+		if val > dstInt.max {
+			return false
+		}
+	}
+
+	for _, val := range explicitNegVals {
+		if val < dstInt.min {
+			return false
+		}
+	}
+
+	return true
 }
 
 func min[T constraints.Integer](a T, b *T) T {
--- a/analyzers/conversion_overflow.go
+++ b/analyzers/conversion_overflow.go
@@ -149,7 +149,6 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
 }
 
 func hasExplicitRangeCheck(instr *ssa.Convert, dstType string) bool {
-	block := instr.Block()
 	dstInt, err := parseIntType(dstType)
 	if err != nil {
 		return false
@@ -168,85 +167,96 @@ func hasExplicitRangeCheck(instr *ssa.Convert, dstType string) bool {
 		return true
 	}
 
-	// Recursive depth-first search of predecessor blocks of the SSA function to find bounds checks on the value being converted
-	var checkPredecessors func(block *ssa.BasicBlock, depth int) bool
-	checkPredecessors = func(block *ssa.BasicBlock, depth int) bool {
-		if depth > maxDepth {
-			return false
-		}
-
-		for _, pred := range block.Preds {
-			minChecked, maxChecked := checkBlockForRangeCheck(pred, instr, dstInt)
-
+	visitedIfs := make(map[*ssa.If]bool)
+	for _, block := range instr.Parent().Blocks {
+		var minChecked, maxChecked bool
+		for _, blockInstr := range block.Instrs {
+			switch v := blockInstr.(type) {
+			case *ssa.If:
+				minChecked, maxChecked = checkIfForRangeCheck(v, instr, dstInt, visitedIfs)
+			case *ssa.Call:
+				// len(slice) results in an int that is guaranteed >= 0, which
+				// satisfies the lower bound check for int -> uint conversion
+				if v != instr.X {
+					continue
+				}
+				if fn, isBuiltin := v.Call.Value.(*ssa.Builtin); isBuiltin && fn.Name() == "len" && !dstInt.signed {
+					minChecked = true
+				}
+			case *ssa.Convert:
+				if v == instr {
+					break
+				}
+			}
 			minBoundChecked = minBoundChecked || minChecked
 			maxBoundChecked = maxBoundChecked || maxChecked
 
 			if minBoundChecked && maxBoundChecked {
 				return true
 			}
-			if checkPredecessors(pred, depth+1) {
-				return true
-			}
 		}
-		return false
 	}
-
-	// Start checking from the initial block
-	checkPredecessors(block, 0)
-
-	if minBoundChecked && maxBoundChecked {
-		return true
-	}
-
 	return false
 }
 
-func checkBlockForRangeCheck(block *ssa.BasicBlock, instr *ssa.Convert, dstInt integer) (minBoundChecked, maxBoundChecked bool) {
-	for _, i := range block.Instrs {
-		switch v := i.(type) {
-		case *ssa.BinOp:
-			if isBoundCheck(v, instr.X) {
-				constVal, isOnLeft := constFromBoundCheck(v)
-				if constVal == nil {
-					continue
-				}
+func checkIfForRangeCheck(ifInstr *ssa.If, instr *ssa.Convert, dstInt integer, visitedIfs map[*ssa.If]bool) (minBoundChecked, maxBoundChecked bool) {
+	if visitedIfs[ifInstr] {
+		// If the if instruction has already been visited, we can skip it
+		return false, false
+	}
+	visitedIfs[ifInstr] = true
 
-				value := constVal.Int64()
+	// check Instrs for other bound checks
+	condBlock := ifInstr.Block()
+	if succIf, ok := condBlock.Succs[1].Instrs[1].(*ssa.If); ok {
+		// this is an OR condition and should be sufficient if it contains the other bound check
+		minBoundChecked, maxBoundChecked = checkIfForRangeCheck(succIf, instr, dstInt, visitedIfs)
+	}
 
-				if isOnLeft {
-					if v.Op == token.LSS || v.Op == token.LEQ {
-						maxBoundChecked = maxBoundChecked || checkMaxBoundValue(value, dstInt)
-					}
-					if v.Op == token.GTR || v.Op == token.GEQ {
-						minBoundChecked = minBoundChecked || checkMinBoundValue(value, dstInt)
-					}
-				} else {
-					if v.Op == token.LSS || v.Op == token.LEQ {
-						minBoundChecked = minBoundChecked || checkMinBoundValue(value, dstInt)
-					}
-					if v.Op == token.GTR || v.Op == token.GEQ {
-						maxBoundChecked = maxBoundChecked || checkMaxBoundValue(value, dstInt)
-					}
-				}
+	// check the instructions of the if block for other bound checks
+	for _, succ := range condBlock.Succs {
+		for _, blockInstr := range succ.Instrs {
+			if succIf, ok := blockInstr.(*ssa.If); ok {
+				// this is an AND condition and is insufficient to check the bounds but we need to visit it
+				// because walking the parent block will visit it again
+				_, _ = checkIfForRangeCheck(succIf, instr, dstInt, visitedIfs)
 			}
-		case *ssa.Call:
-			// len(slice) results in an int that is guaranteed >= 0, which
-			// satisfies the lower bound check for int -> uint conversion
-			if v != instr.X {
-				continue
-			}
-			if fn, isBuiltin := v.Call.Value.(*ssa.Builtin); isBuiltin && fn.Name() == "len" && !dstInt.signed {
-				minBoundChecked = true
-			}
-		case *ssa.Phi:
-			// Handle logical operations
-			continue
-		}
-
-		if minBoundChecked && maxBoundChecked {
-			break
 		}
 	}
+
+	cond := ifInstr.Cond
+	// Check if the condition is a bound check
+	if binOp, ok := cond.(*ssa.BinOp); ok {
+		if isBoundCheck(binOp, instr.X) {
+			constVal, isOnLeft := constFromBoundCheck(binOp)
+			if constVal == nil {
+				return false, false
+			}
+
+			value := constVal.Int64()
+
+			if isOnLeft {
+				if binOp.Op == token.LSS || binOp.Op == token.LEQ {
+					newMaxCheck := checkMaxBoundValue(value, dstInt)
+					maxBoundChecked = maxBoundChecked || newMaxCheck
+				}
+				if binOp.Op == token.GTR || binOp.Op == token.GEQ {
+					newMinCheck := checkMinBoundValue(value, dstInt)
+					minBoundChecked = minBoundChecked || newMinCheck
+				}
+			} else {
+				if binOp.Op == token.LSS || binOp.Op == token.LEQ {
+					newMinCheck := checkMinBoundValue(value, dstInt)
+					minBoundChecked = minBoundChecked || newMinCheck
+				}
+				if binOp.Op == token.GTR || binOp.Op == token.GEQ {
+					newMaxCheck := checkMaxBoundValue(value, dstInt)
+					maxBoundChecked = maxBoundChecked || newMaxCheck
+				}
+			}
+		}
+	}
+
 	return minBoundChecked, maxBoundChecked
 }
 
--- a/analyzers/conversion_overflow.go
+++ b/analyzers/conversion_overflow.go
@@ -197,28 +197,41 @@ func hasExplicitRangeCheck(instr *ssa.Convert, dstType string) bool {
 	return false
 }
 
-func checkBlockForRangeCheck(block *ssa.BasicBlock, instr *ssa.Convert, dstInt integer) (bool, bool) {
-	minBoundChecked := false
-	maxBoundChecked := false
-
+func checkBlockForRangeCheck(block *ssa.BasicBlock, instr *ssa.Convert, dstInt integer) (minBoundChecked, maxBoundChecked bool) {
 	for _, i := range block.Instrs {
-		if binOp, ok := i.(*ssa.BinOp); ok && isRelevantBinOp(binOp, instr.X) {
-			constVal := extractConst(binOp)
-			if constVal == nil {
-				continue
-			}
+		switch v := i.(type) {
+		case *ssa.BinOp:
+			if isBoundCheck(v, instr.X) {
+				constVal, isOnLeft := constFromBoundCheck(v)
+				if constVal == nil {
+					continue
+				}
 
-			value, err := strconv.ParseInt(constVal.Value.String(), 10, 64)
-			if err != nil {
-				continue
-			}
+				value := constVal.Int64()
 
-			minBoundChecked = minBoundChecked || checkMinBoundValue(value, dstInt)
-			maxBoundChecked = maxBoundChecked || checkMaxBoundValue(value, dstInt)
-
-			if minBoundChecked && maxBoundChecked {
-				break
+				if isOnLeft {
+					if v.Op == token.LSS || v.Op == token.LEQ {
+						maxBoundChecked = maxBoundChecked || checkMaxBoundValue(value, dstInt)
+					}
+					if v.Op == token.GTR || v.Op == token.GEQ {
+						minBoundChecked = minBoundChecked || checkMinBoundValue(value, dstInt)
+					}
+				} else {
+					if v.Op == token.LSS || v.Op == token.LEQ {
+						minBoundChecked = minBoundChecked || checkMinBoundValue(value, dstInt)
+					}
+					if v.Op == token.GTR || v.Op == token.GEQ {
+						maxBoundChecked = maxBoundChecked || checkMaxBoundValue(value, dstInt)
+					}
+				}
 			}
+		case *ssa.Phi:
+			// Handle logical operations
+			continue
+		}
+
+		if minBoundChecked && maxBoundChecked {
+			break
 		}
 	}
 	return minBoundChecked, maxBoundChecked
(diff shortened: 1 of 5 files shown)


analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,87 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
 
+func hasExplicitRangeCheck(instr *ssa.Convert, dstType string) bool {
+	block := instr.Block()
+	dstInt, err := parseIntType(dstType)
+	if err != nil {
+		return false
+	}
+
+	minBoundChecked := false
+	maxBoundChecked := false
```
ben-krieger, line 158: These values don't need to be preinitialized. I suggest just using:

```go
minBoundChecked := checkSourceMinBound(srcInt, dstInt)
maxBoundChecked := checkSourceMaxBound(srcInt, dstInt)
```

analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,87 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+	maxBoundChecked = checkSourceMaxBound(srcInt, dstInt)
+
+	// If both bounds are already checked, return true
+	if minBoundChecked && maxBoundChecked {
+		return true
+	}
+
+	// Recursive function to check predecessors
+	var checkPredecessors func(block *ssa.BasicBlock) bool
+	checkPredecessors = func(block *ssa.BasicBlock) bool {
```
ben-krieger, line 175:
```suggestion
	checkPredecessors := func(block *ssa.BasicBlock) bool {
```

analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,87 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+
+	minBoundChecked = checkSourceMinBound(srcInt, dstInt)
+	maxBoundChecked = checkSourceMaxBound(srcInt, dstInt)
+
+	// If both bounds are already checked, return true
+	if minBoundChecked && maxBoundChecked {
+		return true
+	}
+
+	// Recursive function to check predecessors
```
ben-krieger, line 173:
```suggestion
	// Recursive depth-first search of predecessor blocks of the SSA function to find bounds checks on the value being converted 
```

analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,87 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+	var checkPredecessors func(block *ssa.BasicBlock) bool
+	checkPredecessors = func(block *ssa.BasicBlock) bool {
+		for _, pred := range block.Preds {
+			minChecked, maxChecked := checkBlockForRangeCheck(pred, instr, dstInt)
+			if minChecked {
+				minBoundChecked = true
+			}
+			if maxChecked {
+				maxBoundChecked = true
+			}
```
ben-krieger, line 183:
```suggestion
			minBoundChecked = minBoundChecked || minChecked
			maxBoundChecked = maxBoundChecked || maxChecked
```

analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,87 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+	if minBoundChecked && maxBoundChecked {
+		return true
+	}
+
+	return false
+}
+
+func checkBlockForRangeCheck(block *ssa.BasicBlock, instr *ssa.Convert, dstInt integer) (bool, bool) {
+	minBoundChecked := false
+	maxBoundChecked := false
```
ben-krieger, line 206:
```suggestion
func checkBlockForRangeCheck(block *ssa.BasicBlock, instr *ssa.Convert, dstInt integer) (minBoundChecked, maxBoundChecked bool) {
```

analyzers/conversion_overflow.go
```diff
@@ -224,3 +291,71 @@ func isIntOverflow(src string, dst string) bool {
 
 	return false
 }
+
+func isRelevantBinOp(binOp *ssa.BinOp, x ssa.Value) bool {
```
ben-krieger, line 295: Maybe rename this something like `isBoundCheck` to match other naming?

analyzers/conversion_overflow.go
```diff
@@ -224,3 +291,71 @@ func isIntOverflow(src string, dst string) bool {
...
+	return (binOp.X == x || binOp.Y == x) &&
+		(binOp.Op == token.LSS || binOp.Op == token.LEQ || binOp.Op == token.GTR || binOp.Op == token.GEQ)
+}
+
+func extractConst(binOp *ssa.BinOp) *ssa.Const {
```
ben-krieger, line 300: Maybe name this something like `constFromBoundCheck` to match other naming?

ben-krieger (commented):
I left some comments on how I might make the code a bit more clear. Feel free to completely ignore.

I did test this out on my codebase locally and it removed the same false negatives as my PR #1193. Therefore, I will be closing my PR in favor of yours.

analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,87 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+			if constVal == nil {
+				continue
+			}
+
+			value, err := strconv.ParseInt(constVal.Value.String(), 10, 64)
+			if err != nil {
+				continue
+			}
+
+			minBoundChecked = minBoundChecked || checkMinBoundValue(value, dstInt)
```
rittneje, line 220: Doesn't min vs. max need to be determined based on (1) the op, and (2) which side of the op `x` appears on?

analyzers/conversion_overflow.go
```diff
@@ -224,3 +291,71 @@ func isIntOverflow(src string, dst string) bool {
...
+		var b uint = (1 << dstInt.size) - 1
+		return a <= b
+	}
+	// Both source and destination are unsigned
+	return (1<<srcInt.size)-1 <= (1<<dstInt.size)-1
+}
+
+func checkMinBoundValue(value int64, dstInt integer) bool {
+	if dstInt.signed {
+		return value == -(1 << (dstInt.size - 1))
```
rittneje, line 351: Shouldn't this be `value >= -(1 << (dstInt.size - 1))`? For example, it should be permitted to do `if x > 1` instead of `if x > 0` when converting `int` to `uint`.

Likewise for all the other checks.

rittneje:
@czechbol Unfortunately, I've never worked with the Go parser/AST before, so I won't be much help.

analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,87 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+
+	return false
+}
+
+func checkBlockForRangeCheck(block *ssa.BasicBlock, instr *ssa.Convert, dstInt integer) (bool, bool) {
+	minBoundChecked := false
+	maxBoundChecked := false
+
+	for _, i := range block.Instrs {
+		if binOp, ok := i.(*ssa.BinOp); ok && isRelevantBinOp(binOp, instr.X) {
```
rittneje, line 209: Need to consider `&&`, `||`, and `!` here. For example, `x >= 0 && x <= math.MaxUint32` is very different from `x >= 0 || x <= math.MaxUint32`.

analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,87 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+				return true
+			}
+		}
+		return false
+	}
+
+	// Start checking from the initial block
+	checkPredecessors(block)
+
+	if minBoundChecked && maxBoundChecked {
```
ben-krieger, line 197: If src and dst types are both unsigned, then we don't care about `minBoundChecked`.

rittneje: I think more generally, if src is unsigned, then the min bound check is implicitly satisfied, no matter what dst is.

analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,87 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+			if minChecked {
+				minBoundChecked = true
+			}
+			if maxChecked {
+				maxBoundChecked = true
+			}
+			if minBoundChecked && maxBoundChecked {
+				return true
+			}
+			if checkPredecessors(pred) {
```
ccojocar, line 187: I would keep track of the depth and stop after a certain value to avoid an infinite loop or stack overflows (e.g. 10 recursions).

analyzers/conversion_overflow.go
deerbone: I can't do that since I call checkPredecessors inside itself so it needs to be predeclared in this case. Otherwise I get an undefined error

analyzers/conversion_overflow.go
deerbone: resolved

analyzers/conversion_overflow.go
deerbone: resolved

analyzers/conversion_overflow.go
deerbone: resolved

analyzers/conversion_overflow.go
deerbone: resolved

analyzers/conversion_overflow.go
deerbone: resolved

analyzers/conversion_overflow.go
deerbone: resolved

analyzers/conversion_overflow.go
deerbone: should be resolved

analyzers/conversion_overflow.go
deerbone: this check is performed by `checkSourceMinBound` which I tweaked a bit based on @rittneje's comment

analyzers/conversion_overflow.go
deerbone: should be resolved

analyzers/conversion_overflow.go
deerbone: This one I'll need to spend some more time on

czechbol:
I apologise for the email spam. Got my github accounts mixed up.

Everything should now be accounted for apart from this comment -  https://github.com/securego/gosec/pull/1194#discussion_r1733645751.

I'm still thinking through how to tackle it.

ccojocar:
> Everything should now be accounted for apart from this comment - https://github.com/securego/gosec/pull/1194#discussion_r1733645751.

> I'm still thinking through how to tackle it.

@czechbol Thanks for addressing the review comments. Please just let me know when this is ready for final review.

ccojocar:
I think this requires a rebase, since there are some conflicts in `testutils/g115_samples.go` which need to be resolved.

ben-krieger:
Unfortunately with the latest changes, I've seen two regressions running gosec on my private codebase.

First, is a case of:

```go
	xbLen := len(xb)
	if xbLen > math.MaxUint16 {
		return nil, fmt.Errorf("x contains too many bytes")
	}

	b = binary.BigEndian.AppendUint16(b, uint16(xbLen))
```

Second is a case of:

```go
	numEntries := len(ov.Entries)
	if numEntries > math.MaxUint8 {
		return nil, fmt.Errorf(...)
	}

	_ = uint8(numEntries)

```

I'm guessing this is because int -> uintX doesn't guarantee that the value is > 0. However, `len` semantically does... This may be a little hard to check for, but I think it's worthwhile.

czechbol:
> I think this requires a rebase, since there are some conflicts in testutils/g115_samples.go which need to be resolved.

sure, that is not an issue

> I'm guessing this is because int -> uintX doesn't guarantee that the value is > 0. However, len semantically does... This may be a little hard to check for, but I think it's worthwhile.

I'll look into it too after I figure out the conditional stuff.

analyzers/conversion_overflow.go
```diff
@@ -161,6 +147,99 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+				} else {
+					if v.Op == token.LSS || v.Op == token.LEQ {
+						minBoundChecked = minBoundChecked || checkMinBoundValue(value, dstInt)
+					}
+					if v.Op == token.GTR || v.Op == token.GEQ {
+						maxBoundChecked = maxBoundChecked || checkMaxBoundValue(value, dstInt)
+					}
+				}
+			}
+		case *ssa.Phi:
```
ben-krieger, line 231: Here's a way to find if a signed value is created from a call to the builtin `len` and therefore satisfies the minimum bound for uints.

```suggestion
		case *ssa.Call:
			// len(slice) results in an int that is guaranteed >= 0, which
			// satisfies the lower bound check for int -> uint conversion
			if v != instr.X {
				continue
			}
			if fn, isBuiltin := v.Call.Value.(*ssa.Builtin); isBuiltin && fn.Name() == "len" && !dstInt.signed {
				minBoundChecked = true
			}
		case *ssa.Phi:
```

czechbol:
This code is definitely not final. It will most likely need some refactoring after the following cases are handled as well.

I currently have some way to handle AND and OR conditions in the if statements. Since the SSA does not give me a nice way to check the entire if statement and breaks it up into component instructions I need to look through them to determine their relationship. If you have suggestions how to do it differently, please let me know.

**Currently unhandled cases:**
- Convert instruction inside the if block. Even if I'm not a fan of this code style, the false positives should be resolved.
- NOT operation within the conditions. 

I will hopefully be able to solve these in the coming days.

analyzers/conversion_overflow.go
```diff
@@ -162,6 +148,118 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+			switch v := blockInstr.(type) {
+			case *ssa.If:
+				minChecked, maxChecked = checkIfForRangeCheck(v, instr, dstInt, visitedIfs)
+			case *ssa.Call:
+				// len(slice) results in an int that is guaranteed >= 0, which
+				// satisfies the lower bound check for int -> uint conversion
+				if v != instr.X {
+					continue
+				}
+				if fn, isBuiltin := v.Call.Value.(*ssa.Builtin); isBuiltin && fn.Name() == "len" && !dstInt.signed {
```
rittneje, line 183: Why is `dst.signed` relevant? Even if it's signed, all signed integers have a minimum bound < 0, so I think `minChecked` can be set to true no matter what. For example, the following is valid:
```go
if len(x) < math.MaxInt16 {
    return int16(len(x))
}
```

rittneje, line 183: Can we also account for `cap`? It's not as widely used as `len`, but should be easy to add.

rittneje:
@czechbol I am currently getting a panic on the following input:

```go
package main

import "math"

func foo(x []string) int16 {
	if len(x) < math.MaxInt16 {
		return int16(len(x))
	}
	return 0
}

func main() {

}
```

> panic: runtime error: index out of range [1] with length 1
>
> goroutine 1 [running]:
github.com/securego/gosec/v2/analyzers.checkIfForRangeCheck(0xc0004aa048, 0xc0008107d0, {0xf8?, 0x7f61505e3178?}, 0xc000cab570)
        /home/rittneje/gosec/analyzers/conversion_overflow.go:211 +0x568
github.com/securego/gosec/v2/analyzers.hasExplicitRangeCheck(0xc0008107d0, {0xc000289cc0?, 0x5?})
        /home/rittneje/gosec/analyzers/conversion_overflow.go:176 +0x2b9
github.com/securego/gosec/v2/analyzers.isSafeConversion(0xc0008107d0)
        /home/rittneje/gosec/analyzers/conversion_overflow.go:93 +0xaa
github.com/securego/gosec/v2/analyzers.runConversionOverflow(0xc000190a80)
        /home/rittneje/gosec/analyzers/conversion_overflow.go:54 +0x27e
github.com/securego/gosec/v2.(*Analyzer).CheckAnalyzers(0xc0007c6080, 0xc00032cd80)
        /home/rittneje/gosec/analyzer.go:446 +0x4a2
github.com/securego/gosec/v2.(*Analyzer).Process(0xc0007c6080, {0x0, 0x0, 0x0}, {0xc0007d6c50, 0x1, 0x3d?})
        /home/rittneje/gosec/analyzer.go:317 +0x488
main.main()
        /home/rittneje/gosec/cmd/gosec/main.go:476 +0xde5

rittneje:
@czechbol I haven't thought through it much, but I'm wondering if it would be better to keep track of the known minimum and maximum bounds themselves, instead of just flags. That way we have a good foundation for adding support for more interesting cases in the future. As a rough idea:

```go
func foo(x int64) int16 {
    // Initially our "context" is that x is in the range [MinInt64, MaxInt64].

    if x > math.MaxInt16 {
        return math.MaxInt16
    }

    // Now because of the return, we can negate the if check and "apply" it to the context.
    // Thus x is in the range [MinInt64, MaxInt16].

    if x < math.MinInt16 {
        return math.MinInt16
    }

    // Again because of the return, we can negate the if check and "apply" it to the context.
    // Thus x is in the range [MinInt16, MaxInt16].

    // We know that x must be in range, so this conversion is safe.
    return int16(x)
}
```

In the future, I think this bookkeeping could help with situations involving arithmetic:
```go
func foo(x uint64) uint16 {
   if x + 1 > math.MaxUint16 {
       return math.MaxUint16
   }
    return uint16(x + 1)
}
```

ben-krieger:
> @czechbol I am currently getting a panic on the following input:
> 
> ```go
> package main
> 
> import "math"
> 
> func foo(x []string) int16 {
> 	if len(x) < math.MaxInt16 {
> 		return int16(len(x))
> 	}
> 	return 0
> }
> 
> func main() {
> 
> }
> ```
> 
> > panic: runtime error: index out of range [1] with length 1
> > goroutine 1 [running]:
> > github.com/securego/gosec/v2/analyzers.checkIfForRangeCheck(0xc0004aa048, 0xc0008107d0, {0xf8?, 0x7f61505e3178?}, 0xc000cab570)
> > /home/rittneje/gosec/analyzers/conversion_overflow.go:211 +0x568
> > github.com/securego/gosec/v2/analyzers.hasExplicitRangeCheck(0xc0008107d0, {0xc000289cc0?, 0x5?})
> > /home/rittneje/gosec/analyzers/conversion_overflow.go:176 +0x2b9
> > github.com/securego/gosec/v2/analyzers.isSafeConversion(0xc0008107d0)
> > /home/rittneje/gosec/analyzers/conversion_overflow.go:93 +0xaa
> > github.com/securego/gosec/v2/analyzers.runConversionOverflow(0xc000190a80)
> > /home/rittneje/gosec/analyzers/conversion_overflow.go:54 +0x27e
> > github.com/securego/gosec/v2.(*Analyzer).CheckAnalyzers(0xc0007c6080, 0xc00032cd80)
> > /home/rittneje/gosec/analyzer.go:446 +0x4a2
> > github.com/securego/gosec/v2.(*Analyzer).Process(0xc0007c6080, {0x0, 0x0, 0x0}, {0xc0007d6c50, 0x1, 0x3d?})
> > /home/rittneje/gosec/analyzer.go:317 +0x488
> > main.main()
> > /home/rittneje/gosec/cmd/gosec/main.go:476 +0xde5

Putting the panic aside, we actually won't be able to allow this conversion. SSA IR will have each `len(x)` as a separate value. Therefore, the latter value which is converted won't be considered upper-bound-checked, since the bound check was actually on a different value.

I imagine that this case will frustrate a lot of people with gosec, but there's nothing we can do. Realizing that both len calls will have the same value is a compiler optimization at best, since I'm pretty sure you can mutate a slice's length using reflection.

analyzers/conversion_overflow.go
ben-krieger: That's embarrassingly obvious. Hooray open source code reviewer volunteers?

czechbol:
> I am currently getting a panic on the following input

@rittneje Yes, I know there's panics, that's why I said I "have some way" of handling conditions, that Converts inside if blocks don't work and that it's not ready to be merged yet. The logic still has work to be done, but I'm slowly covering more test cases and want to share the changes with you. If  you find any other curious test cases, feel free to suggest them or open a PR to my branch. 

I'll be honest, I definitely didn't fully understand what I was getting myself into when I opened #1189 but I'm in too deep now and I still want to fix all the false positives I can.

If what @ben-krieger says is true then it's a shame but what can we do. 
Perhaps we could push this PR into a good enough middle ground solution and iterate from there

In the meantime I'm switching this PR to a Draft.

ben-krieger:
> > I am currently getting a panic on the following input
> 
> @rittneje Yes, I know there's panics, that's why I said I "have some way" of handling conditions, that Converts inside if blocks don't work and that it's not ready to be merged yet. There still is work to do, but I'm slowly covering more test cases and want to share the changes with you. If you find any other curious test cases, feel free to suggest them or open a PR to my branch.
> 
> I'll be honest, I definitely didn't fully understand what I was getting myself into when I opened #1189 but I'm in too deep now and I still want to fix all the false positives I can.
> 
> If what @ben-krieger says is true then it's a shame but what can we do. Perhaps we could push this PR into a good enough middle ground solution and iterate from there.
> 
> In the meantime I'm switching this PR to a Draft.

@czechbol I also feel like I jumped into something out of my depth. I was absolutely [nerd-sniped](https://xkcd.com/356/).

But you've done a great job! In fact, I'm quite impressed with what 3 unprepared devs could do to significantly tackle these false positives.

In my opinion, the main thing holding us back is not even our experience with SSA or language analysis, but the fact that this probably really requires complex inspection of the abstract syntax tree to understand things like "does a failed bounds check result in conversion code being unreachable" and "does a possibly complex boolean expression check either bound, none, or both?" This is no doubt difficult even for developers of compilers!

So I think we should just take a few more good looks at what we can do here, document the most common false positives (and the couple false negatives we will introduce), and ship it. 

I know this PR has merit, because it saves me from 22 obviously false positives in just one library I maintain. This is good work!

ccojocar:
> I currently have some way to handle AND and OR conditions in the if statements. Since the SSA does not give me a nice way to check the entire if statement and breaks it up into component instructions, I need to look through them to determine their relationship. If you have suggestions how to do it differently, please let me know.

@czechbol I think from `If` instruction you can get the `Block` to which it belongs to, and within that Block you have all outer scope instructions, which you can reference from the `Cond` instruction. Also you can get all function instructions block via the `Parent()`. It is more complicated if you want to track back through multiple function calls since the current SSA implementation doesn't provide the graph outside of the package.

I would focus to cover the most common uses cases at the function level. Validation outside of the function block can be addressed later.

czechbol:
Now this is an interesting corner I've gotten myself into. 

I've involuntarily introduced a situation where my own logic looks kind of stupid.

analyzers/conversion_overflow.go
```diff
@@ -140,9 +140,22 @@ func parseIntType(intType string) (integer, error) {
 	var max uint
 
 	if signed {
-		max = (1 << uint(intSize-1)) - 1
+		shiftAmount := intSize - 1
+
+		// Perform a bounds check
+		if shiftAmount < 0 {
+			return integer{}, fmt.Errorf("invalid shift amount: %d", shiftAmount)
+		}
```
czechbol, line 153: I've added basic support for EQL and NEQ. 

There are still annoyances like this one where I cannot subtract from a value that I have already validated without a redundant check, even though the conversion is completely safe in this case.

ccojocar:
> If someone wants to explicitly check a value with an equals operator, I currently have no way of handling that.
What do you think? Should I consider this out of scope of this PR?

It sounds good to me to handle it later since the current implementation in converting quite a good chunk of use cases. A good solution is better than a perfect solution which is not implemented.

ccojocar (commented):
LGTM, thanks you very much for this great work!

Just let me know when you feel ready to merge it. Thanks

czechbol:
Thanks! I'm marking this as ready for review then. Let's wait a bit for @ben-krieger and @rittneje for their opinions and let's merge it.

ben-krieger:
> Thanks! I'm marking this as ready for review then. Let's wait a bit for @ben-krieger and @rittneje for their opinions and let's merge it.

I'm on a short trip this weekend, but I'll review it as soon as I can.

ccojocar:
@czechbol Can we also cover this use case #1202? Thanks

czechbol:
@ccojocar I cannot reproduce that error with my code. I added a test case for it and it passes.

ccojocar:
> I cannot reproduce that error with my code. I added a test case for it and it passes.

it sounds great!

analyzers/conversion_overflow.go
```diff
@@ -162,66 +242,259 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+		result.MaxValue = uint(constVal.Uint64())
+		if op == token.LSS {
+			result.MaxValue--
+		}
 	}
+}
 
-	dstInt, err := parseIntType(dst)
-	if err != nil {
+// walkBranchForConvert walks the branch of the if statement to find the range of the variable and where the conversion is
```
ccojocar, line 415: ditto

ccojocar (approved):

czechbol:
@ccojocar I'm seeing weird warnings all over the repo, even if I rebase on your latest changes. Did you update anything related to typecheck?

Or perhaps was it typecheck that released a new version?

ccojocar:
> Or perhaps was it typecheck that released a new version?

I updated the dependencies but the build is green on master https://github.com/securego/gosec/actions/runs/10662974983. It might be something in your changes.

ccojocar:
Could you try to rebase, and trigger the build? It looks like a GitHub weirdness. That file wasn't touch.

ccojocar:
I rebased and fix the go module. It should be fine now.

czechbol:
Awesome, thanks!

czechbol:
Hey @ccojocar, I think we waited long enough. I'd say if there are any issues let's address them later. 

I'd go ahead and merge it as is.

ben-krieger:
Sorry for being slow on the review. I'm now getting 5 new hits on my codebase since I ran this PR a few days ago.

Based on a quick glance, these actually look like completely code locations to trigger G115. Since this PR is related to relaxing rules to reduce false positives, perhaps another change from the main branch caused it.

Don't let it stop you from merging this PR. I may have a follow-up PR.

ben-krieger:
> Sorry for being slow on the review. I'm now getting 5 new hits on my codebase since I ran this PR a few days ago.
> 
> Based on a quick glance, these actually look like completely code locations to trigger G115. Since this PR is related to relaxing rules to reduce false positives, perhaps another change from the main branch caused it.
> 
> Don't let it stop you from merging this PR. I may have a follow-up PR.

Followup: All my new overflow analysis hits were int->uint64 or int64->uint64. There must have been a fix for these previously missed overflows. The code was easy to refactor to avoid overflowing - proof that the errors were actually useful.

However, one conversion overflow could not be resolved. I distilled it to this failing test:

```diff
diff --git a/testutils/g115_samples.go b/testutils/g115_samples.go
index 84db8e0..483647f 100644
--- a/testutils/g115_samples.go
+++ b/testutils/g115_samples.go
@@ -694,6 +694,25 @@ func main() {
 	i := uintptr(123)
 	p := unsafe.Pointer(i)
 	_ = p
+}
+	`,
+	}, 0, gosec.NewConfig()},
+	{[]string{
+		`
+package main
+
+import (
+	"fmt"
+	"math/rand"
+)
+
+func main() {
+	a := rand.Int63()
+	if a >= 0 {
+		panic("no positivity allowed")
+	}
+	b := uint64(-a)
+	fmt.Printf("%d\n", b)
 }
 	`,
 	}, 0, gosec.NewConfig()},

```

testutils/g115_samples.go
```diff
@@ -390,4 +434,267 @@ func main() {
...
+
+import "unsafe"
+
+func main() {
+	i := uintptr(123)
+	p := unsafe.Pointer(i)
+	_ = p
+}
+	`,
+	}, 0, gosec.NewConfig()},
```
ben-krieger, line 699: Here's a currently failing test. It is safe to convert a negative int64 to a uint64.

```suggestion
	}, 0, gosec.NewConfig()},
	{[]string{
		`
package main

import (
	"fmt"
	"math/rand"
)

func main() {
	a := rand.Int63()
	if a >= 0 {
		panic("no positivity allowed")
	}
	b := uint64(-a)
	fmt.Printf("%d\n", b)
}
	`,
	}, 0, gosec.NewConfig()},
```

ben-krieger: Here's the generated SSA IR from the test:

```
func main():
0:                                                                entry P:0 S:2
	t0 = math/rand.Int63()                                            int64
	t1 = t0 >= 0:int64                                                 bool
	if t1 goto 1 else 2
1:                                                              if.then P:1 S:0
	t2 = make interface{} <- string ("no positivity all...":string) interface{}
	panic t2
2:                                                              if.done P:1 S:0
	t3 = -t0                                                          int64
	t4 = convert uint64 <- int64 (t3)                                uint64
	t5 = new [1]any (varargs)                                       *[1]any
	t6 = &t5[0:int]                                                    *any
	t7 = make any <- uint64 (t4)                                        any
	*t6 = t7
	t8 = slice t5[:]                                                  []any
	t9 = fmt.Printf("%d\n":string, t8...)                (n int, err error)
	return
```

So to handle this case, we'd need to (as a special case?) see if the value being converted is the inverse of another value. If it is, then we can take the bounds of said value and flip them, then combine them with the existing bounds with the existing min-max logic.

I'd like to make the case that an inversion is a special case worth supporting, even if not all arithmetic operations are.

analyzers/conversion_overflow.go
```diff
@@ -162,66 +242,263 @@ func isStringToIntConversion(instr *ssa.Convert, dstType string) bool {
...
+				if fn, isBuiltin := v.Call.Value.(*ssa.Builtin); isBuiltin {
+					switch fn.Name() {
+					case "len", "cap":
+						minValue = 0
+					}
+				}
+			case *ssa.Convert:
+				if v == instr {
+					break
+				}
```
ben-krieger, line 291: This case does nothing and can be omitted. The `break` statement exits the switch, which will be exited regardless.

ccojocar:
> Followup: All my new overflow analysis hits were int->uint64 or int64->uint64. There must have been a fix for these previously missed overflows. The code was easy to refactor to avoid overflowing - proof that the errors were actually useful.

I think we should fix this before merging it. It would be helpful to add the test case to the samples. Thanks

testutils/g115_samples.go
czechbol: @ben-krieger, I fixed this issue. I specifically included just the negation case of an UnOp. BinOps, such as subtraction, multiplication, etc. would probably require more complex inspection so I decided to ignore them for now.

czechbol:
@ben-krieger could you please elaborate on the 5 new cases? From your descriptions it seems to me that, apart from the negation that I already addressed, those cases are completely legitimate and should be behaving that way. But it seems that @ccojocar has a different idea so I'd like to avoid misunderstanding.

During my work on this PR, I ended up extending the `parseIntType` function to return the min and max values too and wrote unit tests for it. I later used the min and max values simplify the `isIntOverflow` function. It is possible that I fixed a previously undiscovered bug, or introduced a new one, depending on how your comment is interpreted.

ben-krieger:
> @ben-krieger could you please elaborate on the 5 new cases? From your descriptions it seems to me that, apart from the negation that I already addressed, those cases are completely legitimate and should be behaving that way. But it seems that @ccojocar has a different idea so I'd like to avoid misunderstanding.
> 
> During my work on this PR, I ended up extending the `parseIntType` function to return the min and max values too and wrote unit tests for it. I later used the min and max values simplify the `isIntOverflow` function. It is possible that I fixed a previously undiscovered bug, or introduced a new one, depending on how your comment is interpreted.

Sorry for the ambiguity. Yes, 4 were new legitimate cases and the fifth was the negation one that I suggested a test for.

czechbol:
@ccojocar golangci-lint keeps complaining about a potential overflow but if I build gosec locally it doesn't flag anything. I'd disregard that action for now as they're still using the old version of gosec.

ccojocar:
> golangci-lint keeps complaining about a potential overflow but if I build gosec locally it doesn't flag anything. I'd disregard that action for now as they're still using the old version of gosec.

I ignored the warning for now.

ben-krieger (approved):

stephenc:
Does this also allow us to signal intentional truncation without a bounds check, e.g. `uint16(x &0xffff)`

ben-krieger:
> Does this also allow us to signal intentional truncation without a bounds check, e.g. `uint16(x &0xffff)`

No. Someone would need to take up a significant effort in order to support bounds checks that involve 1) arithmetic binary operations, 2) boolean logic operations, or 3) the min and max builtins from Go 1.21.

Perhaps we can create a ticket to track this? But I'm not sure who will sign up for that level of effort, so maybe it should be implemented across multiple PRs.

czechbol:
@stephenc I'm sorry I claimed that it has been resolved I only noticed the sentence about truncation after you commented here.

That being said I currently don't have the time capacity to continue working on this feature, I've already sank too much time into it. I hope you understand that.

stephenc:
@czechbol fixing the ranges is great. As you closed my issue perhaps you could just create a placeholder about adding support for bitmasks to signal intentional truncation so that it does not get lost. No rush to implement, I can live with `// # nosec G115` for now

stephenc:
> > Does this also allow us to signal intentional truncation without a bounds check, e.g. `uint16(x &0xffff)`
> 
> No. Someone would need to take up a significant effort in order to support bounds checks that involve 1) arithmetic binary operations, 2) boolean logic operations, or 3) the min and max builtins from Go 1.21.
> 
> Perhaps we can create a ticket to track this? But I'm not sure who will sign up for that level of effort, so maybe it should be implemented across multiple PRs.

I see those as all different cases. I think intentional truncation is always going to be of a fixed pattern... a mask that fits into the target type and a bitwise and... and for it to be intentional I would pair that with both being inside the cast... such a cast, for the all 1's case, should be optimized as simple truncation by the compiler and thus it is clearly signalling intent to truncate. as things currently stand I don't see how I can signal intentional truncation other than by disabling the rule on the line which could mask an unintended truncation

ben-krieger:
> > > Does this also allow us to signal intentional truncation without a bounds check, e.g. `uint16(x &0xffff)`
> > 
> > 
> > No. Someone would need to take up a significant effort in order to support bounds checks that involve 1) arithmetic binary operations, 2) boolean logic operations, or 3) the min and max builtins from Go 1.21.
> > Perhaps we can create a ticket to track this? But I'm not sure who will sign up for that level of effort, so maybe it should be implemented across multiple PRs.
> 
> I see those as all different cases. I think intentional truncation is always going to be of a fixed pattern... a mask that fits into the target type and a bitwise and... and for it to be intentional I would pair that with both being inside the cast... such a cast, for the all 1's case, should be optimized as simple truncation by the compiler and thus it is clearly signalling intent to truncate. as things currently stand I don't see how I can signal intentional truncation other than by disabling the rule on the line which could mask an unintended truncation

The (ugly) way to get gosec to accept the conversion would be something like:

```go
x = x&0xffff
if x < 0 || x > math.MaxUint16 {
    panic("unreachable")
}
_ = uint16(x)
```

czechbol:
@ben-krieger, @stephenc, @rittneje, @ccojocar FYI I opened #1212 to compile all the false positives we can find in a single issue.

PR #1197, opened 2024-08-29
Add new rule G407 to detect hard-coded nonce and initialization vectors in crypto algorithms

### LINKED ISSUE #1196: Gosec does not detect hard-coded nonces/initialization vectors for multiple encryption algorithms
### Summary 
Gosec does not detect hard-coded nonces/initialization vectors for multiple encryption algorithms such as AES GCM,CBC,CFB,CTR,OFB,ASCON,CHACHA20
none, rule for that doesn't exists

### Expected behavior

Detect when a hardcoded nonce/iv is passed to one of those algorithms, because reusing the same nonce/iv substantially lowers the algorithm's security

### Actual behavior
Nothing detected

expp121:
This rule looks for hard-coded nonce/ initialization vectors or variables which have hard-coded nonces/IV, that are given to multiple encryption algorithms such as: AES CBC,CTR,OFB,CFB,ASCON,CHACHA20

The rule has the following features:
* Ability to track hard-coded args passed directly to the function
* Ability to track variables which are assigned to hard-coded values
* Ability to track when a variable is assigned to the result of a function, it will go and inspect the function whether it has a call to cipher/rand.Read and flag it as vulnerable if it doesn't
*  Ability to backtrack when the encryption algorithm is in another fucntion (example `func cool(){ cipher.NewCTR(bloc,nonce)}`

closes #1196


expp121:
> @expp121 there are some lint errors, please could you have a look https://github.com/securego/gosec/actions/runs/10630482755/job/29469424412?pr=1197?

I was about to fix them! But thanks!

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+		// Check if it's a dereference operation (a.k.a pointer)
+		if valType.Op == token.MUL {
+			issueDescription += " by passing pointer which points to hardcoded variable"
+			tmp, hasErr := iterateThroughReferrers(val, funcsToTrack, pass.Analyzer.Name, issueDescription, pass.Fset, issue.High, issue.Low)
+			gosecIssue = append(gosecIssue, tmp...)
+			err = hasErr
+		}
+
+	// When the value assigned to a variable is a function call.
+	// It goes and check if this function contains call to crypto/rand.Read in it's body(Assuming that calling crypto/rand.Read in a function, is used for the generation of nonce/iv )
```
ccojocar, line 98: nit: could you break this comment in multiple lines. I would try to stay around 80 chars length.

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+	case *ssa.Call:
+		if calledFunction, ok := valType.Call.Value.(*ssa.Function); ok {
```
ccojocar, line 100: check also that `Value` is not nil

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+	case *ssa.Convert:
+		if valType.Type().String() == "[]byte" && valType.X.Type().String() == "string" {
+			issueDescription += " by passing converted string"
+			tmp, hasErr := iterateThroughReferrers(val, funcsToTrack, pass.Analyzer.Name, issueDescription, pass.Fset, issue.High, issue.High)
+			gosecIssue = append(gosecIssue, tmp...)
+			err = hasErr
+		}
+
+	case *ssa.Parameter:
+		//arg given to tracked function is wrapped in another function, example:
```
ccojocar, line 122: Add a space in front of these 3 comment lines.

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+				if funcArgs.Name() == valType.Name() && funcArgs.Type() == valType.Type() {
+					trackedFunctions[valType.Parent().String()] = []int{len(valType.Parent().Params), index}
+				}
+			}
+			result := iterateAndGetArgsFromTrackedFunctions(ssaFuncs, &trackedFunctions)
+
+			issueDescription += " by passing a parameter to a function and"
+			// recursively backtrack to where the origin of a variable passed to multiple functions is
+			for _, trackedVariable := range *result {
+
```
ccojocar, line 139: remove blank line

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+				gosecIssue = append(gosecIssue, tmp...)
+				err = hasErr
+			}
+		}
+	}
+	return gosecIssue, err
+}
+
+// Iterate through all places that use the `variable` argument and check if it's used in one of the tracked functions
+func iterateThroughReferrers(variable *ssa.Value, funcsToTrack *map[string][]int, analyzerID string, issueDescription string, fileSet *token.FileSet, issueSeverity issue.Score, issueConfidence issue.Score) ([]*issue.Issue, error) {
```
ccojocar: Break this function arguments on multiple line.

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+	if funcsToTrack == nil || variable == nil || analyzerID == "" || issueDescription == "" || fileSet == nil {
+		return nil, errors.New("received a nil object")
+	}
+	var gosecIssues []*issue.Issue = nil
```
ccojocar, line 154: no nil init required here

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+	// Go trough all functions that use the given arg variable
+	for _, referrer := range *(*variable).Referrers() {
```
ccojocar, line 156: check the `Referrers()` result pointer to not be nil before deference it.

ccojocar, line 156: I think, you don't need to dereference the `variable` pointer, the compiler does that automatically when calling the Referrers method.

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+					}
+				}
+			}
+		}
+	}
+	return false, nil
+}
+
+func addToVarsMap(value *ssa.Value, mapToAddTo *map[string]*ssa.Value) {
+	key := (*value).Name() + (*value).Type().String() + (*value).String() + (*value).Parent().String()
```
ccojocar, line 192: You don't need to deference the `value` when calling all these methods. The compiler does that automatically, and it is also safer when the `value` pointer is nil.

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+						return true, nil
+					}
+				}
+			}
+		}
+	}
+	return false, nil
+}
+
+func addToVarsMap(value *ssa.Value, mapToAddTo *map[string]*ssa.Value) {
```
ccojocar, line 191: map doesn't need to be passed by pointer, it is already a shared reference. You can just pass it around as it is.

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+	key := (*value).Name() + (*value).Type().String() + (*value).String() + (*value).Parent().String()
+	(*mapToAddTo)[key] = value
+}
+
+func isContainedInMap(value *ssa.Value, mapToCheck *map[string]*ssa.Value) bool {
```
ccojocar, line 196: ditt: not map as pointer

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+	key := (*value).Name() + (*value).Type().String() + (*value).String() + (*value).Parent().String()
```
ccojocar, line 197: ditto: do not dereference here

analyzers/hardcodedNonce.go
ccojocar: funcsToTrack map doesn't need to be passed as a pointer.

analyzers/hardcodedNonce.go
```diff
@@ -0,0 +1,231 @@
...
+		tmp, err := raiseIssue(savedArg, &calls, ssaPkgFunctions, pass, "")
+		issues = append(issues, tmp...)
+		if err != nil {
+			return nil, err
+		}
+	}
+	return issues, nil
+}
+
+func raiseIssue(val *ssa.Value, funcsToTrack *map[string][]int, ssaFuncs []*ssa.Function, pass *analysis.Pass, issueDescription string) ([]*issue.Issue, error) {
```
ccojocar, line 73: no map as pointer in the function args

ccojocar (changes requested):
Thanks for contributing this new analyzer. I left some comments, it would be great if you could address them.

analyzers/hardcodedNonce.go
expp121: I've tried but the intellisense complains, I am not sure whether it's because I have to dereference it or just the intellisense is misleading:
image

Sorry, but I am not sure, what you mean exactly?

ccojocar:
> I wasn't sure what you meant in all of your suggestions, but if there's something else that has to be change, please feel free to inform me!

Thanks for fixing my review comments. What I meant with `(*value)` is that you can directly use the `value.Method` if you pass the value not as pointer because it is already an interface.

I pushed the change in your pull request.

ccojocar (approved):

ccojocar (approved):

Jeeppler:
@expp121 Thanks for contributing the rule and the research effort required for it :rocket:.

@ccojocar Thank you for reviewing and merging it :tada: :+1: .

ccojocar:
@expp121 Please could you have a look at these two issues #1209 and  #1211 which we need to address since otherwise the analyzer is generating too many false positives. Thanks

PR #1201, opened 2024-08-30
fix: G602 support for nested conditionals with bounds check

### LINKED ISSUE #1200: G602: False positives for slices with bounds check on a higher hierarchy
### Summary 
Noticed that we're receiving errors for the G602 (slices out of bounds) rules when we have proper bound checks 
Code sample (inspired by the current unit tests):
```go
package main

import "fmt"

func main() {
	s := make([]byte, 0)
	if len(s) > 0 {
		switch s[0] {
		case 0:
			fmt.Println("zero")
			return
		default:
			fmt.Println(s[0])
			return
		}
	}
}
```

```go
package main

import "fmt"

func main() {
	s := make([]byte, 0)
	if len(s) > 0 {
		switch s[0] {
		case 0:
			b := true
			if b == true {
				// Should work for many-levels of nesting when the condition is not on the target slice
				fmt.Println(s[0])
			}
			return
		default:
			fmt.Println(s[0])
			return
		}
	}
}
```

``` go
package main

import "fmt"

func main() {
	s := make([]byte, 0)
	if len(s) > 0 {
		if len(s) > 1 {
			fmt.Println(s[1])
		}
		fmt.Println(s[0])
	}
}
```
### gosec version
`v2.20.1-0.20240826145712-bcec04e78483`
`go version go1.23.0 darwin/arm64`
Mac OS Sonoma 14.6.1 (M1/arm64)
### Expected behavior
I'd expect these samples to not return any errors
### Actual behavior
They all return errors since the bounds check does not look deeper into conditionals

xWiiLLz:
Trying to fix  #1200 with this PR.
Suggestions/comments more than welcome

### DIFF
--- a/analyzers/slice_bounds.go
+++ b/analyzers/slice_bounds.go
@@ -118,8 +118,12 @@ func runSliceBounds(pass *analysis.Pass) (interface{}, error) {
 			if i == 1 {
 				bound = invBound(bound)
 			}
-			var processBlock func(block *ssa.BasicBlock)
-			processBlock = func(block *ssa.BasicBlock) {
+			var processBlock func(block *ssa.BasicBlock, depth int)
+			processBlock = func(block *ssa.BasicBlock, depth int) {
+				if depth == maxDepth {
+					return
+				}
+				depth++
 				for _, instr := range block.Instrs {
 					if _, ok := issues[instr]; ok {
 						switch bound {
@@ -146,13 +150,13 @@ func runSliceBounds(pass *analysis.Pass) (interface{}, error) {
 						}
 					} else if nestedIfInstr, ok := instr.(*ssa.If); ok {
 						for _, nestedBlock := range nestedIfInstr.Block().Succs {
-							processBlock(nestedBlock)
+							processBlock(nestedBlock, depth)
 						}
 					}
 				}
 			}
 
-			processBlock(block)
+			processBlock(block, 0)
 		}
 	}
 
--- a/analyzers/slice_bounds.go
+++ b/analyzers/slice_bounds.go
@@ -118,32 +118,41 @@ func runSliceBounds(pass *analysis.Pass) (interface{}, error) {
 			if i == 1 {
 				bound = invBound(bound)
 			}
-			for _, instr := range block.Instrs {
-				if _, ok := issues[instr]; ok {
-					switch bound {
-					case lowerUnbounded:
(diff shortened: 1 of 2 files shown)


analyzers/slice_bounds.go
```diff
@@ -118,32 +118,41 @@ func runSliceBounds(pass *analysis.Pass) (interface{}, error) {
...
-								delete(issues, instr)
-							}
-						case *ssa.IndexAddr:
-							indexValue, err := extractIntValue(tinstr.Index.String())
-							if err != nil {
-								break
-							}
-							if isSliceIndexInsideBounds(0, value, indexValue) {
-								delete(issues, instr)
+			var processBlock func(block *ssa.BasicBlock)
```
ccojocar, line 121: nit: I would use the short variable declaration here.

analyzers/slice_bounds.go
```diff
@@ -118,32 +118,41 @@ func runSliceBounds(pass *analysis.Pass) (interface{}, error) {
...
+									break
+								}
+								if isSliceIndexInsideBounds(0, value, indexValue) {
+									delete(issues, instr)
+								}
 							}
 						}
+					} else if nestedIfInstr, ok := instr.(*ssa.If); ok {
+						for _, nestedBlock := range nestedIfInstr.Block().Succs {
+							processBlock(nestedBlock)
```
ccojocar, line 149: I would keep track of the depth of the recursion, and stop after 10 calls to avoid stack overflows and infinite loops.

ccojocar (changes requested):
Thanks for contributing this fix. I left a few comments, it would be great if you could address them. Thanks again!

analyzers/slice_bounds.go
xWiiLLz: You mean `processBlock := func(...)`? That's what I tried doing at first but the compiler yells since we're calling it recursively; I assume it has to see the declaration in its scope to use it.

analyzers/slice_bounds.go
xWiiLLz: Sounds good, just noticed you had used that same pattern elsewhere (my bad). Will use `maxDepth` count instead of 10 as this is the pattern in use currently for `checkAllSlicesBounds` and `trackSliceBounds`

ccojocar (approved):

PR #1210, opened 2024-09-04
Rollback the SARIF version to 2.1 since github doesn't support 2.2

ccojocar:


niij:
^ I have the same issue as the linked PR above (@Michad). Our builds are failing with a rejection from github that the SARIF file is invalid. This is occurring when using both of the following actions:
- securego/gosec@master (v2.21.1 014751c91c04fe3fe05e381dea9faadbd4ce8e78) 
- github/codeql-action/upload-sarif (v2.26.6 be8b74c09c1778bcdbea38e1a45efea2cb73e18c)

I'm pinning securego/gosec@v2.20.0 as a workaround for now.

Issue #1212, opened 2024-09-04
G115 is reporting false positives (a summary)


czechbol:
### Summary 
G115 is still reporting false positives even after #1189 and #1194.

### Known false positives
any kind of arithmetic done on the checked variable after the check:
```go
func foo(a int) uint {
    if a != 3 || a != 4 {
        panic("not supported")
    }
    return uint(a-1)
}
```
binary truncation (reported by @stephenc):
```go
func foo(a int) uint16 {
    return uint16(a &0xffff)
}
```
builtin min and max functions (reported by @ben-krieger):
```go
func foo() uint16 {
    a, b := 1234, 2345
    res := min(a,b)

    return uint(result)
}
```

### Expected behavior
These shouldn't raise any warnings.

### Actual behavior
They raise G115 warnings.

### Notes:
Do not attempt to implement this as a single PR, the changes necessary to support everything will be pretty complex. It will be much better to open small PRs that fix one issue at a time.

ccoVeille:
At least these use cases are less frequent than the previous ones.

deerbone:
I'd say that people will hit the first one pretty frequently, and it also seems to me it's the most difficult to tackle without bringing in any false negatives.

Victoremepunto:
I think I may be hitting this: https://github.com/RedHatInsights/frontend-operator/pull/188/files#diff-e8b328ffb5ced885894c80b942846601597e05020f831c2b1442d8fb8a654c8dR120

czechbol:
@Victoremepunto you're getting that warning since golangci-lint has not released a new version with the changes from #1194.

@ccoVeille do you know and can share at least approximately when we can expect it?

ccoVeille:
I'm just a random gopher. I'm pretty active, but I'm not part of golangci-lint release.

@ldez might be a better candidate to reply that question

PlasmaPower:
I'm not sure if I should open a separate issue for this, but another false positive is casting the index produced by iterating over a range to an unsigned int. We do this a lot in our codebase. E.g.:

```go
myArr := []string{"hello", "world"}

for i := range len(myArr) {
    _ = uint64(i) // shouldn't error but does with G115
}

for i, _ := range myArr {
    _ = uint64(i) // shouldn't error but does with G115
}
```

Ideally it'd be able to look at bounds in for loops as well, but I get that's more complicated:

```go
for i := 0; i < 10; i++ {
    _ = uint64(i) // shouldn't error but does with G115
}

[...]

PR #1221, opened 2024-09-12
G115 Struct Attribute Checks

czechbol:
Similarly to the arithmetic example in #1212, the value being checked is, according to the SSA, different from the value being converted. So we have to be a bit smarter. Not as smart as with the arithmetic though.

I also found that explicit checks are still lacking, I partly fixed them and I'll document the rest in the issue.

### DIFF
--- a/analyzers/conversion_overflow.go
+++ b/analyzers/conversion_overflow.go
@@ -40,7 +40,7 @@ type integer struct {
 type rangeResult struct {
 	minValue             int
 	maxValue             uint
-	explixitPositiveVals []uint
+	explicitPositiveVals []uint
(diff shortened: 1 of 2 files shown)


czechbol:
Thanks @Victoremepunto for reporting this issue twice, I didn't see it the first time.

Victoremepunto:
> Thanks @Victoremepunto for reporting this issue twice, I didn't see it the first time. I've tested this PR against your codebase and it's not reporting G115 anymore.

Saw that. Thanks, I appreciate the help, looking forward to the release!

ccojocar (approved):

PR #1272, opened 2024-12-16
Rule documentation updates

dannyc-grafana:
More docs updates for ya'll. If you'd prefer I open issues before making PRs let me know.

### DIFF
--- a/RULES.md
+++ b/RULES.md
@@ -1,0 +1,61 @@
+# Rule Documentation
+
+## Rules accepting parameters
+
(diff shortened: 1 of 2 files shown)


RULES.md
```diff
@@ -0,0 +1,61 @@
+# Rule Documentation
```
ccojocar, line 1: This is fine for configuration but we started a while a go this project to document each rule https://github.com/securego/securego.github.io. If you are interested in this area, you can contribute there. Thanks

ccojocar (approved):

PR #1328, opened 2025-04-01
remove G113. It only affects old/unsupported versions of Go

niij:
Newer versions of go (>=1.16.14, >=1.17.7, 1.18+) are not affected by this. Don't warn at all on those newer versions. See https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-23772


rules/math_big_rat.go
```diff
@@ -21,13 +21,12 @@ func (r *usingOldMathBig) Match(node ast.Node, ctx *gosec.Context) (gi *issue.Is
 		return nil, nil
 	}
 
-	confidence := issue.Low
 	major, minor, build := gosec.GoVersion()
 	if major == 1 && (minor == 16 && build < 14 || minor == 17 && build < 7) {
```
ccoVeille, line 25: Shouldn't be this? 🤔

```suggestion
	if major == 1 && (minor < 16 || minor == 16 && build < 14 || minor == 17 && build < 7) {
```

niij: Good catch! It should have been that way from the beginning, I assume.

ccoVeille (approved):

ccojocar (approved):

ccojocar:
Thanks for this contribution but we don't support these versions anymore.

ccojocar:
I think we can deprecate complete the check.

ccojocar:
@niij There is a test failing. Please can you update it, and I would suggest to remove completely the check. Thanks

niij:
Oh I see why it's failing. Since it expects to catch this failure _at some severity_ but the tests are running on the latest version of Go, so it's excluded.

Go 1.17 went out of support August 2022. Seems reasonable to drop support for this version-specific issue at this point. If you're compiling old versions of 1.16/1.17 and processing untrusted user input through them at this point you likely have bigger problems :)

I'll delete it today.

README.md
```diff
@@ -138,7 +138,7 @@ directory you can supply `./...` as the input argument.
 - G110: Potential DoS vulnerability via decompression bomb
 - G111: Potential directory traversal
 - G112: Potential slowloris attack
-- G113: Usage of Rat.SetString in math/big with an overflow (CVE-2022-23772)
+- G113: ~~Usage of Rat.SetString in math/big with an overflow (CVE-2022-23772)~~. This has been removed. This affected Go <1.16.14 and Go <1.17.7, which are no longer supported versions in gosec.
```
ccojocar, line 141: Please remove it from here, and add it to https://github.com/securego/gosec?tab=readme-ov-file#retired-rules.

ccojocar (changes requested):

ccojocar:
@niij There is a lint issue. Please could you fix it? Thanks

niij:
@ccojocar `gofmt` done

ccojocar (approved):

niij:
thanks

PR #1332, opened 2025-04-03
Update version in 'action.yml' to 2.22.3 (anticipating next version)

sjenna:
Hello! 👋 

This change is needed for GitHub Actions to pull the correct version. Ideally this would get merged before the release is created, so Dependabot pulls the right version when updating GitHub Actions.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -10,7 +10,7 @@ inputs:
 
 runs:
     using: 'docker'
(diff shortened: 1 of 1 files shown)


ccojocar:
Thanks for this update, v2.22.3 is not yet released. I'll merge it when is released.

ccojocar (approved):

PR #1384, opened 2025-09-12
rules(G202): detect SQL concat in ValueSpec declarations; add test sa…

### LINKED ISSUE #1309: The way of string concatenation cannot be detected correctly
the code as following：
`package main

import (
    "database/sql"
    "fmt"
    _ "github.com/lib/pq"
)

func main() {
    db, err := sql.Open("postgres", "user=postgres password=password dbname=mydb sslmode=disable")
    if err!= nil {
        panic(err)
    }
    defer db.Close()

    var username string
    fmt.Println("请输入用户名:")
    fmt.Scanln(&username)

    var query string = "SELECT * FROM users WHERE username = '" + username + "'"
    rows, err := db.Query(query)
    if err!= nil {
        panic(err)
    }
    defer rows.Close()

    // 处理查询结果
}`

eshentials:
Summary
-Fixes a gap in G202 where SQL string concatenation defined via ValueSpec declarations wasn’t detected (e.g., var query string = "..." + user).
-Adds a test sample mirroring issue #1309’s reproduction.
-Keeps existing behavior for AssignStmt-based concatenations; extends coverage to ValueSpec.

Motivation
-Issue #1309 shows a common pattern where a query string is built at declaration time via concatenation, then passed to db.Query/Exec. G202 detected AssignStmt concatenations, but missed ValueSpec concatenations. This PR resolves that gap.

What changed
-rules/sql.go (G202): In sqlStrConcat.checkQuery(...), when the SQL argument is an identifier, also handle declarations of type ast.ValueSpec. Reuses findInjectionInBranch on ValueSpec.Values to detect tainted concatenations.

-testutils/g202_samples.go: Added a new CodeSample using Postgres that initializes a query via var query string = "SELECT ..." + username and passes it to db.Query. Expected errors: 1.

Behavior before/after

Before: G202 flagged string concatenation when built via assignment or inline, but not when built at declaration with var query string = "..." + user.

After: G202 flags all these forms, including the ValueSpec case.
Example covered by the new sample

Detected:
var query string = "SELECT ... WHERE username = '" + username + "'"
rows, err := db.Query(query)

Testing
Ran rules tests: 42 passed, 0 failed.
New sample triggers 1 G202 finding as expected.

Risk/compatibility
Low risk: uses the existing binary-concatenation detection and SQL pattern-matching logic; extends it to another declaration form without changing rule semantics.
No user-facing config changes.

Future work
Broader SQL static analysis (e.g., parse queries/params) and integration with tools like sqlvet is out of scope for this PR, but can be explored in follow-ups.

Fixes #1309

### DIFF
--- a/rules/sql.go
+++ b/rules/sql.go
@@ -191,6 +191,11 @@ func (s *sqlStrConcat) checkQuery(call *ast.CallExpr, ctx *gosec.Context) (*issu
 			if injection := s.findInjectionInBranch(ctx, decl.Rhs); injection != nil {
 				return ctx.NewIssue(injection, s.ID(), s.What, s.Severity, s.Confidence), nil
 			}
+		case *ast.ValueSpec:
+			// handle: var query string = "SELECT ...'" + user
+			if injection := s.findInjectionInBranch(ctx, decl.Values); injection != nil {
+				return ctx.NewIssue(injection, s.ID(), s.What, s.Severity, s.Confidence), nil
+			}
 		}
 	}
 
--- a/testutils/g202_samples.go
+++ b/testutils/g202_samples.go
@@ -308,4 +308,32 @@ func main() {
 	fmt.Println(result)
 }
 `}, 0, gosec.NewConfig()},
+	{[]string{`
+package main
+
+import (
+	"database/sql"
+	"fmt"
+	_ "github.com/lib/pq"
+)
+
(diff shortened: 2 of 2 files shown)


eshentials:
@ccojocar I have also worked on the SQL string concatenation, please check.

PR #1395, opened 2025-09-29
Fix/ssa panic log

### LINKED ISSUE #1394: Broken SSA panic message
seems like there is a error/panic being swallowed as that error in that current state is not really helpful
```
[gosec] 2025/09/29 09:37:40 Panic when running SSA analyser on package: trace
[gosec] 2025/09/29 09:37:40 Error building the SSA representation of the package "trace": %!s(<nil>)
```

would be helpful if it would be logged here:
https://github.com/securego/gosec/blob/0d8255e75e4957f2469bf048a461484854419eab/analyzer.go#L488

the lower log level is because of this:
https://github.com/securego/gosec/blob/0d8255e75e4957f2469bf048a461484854419eab/analyzer.go#L416

sruehl:
Fixes #1394


sruehl:
@ccojocar mind reviewing so I can figure out what goes wrong on my scan?

PR #1396, opened 2025-09-29
check nil slices, partially check bounds

### LINKED ISSUE #1392: G602 doesn't detect unallocated slices
```go
package main

func main() {
	var a []int
	for i := range 5 {
		a[i] = 0
	}

	b := make([]int, 0, 100)
	for i := range 5 {
		b[i] = 0
	}
}
```

```shell
%go install github.com/securego/gosec/v2/cmd/gosec@latest

% gosec -include=G602 ./...
[gosec] 2025/09/29 09:48:53 Including rules: G602
[gosec] 2025/09/29 09:48:53 Excluding rules: default
[gosec] 2025/09/29 09:48:53 Including analyzers: G602
[gosec] 2025/09/29 09:48:53 Excluding analyzers: default
[gosec] 2025/09/29 09:48:53 Import directory: /Users/kondratevpavel/make/test3/bonds
[gosec] 2025/09/29 09:48:54 Checking package: main
[gosec] 2025/09/29 09:48:54 Checking file: /Users/kondratevpavel/make/test3/bonds/main.go
Results:


Summary:
  Gosec  : dev
  Files  : 1
  Lines  : 13
  Nosec  : 0
  Issues : 0
```

```shell
% go run main.go 
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
main.main()
        /Users/kondratevpavel/make/test3/bonds/main.go:6 +0x24
exit status 2
```

kondratev:
fix #1392 partially

### DIFF
--- a/analyzers/slice_bounds.go
+++ b/analyzers/slice_bounds.go
@@ -372,7 +372,7 @@ var errExtractBinOp = fmt.Errorf("unable to extract constant from binop")
 func extractBinOpBound(binop *ssa.BinOp) (bound, int, error) {
 	if binop.X != nil {
 		if x, ok := binop.X.(*ssa.Const); ok {
-			if x == nil || ok {
+			if x.Value == nil {
 				return lowerUnbounded, 0, errExtractBinOp
 			}
 			value, err := strconv.Atoi(x.Value.String())
@@ -393,7 +393,7 @@ func extractBinOpBound(binop *ssa.BinOp) (bound, int, error) {
 	}
 	if binop.Y != nil {
 		if y, ok := binop.Y.(*ssa.Const); ok {
-			if y == nil || ok {
+			if y.Value == nil {
 				return lowerUnbounded, 0, errExtractBinOp
 			}
 			value, err := strconv.Atoi(y.Value.String())
--- a/analyzers/slice_bounds.go
+++ b/analyzers/slice_bounds.go
@@ -119,24 +119,25 @@ func runSliceBounds(pass *analysis.Pass) (interface{}, error) {
 							}
 						}
 					case *ssa.Alloc:
-						typeStr := indexInstr.Type().String()
-						arrayLen, err := extractArrayAllocValue(typeStr) // preallocated array
-						if err != nil {
-							break
-						}
+						if instr.Pos() > 0 {
+							typeStr := indexInstr.Type().String()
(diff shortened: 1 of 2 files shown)


analyzers/slice_bounds.go
```diff
@@ -104,6 +104,22 @@ func runSliceBounds(pass *analysis.Pass) (interface{}, error) {
 							}
 						}
 					}
+				case *ssa.Store: // check store to nil slice or out of bounds
+				case *ssa.IndexAddr:
+					if constantValue, ok := instr.X.(*ssa.Const); ok && constantValue.Type().String()[:2] == "[]" {
+						if constantValue.Value == nil {
```
ccojocar: Shouldn't this check be placed before accessing the `.Type()`? What happens if is already nil?

analyzers/slice_bounds.go
```diff
@@ -104,6 +104,22 @@ func runSliceBounds(pass *analysis.Pass) (interface{}, error) {
 							}
 						}
 					}
+				case *ssa.Store: // check store to nil slice or out of bounds
+				case *ssa.IndexAddr:
+					if constantValue, ok := instr.X.(*ssa.Const); ok && constantValue.Type().String()[:2] == "[]" {
```
ccojocar: Can the type length be less than 2 under any circumstances?

ccojocar (commented):
Thank you for contributing this fix. Please add some tests in the rule test file to cover this use case https://github.com/securego/gosec/blob/master/testutils/g602_samples.go.

analyzers/slice_bounds.go
kondratev: if type name string less then []*, then this is not slice in this case

analyzers/slice_bounds.go
kondratev: The code assumes that someone will continue to search for errors. For example, when var a [5]int this const value not nil, in this case the simplest error is determined when instead of append(a,...) index assignment is used.

kondratev:
@ccojocar review please

ccojocar:
There are some lint warnings which need to be fixed before merging, otherwise LGTM.

ccojocar:
It seems that there is a test failure.

ccojocar:
There are still some issues when the gosec built from this pull request. PTAL. Thanks

Issue #1393, opened 2025-09-29
SARIF upload broken with github/codeql-action/upload-sarif@v3


sruehl:
apparently the sarif file is broken with `github/codeql-action/upload-sarif@v3`
```
Error: Unable to upload "results.sarif" as it is not valid SARIF:
- instance.runs[0].results[26].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[26].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[27].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[27].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[28].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[28].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[29].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[29].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[30].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[30].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[31].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[31].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[32].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[32].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[33].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[33].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[34].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[34].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[35].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[35].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[36].fixes[0].description is not any of [subschema 0],[subschema 1]
- instance.runs[0].results[36].fixes[0].artifactChanges is not of a type(s) array
- instance.runs[0].results[37].fixes[0].description is not any of [subschema 0],[subschema 1]
[...]

PR #1397, opened 2025-10-01
fix: text must be supplied when markdown is used

sruehl:
as noted in #1393 and stated in the schema https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790720 (considering this is the right version) Text SHALL be present when Markdown is used. This PR just uses the autofix for text right now. A further issue can be opened if the autofix is indeed markdown but for now this PR should at least enable SRAIF upload again

Fixes https://github.com/securego/gosec/issues/1393

### DIFF
--- a/report/sarif/builder.go
+++ b/report/sarif/builder.go
@@ -91,6 +91,8 @@ func NewResult(ruleID string, ruleIndex int, level Level, message string, suppre
 		result.Fixes = []*Fix{
 			{
 				Description: &Message{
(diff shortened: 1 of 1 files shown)


sruehl:
@ccojocar this PR is ready for review and kind of urgent because it seems that Github (or github/codeql-action/upload-sarif@v3) is enforcing a stricter schema validation now

Issue #1412, opened 2025-11-04
Build tags not propagating correctly


matthewhartstonge:
Howdy! 🤠

## Summary

SSA panics are occurring when performing static analysis due to references not being included when build tags are involved.

## Description

I've been seeing static analysis panics occurring when references can't be found that should have been included by build tags. 

See failing `gosec` report log here: https://github.com/matthewhartstonge/exp-buildtag/blob/main/gosec-log.txt

## Prior Art

I noticed that a couple of issues and PRs in the past have attempted to fix this issue:
* https://github.com/securego/gosec/issues/393
* https://github.com/securego/gosec/pull/481

## Workaround

You can currently set the `GOFLAGS` envvar with your expected build tags, gosec will then run as expected.

```
export GOFLAGS='-tags=tag'
```

## Reproducible

Refer: https://github.com/matthewhartstonge/exp-buildtag

## Fix

I have a PR incoming...

PR #1053, opened 2023-10-18
Fix data race in analyzer

ccojocar:
- Fix data race in the analyzer
- Run the gosec with data race detector active during tests

fixes #1049

PR #1413, opened 2025-11-04
fix: build tag parsing.

matthewhartstonge:
Aims to fix https://github.com/securego/gosec/issues/1412

Couple of thoughts:
* I removed the mutex as I build the required config on load due to requiring transforming. This could be optimised by making `load()` take in both `buildTags` _and_ a `*package.Config` so it's only processed and in memory once.
* The previous build tag test never compiled, which leads to a panic, which is then hidden due to the graceful recover code. Wondering if it's best to remove the test? I changed the test sample name from `SampleCodeBuildTag ` => `SampleCodeCompilationFail` and added `should not panic if a file can not compile` as a test case

### DIFF
--- a/analyzer_test.go
+++ b/analyzer_test.go
@@ -814,7 +814,7 @@ var _ = Describe("Analyzer", func() {
 			Expect(err).ShouldNot(HaveOccurred())
 
 			issues, _, _ := analyzer.Report()
-			Expect(issues).Should(HaveLen(0))
+			Expect(issues).Should(BeEmpty())
 		})
 
 		It("should attempt to analyse a file with build tags", func() {
--- a/testutils/g601_samples.go
+++ b/testutils/g601_samples.go
@@ -2,10 +2,9 @@ package testutils
 
 import "github.com/securego/gosec/v2"
 
-var (
-	// SampleCodeG601 - Implicit aliasing over range statement
-	SampleCodeG601 = []CodeSample{
-		{[]string{`
+// SampleCodeG601 - Implicit aliasing over range statement
+var SampleCodeG601 = []CodeSample{
+	{[]string{`
 package main
 
 import "fmt"
@@ -40,7 +39,7 @@ func main() {
 	fmt.Printf("%d %v %s", zero, c_star, c)
 }
 `}, 1, gosec.NewConfig()},
-		{[]string{`
+	{[]string{`
 // see: github.com/securego/gosec/issues/475
 package main
 
@@ -56,7 +55,7 @@ func main() {
 	}
 }
 `}, 0, gosec.NewConfig()},
-		{[]string{`
+	{[]string{`
 package main
 
 import (
@@ -77,7 +76,7 @@ func main() {
 	}
 }
 `}, 0, gosec.NewConfig()},
-		{[]string{`
+	{[]string{`
 package main
 
 import (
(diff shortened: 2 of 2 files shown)


ccojocar:
Please could you fix the lint warnings?

analyzer.go
```diff
@@ -187,7 +187,6 @@ type Analyzer struct {
 	trackSuppressions bool
 	concurrency       int
 	analyzerSet       *analyzers.AnalyzerSet
-	mu                sync.Mutex
```
ccojocar, line 190: Why was required to remove this mutex? How is related to this change?

analyzer.go
```diff
@@ -362,10 +353,12 @@ func (gosec *Analyzer) load(pkgPath string, conf *packages.Config) ([]*packages.
...
 	}
 
-	// step 3/3 remove build tags from conf to proceed build correctly.
-	gosec.mu.Lock()
-	conf.BuildFlags = nil
-	defer gosec.mu.Unlock()
+	// step 2/2: pass in cli encoded build flags to build correctly.
+	conf := &packages.Config{
+		Mode:       LoadMode,
+		BuildFlags: CLIBuildTags(buildTags),
```
ccojocar, line 359: Ins't this the only fix required to get properly the build flags propagated? Why's the reason for the extended refactoring?

analyzer.go
matthewhartstonge: This mutex was only used when mutating the config (refer: https://github.com/securego/gosec/pull/1053/files). But the mutation of the config causes potential issues for the next worker.

Given the config was passed in via a pointer to `func (gosec *Analyzer) load(pkgPath string, conf *packages.Config)`, the previous code lines mutate the top level config, potentially removing the build flags for the next worker:

```go
	gosec.mu.Lock()
	conf.BuildFlags = nil
	defer gosec.mu.Unlock()
```

analyzer.go
matthewhartstonge: Couple of reasons:
* The default builder wants _build tags_ in the form `[]string{"tag1", "tag2"}`
* `packages.Load` wants cli _build flags_ in the form of `[]string{"-tags=tag1,tag2", "-ldflags=\"-s -w\""}`
* Modifying the top level config here by nil'ing out the build flags here could remove build tags for the next worker.

The change was made to stop top level config mutation (passed in as a pointer). By defining the config locally the mutex isn't required.

In the PR description I mentioned a possible refactor: moving the config creation back and making `load()` take in both `buildTags []string` and the fixed `conf *packages.Config` which could reduce code churn.

However, passing a shared `conf *packages.Config` again risks future contributors accidentally mutating it, so preferred creating the config where it's required.

Other option is to deep-copy the config before calling `gosec.load()`.

Happy to refine things as required! 😄

testutils/pkg.go
```diff
@@ -27,6 +27,17 @@ type TestPackage struct {
 	build  *buildObj
 }
 
+// Option provides a way to adjust the package config depending on testing
+// requirements
+type Option func(conf *packages.Config)
+
+// WithBuildTags enables injecting build tags into the package config on build.
+func WithBuildTags(tags []string) Option {
```
matthewhartstonge, line 35: To not break the API for existing tests, I decided to use variadic options to inject expected build tags under test.

analyzer.go
ccojocar: Can you please run the `make build-race` and post the result? Thanks

Issue #1407, opened 2025-10-16
G304 produces artifactChanges object set to null causing SARIF validation failure


komish:
I've got a gosec SARIF that contains this block (truncated for clarity):

```json
{
	"runs": [
		{
			"results": [
				{
					"fixes": [
						{
							"artifactChanges": null,
							"description": {
								"markdown": "Consider using os.Root to scope file access under a fixed root (Go \u003e=1.24). Prefer root.Open/root.Stat over os.Open/os.Stat to prevent directory traversal.",
								"text": "Consider using os.Root to scope file access under a fixed root (Go \u003e=1.24). Prefer root.Open/root.Stat over os.Open/os.Stat to prevent directory traversal."
							}
						}
					],
					"level": "error",
					"locations": [
						// ...
					],
					"message": {
						"text": "Potential file inclusion via variable"
					},
					"ruleId": "G304"
				},

		// ...
```

When using `github/codeql-action/upload-sarif@f443b600d91635bebf5b0d9ebc620189c0d6fba5 # v4.30.8` to upload the SARIF, I get a schema validation failure because the [schema](https://github.com/github/codeql-action/blob/ee753b4724a8022c6235aff6a88efbfd82155923/src/sarif-schema-2.1.0.json#L978-L1004) indicates artifactChanges is required and must be an array. 

From what I can tell, the [autoFix logic](https://github.com/securego/gosec/blob/0adab9d9bee15a7d79feeddb2998961c97dd9c19/report/sarif/builder.go#L90-L100) JUST adds "description" values, but no artifactChanges. I'm not sure where artifactChanges is supposed to come from.

If it helps, I see AI-powered autofix suggestions in the code, but I'm not using it. For this rule, I believe the static [osRootSuggestion](https://github.com/securego/gosec/blob/0adab9d9bee15a7d79feeddb2998961c97dd9c19/rules/readfile.go#L114-L122) is what's being used.

For now, I've just disabled rule G304.

GoSec Version I'm using is here, being used via direct binary installation.
```
GOSEC_VERSION ?= 6be2b51fd78feca86af91f5186b7964d76cb1256 # v2.22.10
```

ccojocar:
@sruehl please could you have a look at this issue? I think is related to #1397. Thanks

sruehl:
fun... same conclusion: "not sure where artifactChanges is supposed to come from". 
looking at this
```json
    "artifactChange": {
      "description": "A change to a single artifact.",
      "additionalProperties": false,
[...]

PR #1417, opened 2025-11-07
test: add sarif validation

sruehl:
relates to #1407.

This PR adds a schema validation to the tests to spot issues with wrong schema early


sruehl:
@ccojocar can you have a look at this PR? it seems that the SARIF reports are all over the place regarding the Schema. I think the validation is vital going forward with the fixes in #1407 for example

sruehl:
one sec, I updated this to v6 validation. For some reason I picked up v5

Issue #1429, opened 2025-11-30
Feature request: Add check for Trojan Source attacks (CVE-2021-42574)


kfess:
**Feature request**

gosec currently does not detect Trojan Source attacks (CVE-2021-42574).
This attack uses Unicode bidirectional control characters to make source code appear different to reviewers than what compilers actually execute.

**What I would like to add**

Add a new rule that detects Unicode Bidi control characters in source code.

**Additional context**

- Bandit (Python) already has this check as [B613](https://bandit.readthedocs.io/en/latest/plugins/b613_trojansource.html)
- Reference: https://trojansource.codes/

I'm happy to submit a PR if this feature is welcomed.

kfess:
I'd like to work on this feature.
I will submit a PR to add a new rule.

ccojocar:
fixed by #1431

PR #1431, opened 2025-12-02
feature: add rule for trojan source

kfess:
### Description

Add rule to detect Trojan Source attacks (CVE-2021-42574) using bidirectional Unicode control characters as G116.

The rule scans entire file content to detect dangerous bidirectional text control characters:

RLO (U+202E), LRI/RLI/FSI (U+2066/2067/2068), LRE/RLE (U+202A/202B), PDF (U+202C), LRO (U+202D), RLM/LRM (U+200F/200E)

References:
- https://trojansource.codes/trojan-source.pdf
- https://www.compart.com/en/unicode/U+200f
- https://www.compart.com/en/unicode/U+200e

### Fixes

Issue: #1429 

### Comment

`testutils/g116_samples.go` uses `#nosec` because G116 scans entire file content (not just AST nodes but also comments). The test file contains actual bidirectional Unicode characters to verify detection works correctly. Without `#nosec`, gosec would flag its own test samples when running `make test`.

### DIFF
--- a/testutils/g116_samples.go
+++ b/testutils/g116_samples.go
@@ -10,153 +10,19 @@ import "github.com/securego/gosec/v2"
 var (
 	// SampleCodeG116 - TrojanSource code snippets
 	SampleCodeG116 = []CodeSample{
-		{[]string{`
-package main
-
-import "fmt"
-
-func main() {
-	// This comment contains bidirectional unicode: access‮⁦ granted⁩‭
-	isAdmin := false
-	fmt.Println("Access status:", isAdmin)
-}
-`}, 1, gosec.NewConfig()},
-		{[]string{`
-package main
-
-import "fmt"
-
-func main() {
-	// Trojan source with RLO character
-	accessLevel := "user"
-	// Actually assigns "nimda" due to bidi chars: accessLevel = "‮nimda"
-	if accessLevel == "admin" {
-		fmt.Println("Access granted")
-	}
-}
-`}, 1, gosec.NewConfig()},
-		{[]string{`
-package main
-
-import "fmt"
-
-func main() {
-	// String with bidirectional override
-	username := "admin‮ ⁦Check if admin⁩ ⁦"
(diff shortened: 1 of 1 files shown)


rules/trojansource.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+func (r *trojanSource) Match(node ast.Node, c *gosec.Context) (*issue.Issue, error) {
+	if file, ok := node.(*ast.File); ok {
+		fobj := c.FileSet.File(file.Pos())
+		if fobj == nil {
+			return nil, nil
+		}
+
+		content, err := os.ReadFile(fobj.Name())
```
ccojocar, line 27: Does this perform well with large files? 

I would use something like to make sure that we don't run out of memory and have performance issues:
```
file, err := os.Open("")
if err != nil {
    log.Fatal(err)
}
defer file.Close()
scanner := bufio.NewScanner(file)
for scanner.Scan() {
    line := scanner.Text()
    // Process line
}
if err := scanner.Err(); err != nil {
    log.Fatal(err)
}

```

ccojocar:
Please can you rebase this pull request? Thanks

kfess:
Thanks for the review.
I've updated the implementation to use `bufio.Scanner` instead of `os.ReadFile` for better memory efficiency.

I evaluated the memory efficiency with `/usr/bin/time` on a large project (`kubernetes/kubernetes`, 10,123 files):

```bash
/usr/bin/time -v go run ./cmd/gosec/ --include=G116 ../kubernetes/...
```

- Before (ReadFile): 342 MB
- After (Scanner): 194 MB

I also rebased the commits.

ccojocar:
It seems that there are still some lint issues. PTAL thanks

kfess:
Sorry for the late response.
Thank you for fixing the linter error.

kfess:
I forgot to add the new rule to the README. Should I open a PR for that?

PR #1451, opened 2026-01-03
feat: add secrets serialization G117

oittaa:
https://github.com/securego/gosec/issues/1416

Does G117 make sense as the new number?

Are there any other things that need to be taken into account?


oittaa:
CWE-499: Serializable Class Containing Sensitive Data seems to fit the best. Other potentially good one would be CWE-532: Insertion of Sensitive Information into Log File

oittaa:
One more thing, I just set severity and confidence to medium quite arbitrarily. Is there some official way to classify those?
https://github.com/securego/gosec/pull/1451/files#diff-0c5053436f148f2c3ee7341e535b8fa3bfa0858058cd447e1cfa92a10ef01ef2R119-R120

ccojocar:
Thanks, it looks good now. I think medium sounds good to me.

Issue #1240, opened 2024-10-10
#nosec doesn't work if it's preceded by an open bracket {


paul-at-cybr:
### Summary 

Nosec doesn't work if there's an open bracket between the comment and the issue it's addressing 

### Steps to reproduce the behavior
```go
package main

import "fmt"

func main() {
	ten := 10
	configVal := uint(ten)
	inputSlice := []int{1, 2, 3, 4, 5}
	if len(inputSlice) <= int(configVal) { // #nosec G115 -- this fails
		fmt.Println("hello world!")
	}

	if len(inputSlice) <= int(configVal) /* #nosec G115 -- this works, but is horrible */ {
		fmt.Println("hello world!")
	}
}

```
### gosec version
```
Version: 2.21.4
Git tag: v2.21.4
Build date: 2024-09-26T11:55:22Z
```

### Go version (output of 'go version')
`go version go1.23.2 linux/amd64`

### Operating system / Environment
Any

### Expected behavior
A nosec comment that works

### Actual behavior
A nosec comment that doesn't work

ccojocar:
Did you try to place the comment on above the if you want to ignore?

paul-at-cybr:
Huh, that worked - Updating the title to be more narrow and added a little bit to the example.

ccojocar:
The nosec typically is ignoring an entire AST node, and not just a line. We did some improvements to be more fine grained, but it seems that this case is not handled.

ravisastryk:
@ccojocar @paul-at-cybr I have added a fix that addresses the issue described here and would appreciate a review of my https://github.com/securego/gosec/pull/1475 Let's improve the tool for the open source community. Thank you for your time.

PR #1467, opened 2026-01-14
Perf: Add global caching and other optimizations

oittaa:
- Implemented thread-safe GlobalCache (LRU) for expensive operations.
- Optimized GetCallInfo with per-package AST node caching.
- Cached GoVersion using sync.Once. -> **Surprisingly profiling showed this to be a huge bottleneck on large code bases.**
- Implemented entropy / secrets pre-filtering.
- Refactored **#nosec** directive parsing to avoid allocations.
- Migrated multiple rules (G101, G111, G114, G306) to use centralized caching.
- Verified 8.6x speedup on CockroachDB SQL package (37s -> 4.3s) and ~600MB memory reduction.

#### Performance Impact (Benchmarked on CockroachDB)
| Metric | Baseline | Optimized | Delta |
| :--- | :--- | :--- | :--- |
| **Execution Time (SQL Pkg)** | 37.0s | **4.29s** | **8.6x Speedup** |
| **Full Root Scan Time** | 47.0s | **12.3s** | **3.8x Speedup** |
| **Peak Memory** | 2.0 GB | **1.4 GB** | **~600 MB Reduction** |

### DIFF
--- a/gosec_cache.go
+++ b/gosec_cache.go
@@ -9,9 +9,16 @@ import (
 // GlobalCache is a shared LRU cache for expensive operations (Regex matching, Entropy analysis).
 var GlobalCache = NewLRUCache[GlobalKey, any](1 << 16)
 
+// Cache kind constants for GlobalKey.Kind.
+const (
+	CacheKindRegex         = iota // Regex match result
+	CacheKindEntropy              // Entropy analysis result
+	CacheKindSecretPattern        // Secret pattern match result
+)
+
 // GlobalKey is a zero-allocation key for the GlobalCache.
 type GlobalKey struct {
-	Kind  int            // 0=Regex, 1=Entropy, 2=SecretPattern
+	Kind  int            // Use CacheKind* constants
 	Regex *regexp.Regexp // Populated for Regex and (optionally) SecretPattern
 	Str   string         // Populated for all
 }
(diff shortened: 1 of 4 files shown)


oittaa:
Now also fixes https://github.com/securego/gosec/issues/1240

oittaa:
I also tried and abandoned some other optimization methods that didn't produce results. Maybe someone in the future is more successful in trying these.
1. Pipeline architecture by separating file reading+load and AST+SSA to their own goroutines. -> It used a bit more memory but runtime didn't improve at all.
2. Tried batch processing packages, but it just ate a whole lot more memory (multiples) and took more time.
3. Caching loaded packages in hope that it would increase performance. There are many common packages that are often included, but at least on my computer the runtime didn't change at all even with very high cache hit ratio. Memory usage just rose significantly (~50%).

oittaa:
Here's another data point. Kubernetes scan time:


Version | Time
-- | --
v2.22.11 | 4min 8.95s
Master | 44.85s 
This | 14.75s

</markdown-accessiblity-table>

oittaa:
Would you prefer if I split this into smaller changes? I just made this branch to investigate where the performance bottlenecks were and which ones were reasonably easy to tackle without going completely crazy. In the end this combination of changes seemed to offer a really nice balance of performance boosts without resorting to anything too weird.

oittaa:
Thanks for the feedback. I've split this into smaller chunks. They should produce exactly the same result combined except I added sync.Pool to AST node caching.

PR #1475, opened 2026-01-19
Fix #1240: nosec comments now work with trailing open brackets

### LINKED ISSUE #1240: #nosec doesn't work if it's preceded by an open bracket {

ravisastryk:
## Summary

Fixes #1240 where `#nosec` comments fail to suppress warnings when placed at the end of lines containing an open bracket `{`.

## Problem

Go's `ast.CommentMap` associates comments with the "largest" node on a line. When a `#nosec` comment appears after an open bracket:

```go
if len(inputSlice) <= int(configVal) { // #nosec G115
```

The comment gets associated with the `IfStmt` instead of the `int(configVal)` expression that triggers the G115 rule, causing the suppression to fail.

## Solution

- Added line-based fallback for nosec detection when `ast.CommentMap` lookup fails
- Added `AllComments` field to `Context` struct for line-based lookup
- Extracted `parseNoSecDirective` helper to reduce code duplication
- Line-based lookup only activates as fallback, preserving existing behavior

## Testing

All existing tests pass.
```
make test
Ran 222 of 222 Specs in 17.057 seconds
SUCCESS! -- 222 Passed | 0 Failed | 0 Pending | 0 Skipped
PASS
```
New tests verify:
- `#nosec` at end of if/for/switch lines with brackets
- Block comments before brackets
- Multiple conversions on same line
- Rule-specific suppression (e.g., `#nosec G115`)

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -144,6 +144,7 @@ func (i ignores) get(file string, line string) map[string][]issue.SuppressionInf
 type Context struct {
 	FileSet      *token.FileSet
 	Comments     ast.CommentMap
+	AllComments  []*ast.CommentGroup // All comments in the file for line-based nosec lookup (fix for #1240)
 	Info         *types.Info
 	Pkg          *types.Package
 	PkgFiles     []*ast.File
@@ -475,6 +476,7 @@ func (gosec *Analyzer) checkRules(pkg *packages.Package) ([]*issue.Issue, *Metri
 			FileSet:      pkg.Fset,
 			Config:       gosec.config,
 			Comments:     ast.NewCommentMap(pkg.Fset, file, file.Comments),
+			AllComments:  file.Comments, // Store all comments for line-based lookup (fix for #1240)
 			Root:         file,
 			Info:         pkg.TypesInfo,
 			Pkg:          pkg.Types,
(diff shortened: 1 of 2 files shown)


ravisastryk:
@ccojocar @paul-at-cybr I have added a fix that addresses the issue described here and would appreciate a review of my https://github.com/securego/gosec/pull/1475 Let's improve the tool for the open source community. Thank you for your time.

oittaa:
Since these regexps are used on basically every comment it had a decent impact on performance. That's why I changed it to manual parsing in https://github.com/securego/gosec/pull/1467 to avoid some allocations and so on.

But at minimum the regexp should always be compiled just once. I've noticed that there are still some of these regexp that might be compiled many times per program execution even though I've already optimized quite a few from rules directory to compile just once. This fix is also a part of the optimizations split out to https://github.com/securego/gosec/pull/1478 but if you see how that could be improved, I'd be really happy to hear especially if that still misses some edge cases.

PR #1477, opened 2026-01-20
feat: ensure GoVersion is cached using sync.Once

oittaa:


oittaa:
Another option would be to put the calculation to init() or something like that, but this initializes lazily.

PR #1478, opened 2026-01-20
fix: optimize nosec directive parsing to avoid allocations

oittaa:
Fixes https://github.com/securego/gosec/issues/1240

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -29,7 +29,6 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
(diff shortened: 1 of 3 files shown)


ccojocar:
Please can you rebase this one? Thanks

oittaa:
Removed all "regexp" dependencies from analyzer.go since they weren't too complicated.
1. Does the string begin `//gosec:disable`
2. Does the string contain G followed by three numbers

Issue #1472, opened 2026-01-18
Add Shannon entropy pre-filter to hardcoded credentials rule?


oittaa:
Would it make sense to add Shannon entropy pre-filter, because it might cause false negatives on "correct horse battery staple"-styled passwords on default settings (cutting at 16 characters), but at least in the real world applications it seems to remove false positives?

Even after #1467 it looked like there might be room for optimizing hardcoded credentials detection. If I understood correctly Zxcvbn punishes leet-coded and other common modifications that simpler password strength calculation algorithms would rank stronger. By adding the following Shannon pre-filtering before Zxcvbn call into `isHighEntropyString()` the scan time of Kubernetes improved another 6.3% and CockroachDB 4.3%. And that's not all, it even dropped several false positives from each. They were basically descriptions, database fields, and so on.

For example this triggers a false positive:
```
cockroach/pkg/cloud/amazon/s3_storage.go
53:       // AWSSecretParam is the query parameter for the 'secret' in an AWS URI.
54:     AWSSecretParam = "AWS_SECRET_ACCESS_KEY"
```

```
For "AWS_SECRET_ACCES" (16 chars):

Zxcvbn Score: 59.49 bits (~3.7 bits/char)
Shannon Score: 45.74 bits (~2.85 bits/char)
The threshold is 3.0 bits/char.

Zxcvbn (3.7) > 3.0 -> Flags it
Shannon (2.85) < 3.0 -> Suppresses it
```

```go
func shannonEntropy(s string) float64 {
	if s == "" {
		return 0
	}
	// Use a stack-allocated array for ASCII characters to avoid heap allocation
	var asciiCounts [256]int
	var unicodeCounts map[rune]int
	length := 0

	for _, char := range s {
		length++
		if char < 256 {
			asciiCounts[char]++
		} else {
			if unicodeCounts == nil {
				unicodeCounts = make(map[rune]int)
			}
			unicodeCounts[char]++
		}
	}

	entropy := 0.0
	flength := float64(length)

	for _, count := range asciiCounts {
		if count > 0 {
			freq := float64(count) / flength
			entropy -= freq * math.Log2(freq)
		}
	}

	for _, count := range unicodeCounts {
		freq := float64(count) / flength
		entropy -= freq * math.Log2(freq)
	}

	return entropy * flength
}
```

ccojocar:
fixed by #1479

PR #1479, opened 2026-01-20
feat: implement entropy pre-filtering to optimize secret detection

oittaa:


oittaa:
I made the educated guess that strings shorter than 8 characters can never hit the default entropy requirements from Zxcvbn without testing too much.

This could be improved further with Shannon entropy filtering before the much more advanced and heavier Zxcvbn. More info: https://github.com/securego/gosec/issues/1472

PR #1480, opened 2026-01-20
feat: implement global cache usage in rules

oittaa:

### DIFF
--- a/gosec_cache.go
+++ b/gosec_cache.go
@@ -2,26 +2,17 @@ package gosec
 
 import (
 	"container/list"
-	"regexp"
 	"sync"
 )
 
-// GlobalCache is a shared LRU cache for expensive operations (Regex matching, Entropy analysis).
-var GlobalCache = NewLRUCache[GlobalKey, any](1 << 16)
-
-// Cache kind constants for GlobalKey.Kind.
-const (
-	CacheKindRegex         = iota // Regex match result
-	CacheKindEntropy              // Entropy analysis result
-	CacheKindSecretPattern        // Secret pattern match result
-)
-
-// GlobalKey is a zero-allocation key for the GlobalCache.
-type GlobalKey struct {
-	Kind  int            // Use CacheKind* constants
-	Regex *regexp.Regexp // Populated for Regex and (optionally) SecretPattern
-	Str   string         // Populated for all
-}
+// GlobalCache is a shared LRU cache for expensive operations.
+// Each use case should define its own named key type to avoid collisions.
+//
+// Key type requirements:
+//   - The key type must be comparable (no slices, maps, or funcs)
+//   - Use type definitions (type MyKey struct{...}), not type aliases (type MyKey = ...)
+//   - Avoid anonymous structs - they collide if the structure matches
+var GlobalCache = NewLRUCache[any, any](1 << 16)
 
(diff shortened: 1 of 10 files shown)


ccojocar:
Please, can you rebase this one? Thanks

gosec_cache.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (c *LRUCache[K, V]) removeOldest() {
+	ent := c.evictList.Back()
+	if ent != nil {
+		c.evictList.Remove(ent)
+		delete(c.items, ent.Value.(*entry[K, V]).key)
+	}
+}
+
+// RegexMatch returns the result of re.MatchString(s), using GlobalCache to store previous results.
```
ccojocar: I would move this function  from here in its own regex utility file.

ccojocar: I would rename it to `RegexMatchWithCache`.

gosec_cache.go
```diff
@@ -0,0 +1,96 @@
...
+	"container/list"
+	"regexp"
+	"sync"
+)
+
+// GlobalCache is a shared LRU cache for expensive operations (Regex matching, Entropy analysis).
+var GlobalCache = NewLRUCache[GlobalKey, any](1 << 16)
+
+// Cache kind constants for GlobalKey.Kind.
+const (
```
ccojocar: These keys don't belong to this package. This should just keep the generic cache implementation. 

Move them where the rule is for each of them is implemented because this is a rule detail. The global cache should hold generic key values without any semantic.

rules/hardcoded_credentials.go
```diff
@@ -205,22 +205,41 @@ func (r *credentials) isHighEntropyString(str string) bool {
 		return false
 	}
 	s := truncate(str, r.truncate)
+	key := gosec.GlobalKey{Kind: gosec.CacheKindEntropy, Str: s}
```
ccojocar: Move the `Kind` in this rule, and make it local instead of exposing it into the entire gosec package. It's only relevant for this rule.

rules/hardcoded_credentials.go
```diff
@@ -205,22 +205,41 @@ func (r *credentials) isHighEntropyString(str string) bool {
...
+type secretResult struct {
+	ok          bool
+	patternName string
 }
 
 func (r *credentials) isSecretPattern(str string) (bool, string) {
 	if len(str) < r.minEntropyLength {
 		return false, ""
 	}
+	key := gosec.GlobalKey{Kind: gosec.CacheKindSecretPattern, Str: str}
```
ccojocar: Move the `Kind` in this rule, and make it local instead of exposing it into the entire gosec package. It's only relevant for this rule.

gosec_cache.go
oittaa: Done

gosec_cache.go
oittaa: Actually a great suggestion. I made the cache file much more generic.

rules/hardcoded_credentials.go
oittaa: Now it's using the Go type system to avoid collisions.

rules/hardcoded_credentials.go
oittaa: Now it's using the Go type system to avoid collisions.

PR #1491, opened 2026-02-06
 Fix G602 analyzer panic that kills gosec process

### LINKED ISSUE #1489: An error occurred during the scanning process using G602, leading to the interruption of analysis and no output results
Using project command analysis, no results can be output
gosec -fmt=json -out=report-gosec-all2.4.2.json -concurrency=4 -exclude-generated  ./...

Image

After using -exclude=G602 for correct analysis, results are output
gosec -fmt=json -out=report-gosec-all2.4.2.json -concurrency=4 -exclude-generated exclude=G602 ./...

ravisastryk:
Fixes #1489  - Running gosec with G602 enabled crashes on certain codebases.

The issue is that the G602 slice-bounds analyzer panics on certain SSA instruction patterns (nil const values, short type strings, nil blocks). Since there is no `recover()` in the analyzer's run path, the panic propagates and terminates the entire gosec process before any results are written.

**Fix:**

- Add `defer recover()` to the analyzer run function - G602 panic degrades to a logged error instead of crashing the process
- Nil-check handling with defensive programming strategy
- Improve code coverage

### DIFF
--- a/analyzers/slice_bounds_test.go
+++ b/analyzers/slice_bounds_test.go
@@ -1,0 +1,108 @@
+// (c) Copyright gosec's authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


ravisastryk:
This PR should follow after the merge of https://github.com/securego/gosec/pull/1492

PR #1520, opened 2026-02-16
Add G120 SSA analyzer for unbounded form parsing in HTTP handlers

ccojocar:
- Introduces new G120 to detect potential memory-exhaustion paths caused by unbounded form parsing in HTTP handlers.
- Uses a pure SSA implementation (no AST fallback), checking ParseForm, ParseMultipartForm, FormValue, and PostFormValue on *http.Request.Suppresses findings when request bodies are explicitly bounded with http.MaxBytesReader.
- Wires G120 into analyzer registration, README rule catalog, and CWE mapping (CWE-400).
- Adds focused vulnerable/safe samples and analyzer test coverage; analyzer tests and lint pass.


SuperSandro2000:
The upstream go doc mentions that ParseForm is capped at 10 MB already 🤔 

https://pkg.go.dev/net/http#Request.ParseForm

Issue #1542, opened 2026-02-23
taint: context.Context propagation from request.Context() causes cascading false positives in G703–G706


richardrigby:
# Summary

The taint engine propagates taint through `context.Context` arguments, causing massive false positives for **G703**/**G704**/**G705**/**G706** in any codebase that follows the standard Go pattern of passing `request.Context()` to downstream function calls (gRPC services, database clients, etc.).

When an HTTP handler calls `request.Context()` on the tainted `*http.Request` parameter, the returned `context.Context` is marked as tainted. Any function that receives this context as an argument then has its entire return value marked as tainted — including error values, response payloads, and every downstream derivation. This creates a cascading chain of false taint that flags every `response.Write()` in the handler.

`context.Context` is a control-flow mechanism (deadlines, cancellation, request-scoped metadata). It does not carry user-controlled input data that could flow to taint sinks like HTTP response bodies, file paths, or SQL queries. The Go idiom of passing `ctx` as the first argument to virtually every I/O function means treating it as a taint vector effectively taints every return value in the call graph reachable from an HTTP handler, which defeats the purpose of taint analysis.

This is related to #1500. PR #1522 improved taint analysis with field-sensitive tracking and additional sanitizers but did not address the context.Context propagation vector.

## Steps to reproduce the behaviour

Any HTTP handler that passes request.Context() to a function call and then writes the result or error to the response will trigger false positives. 

### Minimal example:

```go
import (
      "encoding/json"
      "net/http"


      "google.golang.org/grpc/codes"
      "google.golang.org/grpc/status"
)


type server struct {
      backend BackendClient
}


// BackendClient represents any gRPC or downstream service client.
type BackendClient interface {
      GetData(ctx context.Context, id string) ([]byte, error)
}


func (s *server) Handler(w http.ResponseWriter, r *http.Request) {
      id := "hardcoded-id" // not from user input


      // Pattern 1: gRPC/service call with request.Context()
      data, err := s.backend.GetData(r.Context(), id)
      if err != nil {
              // Pattern 2: error handling — writing gRPC status message
              if st, ok := status.FromError(err); ok {
                      w.WriteHeader(http.StatusBadRequest)
                      msg, _ := json.Marshal(map[string]string{"error": st.Message()})
                      _, _ = w.Write(msg)
                      return
              }
      }


      // Pattern 3: writing response data from downstream call
      _, _ = w.Write(data) // FALSE POSITIVE: G705 flagged here
}
```

The taint chain is:

[...]

Issue #1500, opened 2026-02-12
Is the taint analysis broken? (G703, G704, G705 and G706)


linkdata:
I get a ton of warnings for G703, G704, G705 and G706. So many that I had to disable the warnings globally.

And I think they're all false positives. Here's an example for a G704 that was nicely isolated:

*G704 (CWE): SSRF via taint analysis (Confidence: HIGH, Severity: HIGH)*

```go
func GetPublicIP() (publicip string, err error) {
	ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
	defer cancel()
	var req *http.Request
	if req, err = http.NewRequestWithContext(ctx, http.MethodGet, "https://am.i.mullvad.net/ip", nil); err == nil {
		var resp *http.Response
		if resp, err = http.DefaultClient.Do(req); err == nil { // <--- How is this tainted?
			defer resp.Body.Close()
			if resp.StatusCode == http.StatusOK {
				if body, err := io.ReadAll(resp.Body); err == nil {
					publicip = string(body)
				}
			} else {
				_, _ = io.Copy(io.Discard, resp.Body)
			}
		}
		if i := strings.LastIndexByte(publicip, ':'); i >= 0 {
			publicip = publicip[:i]
		}
		publicip = strings.TrimSpace(publicip)
	}
	return
}
```

ccojocar:
@ravisastryk Please could you have a look at this. It seems to be related to #1497 .

ravisastryk:
Thank you @linkdata for the detailed report. The isssue that the taint engine marks any value as tainted based solely on its type (For example: `*http.Request`), without checking where the data actually comes from. So your `http.NewRequestWithContext` with a hardcoded URL gets treated identically to one carrying user input. The `filepath.WalkDir` case has the same issue.

I have put PR #1506 that addresses this by restricting type-based taint to function parameters from external callers, and also adding proper closure tracing for callbacks like `WalkDir`, and introducing sanitizer support. Both of your examples should produce clean results with that fix applied.

I would appreciate it if you could give it a try on your codebase(with this private branch - PR #1506) and let us know.

linkdata:
Better, but still getting false positives.
The repo in question is https://github.com/linkdata/rinse
Using gosec tip, gives five false positives: 2 x G703, 2 x G705 and a G118.

PR #1543, opened 2026-02-23
taint: skip `context.Context` arguments during taint propagation to fix false positives

richardrigby:
## Summary

- Fix widespread G703/G704/G705/G706 false positives caused by `context.Context` propagating taint from `*http.Request` through to unrelated function return values
- Add `isContextType()` helper and apply it at all four argument-scanning sites in the taint engine
- Harden `isContextType` to unwrap pointer layers (e.g., `*context.Context`) before the named-type check
- Add unit tests for `isContextType` covering positive, pointer, and negative cases
- Add integration test sample verifying the `r.Context()` → service call → `w.Write` pattern produces no false positive

## Problem

The taint engine currently propagates taint through `context.Context` arguments.
In idiomatic Go HTTP handlers, this creates a cascading false positive chain that
renders the taint-based rules (G703–G706) unusable on real-world codebases.

Consider this common pattern:

```go
func (s *server) Handler(w http.ResponseWriter, r *http.Request) {
    resp, err := s.grpcService.DoSomething(
        r.Context(),  // ← tainted: method call on tainted *http.Request
        &pb.Request{UserId: userID},
    )
    if err != nil {
        if st, ok := status.FromError(err); ok {
            w.Write(utils.JSONErrorMessage(st.Message()))  // ← G705 flagged
        }
    }
    w.Write(resp.GetFileData())  // ← G705 flagged
}
```

### The taint chain works as follows:

1. `r *http.Request` — tainted (type source)
2. `r.Context()` — tainted (method call on tainted receiver returns tainted value)
3. `s.grpcService.DoSomething(r.Context(), ...)` — entire return tainted because `r.Context()` is a tainted argument, and for interface/external calls the engine conservatively assumes any tainted argument taints the return
4. `err` from the gRPC call — tainted (extracted from tainted call)
5. `status.FromError(err)` → `st.Message()` — tainted via receiver chain
6. `w.Write(...)` — sink reached with tainted data → G705 reported

In a real codebase with this pattern, this produces 137 false positives across 14 files — every single gRPC call site in every HTTP handler, with zero true positives. This is consistent with the category of false positives reported in #1500.

### Why `context.Context` should not propagate taint

`context.Context` is a control-flow mechanism, not a data-carrying type relevant to taint sinks:

- It carries deadlines, cancellation signals, and request-scoped metadata (trace IDs, auth tokens for internal service-to-service use)
- It does not carry user-controlled input data that could flow to HTTP response bodies, file paths, SQL queries, or other taint sinks
- The Go standard library and ecosystem idiom is to pass `ctx` as the first argument to virtually every function that does I/O — treating it as a taint vector means every function that accepts a context and returns data gets its return value tainted, which defeats the purpose of taint analysis

Even `context.WithValue` stores metadata (trace IDs, auth info) that is not user-controlled input. A theoretical scenario where user input is stored in a context value and later written unsanitized to a response is exotic enough that it should not penalize the overwhelmingly common case.

### `isContextType` helper (`taint/taint.go`)

Add an `isContextType(types.Type) bool` helper that:
1. Unwraps any pointer layers (handles `*context.Context` as well as `context.Context`)
2. Checks the resulting named type's package path (`"context"`) and name (`"Context"`)

```go
func isContextType(t types.Type) bool {
    for {
        ptr, ok := t.(*types.Pointer)
        if !ok {
            break
        }
        t = ptr.Elem()
    }
    named, ok := t.(*types.Named)
    if !ok {
        return false
    }
    obj := named.Obj()
    return obj != nil && obj.Pkg() != nil &&
        obj.Pkg().Path() == "context" && obj.Name() == "Context"
}
```

### Applied at all four taint-propagation sites in `isTainted` / `doTaintedArgsFlowToReturn`:

1. Interface method calls (`val.Call.IsInvoke()`) — e.g., gRPC service interface calls like `s.service.Method(ctx, req)`
2. External static method calls (no SSA body, with receiver) — external library methods
3. External plain function calls (no SSA body, no receiver) — e.g., `status.FromError(err)` where the function body isn't available
4. `doTaintedArgsFlowToReturn` — the interprocedural analysis that checks whether tainted arguments in internal functions flow to return statements

The receiver taint propagation (e.g., `req.URL.Query().Get("name")`) is not affected — this fix only skips `context.Context` when scanning non-receiver arguments for taint propagation to return values.

### Unit tests (`taint/analyzer_internal_test.go`)

Three new tests for `isContextType`:

- `TestIsContextTypeWithContextContext` — `context.Context` (named interface) returns `true`
- `TestIsContextTypeWithPointerToContextContext` — `*context.Context` (pointer-wrapped) returns `true`
- `TestIsContextTypeRejectsNonContextTypes` — table-driven: `http.Request`, `string`, wrong-package-same-name, right-package-wrong-name, pointer to non-context type, `nil` all return `false`

### Integration test (`testutils/g705_samples.go`)

New `SampleCodeG705` entry (expected issues: 0) exercises the pattern from this PR:

```go
type service struct{}

func (s *service) GetData(ctx context.Context, id string) ([]byte, error) {
    return []byte("safe data"), nil
}

func handler(w http.ResponseWriter, r *http.Request) {
    svc := &service{}
    data, _ := svc.GetData(r.Context(), "static-id")
    w.Write(data)
}
```

This is exercised by the existing G705 test in `analyzers/analyzers_test.go`.

## Test plan

- All 24 gosec packages pass (`go test ./...`)
- Taint package: 87 tests pass
- Analyzers package: 156 Ginkgo specs pass (includes new G705 integration sample)
- Performance benchmark: 12.6M ns/op vs baseline 33.6M / limit 38.6M — within bounds
- Tested against a real-world Go HTTP codebase with 32 files / 18,757 lines using gRPC service calls in HTTP handlers: 137 false positives → 0 issues
- Verified that legitimate XSS detection (e.g., writing `r.URL.Query().Get("x")` directly to `http.ResponseWriter`) is not affected — that taint flows through the receiver chain, not through context arguments

## Related

- Follows up on #1500 (taint analysis false positives) and #1522 (field-sensitive taint + sanitizer improvements)
- #1522 added `json.Marshal`/`strconv` sanitizers and interprocedural analysis but did not address the `context.Context` propagation vector, which is the dominant source of false positives in HTTP+gRPC codebases

ref: #1542

### DIFF
--- a/taint/analyzer_internal_test.go
+++ b/taint/analyzer_internal_test.go
@@ -115,6 +115,20 @@ func TestIsContextTypeWithContextContext(t *testing.T) {
 	}
 }
 
+func TestIsContextTypeWithPointerToContextContext(t *testing.T) {
+	t.Parallel()
+
+	pkg := types.NewPackage("context", "context")
+	iface := types.NewInterfaceType(nil, nil)
+	obj := types.NewTypeName(token.NoPos, pkg, "Context", nil)
+	named := types.NewNamed(obj, iface, nil)
+	ptr := types.NewPointer(named)
+
+	if !isContextType(ptr) {
+		t.Fatalf("expected isContextType to return true for *context.Context")
+	}
+}
+
 func TestIsContextTypeRejectsNonContextTypes(t *testing.T) {
 	t.Parallel()
 
@@ -151,12 +165,11 @@ func TestIsContextTypeRejectsNonContextTypes(t *testing.T) {
 			}(),
 		},
 		{
-			name: "pointer to context.Context",
+			name: "pointer to non-context type",
 			typ: func() types.Type {
-				pkg := types.NewPackage("context", "context")
-				obj := types.NewTypeName(token.NoPos, pkg, "Context", nil)
-				named := types.NewNamed(obj, types.NewInterfaceType(nil, nil), nil)
-				return types.NewPointer(named)
+				pkg := types.NewPackage("net/http", "http")
+				obj := types.NewTypeName(token.NoPos, pkg, "Request", nil)
+				return types.NewPointer(types.NewNamed(obj, types.NewStruct(nil, nil), nil))
 			}(),
 		},
 		{
--- a/taint/taint.go
+++ b/taint/taint.go
@@ -28,6 +28,14 @@ const maxTaintDepth = 50
 // Tainted context arguments (e.g., request.Context()) should not propagate taint
 // to function return values, as the context doesn't flow as data to the output.
 func isContextType(t types.Type) bool {
+	// Unwrap pointer layers (e.g., *context.Context) to reach the named type.
+	for {
+		ptr, ok := t.(*types.Pointer)
+		if !ok {
+			break
+		}
+		t = ptr.Elem()
+	}
 	named, ok := t.(*types.Named)
 	if !ok {
 		return false


samstride:
@ccojocar , any chance we can get a release with this fix?

Issue #1549, opened 2026-02-26
G704: False positive when using const URL


leonklingele:
G704 should not report any issues when a constant URL string is being requested.

```go
package main

import (
	"context"
	"net/http"
)

const (
	url = "https://go.dev/"
)

func main2() {
	ctx := context.Background()

	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, http.NoBody)
	if err != nil {
		panic(err)
	}

	_, err = new(http.Client{}).Do(req)
	if err != nil {
		panic(err)
	}
}
```

```shell
$ golangci-lint version
golangci-lint has version 2.10.1 built with go1.26.0 from v2.10.1 on 1970-01-01T00:00:00Z

$ golangci-lint -v run ./...
main.go:20:32: G704: SSRF via taint analysis (gosec)
	_, err = new(http.Client{}).Do(req)
```

ravisastryk:
Thanks @leonklingele for reporting the issue. It is a confirmed false positive. I'll fix this. The taint engine isn't short-circuiting on `*ssa.Const` values. As a matter of fact, `const` URL is a compile-time literal and can never carry attacker-controlled data, but the engine currently propagates taint through it anyway before it reaches the `http.Client.Do` sink.

PR #1551, opened 2026-02-27
Fix/g704 false positive const url

ravisastryk:
## Problem
G704 (SSRF) fires when a `const` URL string is passed to `http.Client.Do` via an intermediate `*http.Request`. The taint engine was not short-circuiting on `*ssa.Const` values, so compile-time literals were being treated as potentially attacker-controlled.

## Solution
Adds a single early-exit guard in the taint propagation walk: if the SSA value is an `*ssa.Const`, return clean immediately. Constants are compile-time literals and can never carry user input by definition.

This also prevents the same class of false positive across all other G70x rules (G701 SQLi, G703 path traversal, etc.) wherever a constant-valued argument reaches a sink.

Fixes: #1549

### DIFF
--- a/taint/taint.go
+++ b/taint/taint.go
@@ -521,6 +521,12 @@ func (a *Analyzer) isTainted(v ssa.Value, fn *ssa.Function, visited map[ssa.Valu
 	}
 	visited[v] = true
 
+	// Constants are compile-time literals and can never carry attacker-controlled
+	// data. Short-circuit immediately — no taint possible.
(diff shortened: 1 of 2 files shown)


ravisastryk:
@ccojocar Thank you for the review. I have rebased on top of master so we have these specific changes addressing the issue. We should be good now.

leonklingele:
Thanks @ravisastryk for the super quick fix! ❤️

Issue #1553, opened 2026-02-27
G705 false finding when no guard can be resolved


leonnicolas:
This simple test case is raising a G705.
```go
package main

import (
	"fmt"
	"os"
)

func main() {
	fmt.Fprint(os.Stdout, os.Args[1])
}
```

The guard is skipped because the `http.ResponseWriter` type is not resolved.
Does it make sense to not satisfy the guards when all (or even any) guard types cannot be resolved?

When adding the type
```go
package main

import (
	"fmt"
	"os"
	"net/http"
)

func main() {
	_ = http.ResponseWriter(nil)
	fmt.Fprint(os.Stdout, os.Args[1])
}
```
gosec does not find anything.

PR #1554, opened 2026-02-27
fix(G705): eliminate false positive when guard type cannot be resolved

leonnicolas:
A guard cannot be satisfied when its type cannot be resolved.
Otherwise gosec fires for code like:
```go
package main

import (
	"fmt"
	"os"
)

func main() {
	fmt.Fprint(os.Stdout, os.Args[1])
}
```
In this case the guard type `http.ResponseWriter` cannot be resolved, all guards (1)
are satisfied and we have a false gosec warning.

Would fix https://github.com/securego/gosec/issues/1553

### DIFF
--- a/testutils/g705_samples.go
+++ b/testutils/g705_samples.go
@@ -165,11 +165,9 @@ package main
 import (
 	"fmt"
 	"os"
-	"net/http"
 )
 
 func main() {
-	_ = http.ResponseWriter(nil)
 	fmt.Fprint(os.Stdout, os.Args[1])
 }
 `}, 0, gosec.NewConfig()},


testutils/g705_samples.go
```diff
@@ -156,6 +156,22 @@ func main() {
...
+package main
+
+import (
+	"fmt"
+	"os"
+	"net/http"
+)
+
+func main() {
+	_ = http.ResponseWriter(nil)
```
leonnicolas, line 172: i meant to remove this

taint/analyzer_internal_test.go
```diff
@@ -307,11 +307,11 @@ func TestGuardsSatisfiedRequiredTypeNotFound(t *testing.T) {
 	t.Parallel()
 	prog := ssa.NewProgram(token.NewFileSet(), 0)
 	// Guard refers to a type that is not present in the program.
-	// The guard must be skipped (continue), so the function returns true.
+	// The guard must not be satisfied.
```
ravisastryk, line 310: Can you please write an elaborate comment here?

```suggestion
	// When the guard type cannot be resolved (package not imported),
    // the argument cannot implement that type, so the guard must NOT
    // be satisfied.
```

ravisastryk (approved):
Thank you for addressing the issue. Please add additional meaning comment in analyzer_internal_test.go. LGTM

ccojocar (approved):

PR #1578, opened 2026-03-01
Fix G115 false positives for guarded int64-to-byte conversions

### LINKED ISSUE #1577: false positives in G115: checked with `if` prior to int64 -> byte
```go
			// allow signed bytes [-128,127] and unsigned bytes [0,255]
			if v < -128 || v > 255 {
				return s.Errorf("invalid byte value %d", v)
			}
			buf[i] = byte(v)
```
getting
```
extensions/extension.go:559:17: G115: integer overflow conversion int64 -> byte (gosec)
                        buf[i] = byte(v)
                                     ^
```
can't overflow with the if before (and yes negative values if any get mapped as needed)

even if I add
```
			if v < -128 || v > 255 {
				return s.Errorf("invalid byte value %d", v)
			}
			if v < 0 {
				v += 256
			}
			buf[i] = byte(v)
```
it still complains

ccojocar:
fixes #1577

### DIFF
--- a/analyzers/conversion_overflow.go
+++ b/analyzers/conversion_overflow.go
@@ -234,14 +234,14 @@ func signedMinForUnsignedSize(size int) int64 {
 	if size >= 64 {
 		return math.MinInt64
 	}
-	return -int64(uint64(1) << (size - 1))
+	return -(int64(1) << (size - 1))
 }
 
 func signedMaxForUnsignedSize(size int) int64 {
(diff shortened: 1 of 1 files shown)


ldemailly:
@ccojocar thanks for the quick fix, yet it doesn't seem to fix it unless I did something wrong https://github.com/securego/gosec/issues/1577#issuecomment-3981043444

PR #1642, opened 2026-04-23
Allow rune to byte conversions

### LINKED ISSUE #1636: False positive G115: integer overflow conversion rune -> byte
This code:

```go
const maxByte = 255
asRunes := []rune(malformed)
final := make([]byte, len(asRunes))
for i, r := range asRunes {
	if r > maxByte {
		continue
	}
	final[i] = byte(r) // <-- FALSE POSITIVE? G115 here
}
```

produces a false positive on the marked line, I think.

Just above it, we skip that line `if r > maxByte`, and maxByte is `const 255`. So, it should never overflow right?

ccojocar:
fixes #1636

### DIFF
--- a/analyzers/range_analyzer.go
+++ b/analyzers/range_analyzer.go
@@ -494,6 +494,12 @@ func (ra *RangeAnalyzer) isNonNegativeRecursive(v ssa.Value) bool {
 		return true
 	}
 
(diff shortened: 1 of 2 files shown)


mholt:
Nice, thanks!

PR #1656, opened 2026-04-25
G409: detect bcrypt cost below DefaultCost

fetchrn:
Adds G409 to flag bcrypt.GenerateFromPassword calls with cost below 10. I followed G403 since it's the same idea, just pulling the second arg and comparing it to a threshold.

Then I noticed bcrypt.MinCost wasn't getting caught. gosec.GetInt only handles int literals, so I added a ctx.Info.Types fallback. That's the second commit.

Tests pass.

This is my first open source contribution, let me know if anything's off.


rules/bcrypt.go
```diff
@@ -0,0 +1,64 @@
+// (c) Copyright 2016 Hewlett Packard Enterprise Development LP
```
ccojocar, line 1: Can you please update to the current header: https://github.com/securego/gosec/blob/4ead098510926e1015958a36dc966bfcb7f6ee11/analyzers/openredirect.go#L1

rules/bcrypt.go
```diff
@@ -0,0 +1,64 @@
...
+		if cost, exact := constant.Int64Val(tv.Value); exact && cost < int64(w.minCost) {
+			return c.NewIssue(n, w.ID(), w.What, w.Severity, w.Confidence), nil
+		}
+	}
+	return nil, nil
+}
+
+// NewWeakBcryptCost detects bcrypt.GenerateFromPassword with cost below bcrypt.DefaultCost
+func NewWeakBcryptCost(id string, _ gosec.Config) (gosec.Rule, []ast.Node) {
+	const minCost = 10 // bcrypt.DefaultCost
```
ccojocar, line 57: Can you set this as default, and allow setting a value form the configuration? 

See https://github.com/securego/gosec/blob/4ead098510926e1015958a36dc966bfcb7f6ee11/rules/hardcoded_credentials.go#L390

ccojocar (changes requested):

rules/bcrypt.go
ravisastryk, line 57: The default of `10` is bcrypt's DefaultCost from 2011, but OWASP's current Password Storage Cheat Sheet recommends `>= 12` for new applications. 

Besides @ccojocar configurability ask, please also update the What message to cite the threshold dynamically (e.g., "bcrypt cost is below configured minimum (%d); OWASP recommends at least 12") so users understand why, not just what.

That is, just keep `10` as default for compatibility, expose via config, and document the OWASP-12 guidance in [RULES.md](https://github.com/securego/gosec/blob/master/RULES.md)

rules/bcrypt.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if cost, err := gosec.GetInt(costArg); err == nil {
+		if cost < int64(w.minCost) {
+			return c.NewIssue(n, w.ID(), w.What, w.Severity, w.Confidence), nil
+		}
+		return nil, nil
+	}
+
+	// fallback for named constants like bcrypt.MinCost
+	if tv, ok := c.Info.Types[costArg]; ok && tv.Value != nil && tv.Value.Kind() == constant.Int {
```
ravisastryk, line 47: There is a significant false-negative gap here.

It appears this only resolves compile-time constants, but the most common real-world failure mode is dynamic `cost, _ := strconv.Atoi(os.Getenv("BCRYPT_COST"))` defaulting to 0 on parse error, or `cfg.BcryptCost` loaded from misconfigured YAML. Neither is statically resolvable, so `tv.Value` is nil and the rule silently passes. 

Recommendation: when the cost arg can't be statically evaluated, emit a `Low/Low` finding ("bcrypt cost is dynamic; verify >= N at runtime"), and explicitly document this AST-level limitation in [RULES.md](https://github.com/securego/gosec/blob/master/RULES.md) Lifting to SSA is the proper long-term fix.

rules/bcrypt.go
```diff
@@ -0,0 +1,64 @@
...
+		}
+	}
+	return nil, nil
+}
+
+// NewWeakBcryptCost detects bcrypt.GenerateFromPassword with cost below bcrypt.DefaultCost
+func NewWeakBcryptCost(id string, _ gosec.Config) (gosec.Rule, []ast.Node) {
+	const minCost = 10 // bcrypt.DefaultCost
+	rule := &weakBcryptCost{
+		callListRule: newCallListRule(id, fmt.Sprintf("bcrypt cost should be at least %d", minCost), issue.Medium, issue.High),
```
ravisastryk, line 59: I am not sure, if we were to bump severity to `High/High`, or leave a code comment justifying Medium so it's a deliberate decision rather than inherited from from `G403`

ravisastryk (changes requested):
Good first contribution and a sensible template choice in `G403`, but I'd hold this from merging in its current shape.

Three concerns, in order of weight: 
(1) the rule misses the dominant real-world failure mode: costs read from env/config — because both detection paths require compile-time constants; a literal 4 in source is the rare shape of this bug. 

(2) Severity Medium understates `CWE-916`; offline cracking of a full credential table is `High/High` territory, on par with `G401` and `G402`. 

(3) Anchoring the threshold at bcrypt's 2011-era DefaultCost = 10 means we'll bless cost=10 calls that OWASP has considered insufficient for years - @ccojocar  configurability ask is necessary, not optional.

ccojocar:
> (1) the rule misses the dominant real-world failure mode: costs read from env/config — because both detection paths > require compile-time constants; a literal 4 in source is the rare shape of this bug.

The rule detects when the default cost is overwritten to another value which is smaller than the default. I think the current default is `10` which is accepted as fine, maybe `12` would be on the safe side to use. 

> (2) Severity Medium understates CWE-916; offline cracking of a full credential table is High/High territory, on par with G401 and G402.

Medium is fine. I would say is maybe high only when someone overwrites a min values below `4` which is the current default minimum, and can lead to serious security flow in the key derivation. derive iterations are 2^(min_value). 

> (3) Anchoring the threshold at bcrypt's 2011-era DefaultCost = 10 means we'll bless cost=10 calls that OWASP has considered insufficient for years - @ccojocar configurability ask is necessary, not optional.

`10` is the current accepted values as a sweet spot for a secure derivation, across different frameworks and languages:
- Python's bcrypt library: defaults to 12, but allows down to 4
- Node.js bcrypt: default is 10
- Go's golang.org/x/crypto/bcrypt: MinCost = 4, DefaultCost = 10
- Spring Security: default 10
- OWASP recommends a minimum of 10, ideally 12 or higher as of recent guidance

Maybe we can set it to `12` to be on the safe side, but be mindful that the higher this number is more interactions are required. 2^cost is the number of iteration bcrypt uses to derive the key.

ravisastryk:
> > (1) the rule misses the dominant real-world failure mode: costs read from env/config — because both detection paths > require compile-time constants; a literal 4 in source is the rare shape of this bug.
> 
> The rule detects when the default cost is overwritten to another value which is smaller than the default. I think the current default is `10` which is accepted as fine, maybe `12` would be on the safe side to use.
> 
> > (2) Severity Medium understates CWE-916; offline cracking of a full credential table is High/High territory, on par with G401 and G402.
> 
> Medium is fine. I would say is maybe high only when someone overwrites a min values below `4` which is the current default minimum, and can lead to serious security flow in the key derivation. derive iterations are 2^(min_value).
> 
> > (3) Anchoring the threshold at bcrypt's 2011-era DefaultCost = 10 means we'll bless cost=10 calls that OWASP has considered insufficient for years - @ccojocar configurability ask is necessary, not optional.
> 
> `10` is the current accepted values as a sweet spot for a secure derivation, across different frameworks and languages:
> 
> * Python's bcrypt library: defaults to 12, but allows down to 4
> * Node.js bcrypt: default is 10
> * Go's golang.org/x/crypto/bcrypt: MinCost = 4, DefaultCost = 10
> * Spring Security: default 10
> * OWASP recommends a minimum of 10, ideally 12 or higher as of recent guidance
> 
> Maybe we can set it to `12` to be on the safe side, but be mindful that the higher this number is more interactions are required. 2^cost is the number of iteration bcrypt uses to derive the key.



> > (1) the rule misses the dominant real-world failure mode: costs read from env/config — because both detection paths > require compile-time constants; a literal 4 in source is the rare shape of this bug.
> 
> The rule detects when the default cost is overwritten to another value which is smaller than the default. I think the current default is `10` which is accepted as fine, maybe `12` would be on the safe side to use.
> 
> > (2) Severity Medium understates CWE-916; offline cracking of a full credential table is High/High territory, on par with G401 and G402.
> 
> Medium is fine. I would say is maybe high only when someone overwrites a min values below `4` which is the current default minimum, and can lead to serious security flow in the key derivation. derive iterations are 2^(min_value).
> 
> > (3) Anchoring the threshold at bcrypt's 2011-era DefaultCost = 10 means we'll bless cost=10 calls that OWASP has considered insufficient for years - @ccojocar configurability ask is necessary, not optional.
> 
> `10` is the current accepted values as a sweet spot for a secure derivation, across different frameworks and languages:
> 
> * Python's bcrypt library: defaults to 12, but allows down to 4
> * Node.js bcrypt: default is 10
> * Go's golang.org/x/crypto/bcrypt: MinCost = 4, DefaultCost = 10
> * Spring Security: default 10
> * OWASP recommends a minimum of 10, ideally 12 or higher as of recent guidance
> 
> Maybe we can set it to `12` to be on the safe side, but be mindful that the higher this number is more interactions are required. 2^cost is the number of iteration bcrypt uses to derive the key.

Agree.

ccojocar:
Hi @fetchrn, Are you planning to address the review comments that we can merge your pull request? Thanks

ccojocar:
Closing this due to inactivity.

PR #1663, opened 2026-05-01
Feat/exclude values

chaimleib:
Adds `.G101.keys` and `.G101.values` fields to the PathExcludeRule type. If provided with a slice of regex patterns, G101 "Possible hardcoded credentials" issues will also be excluded if the hardcoded-value matches any of the `values` patterns, or if its key matches any of the `keys` patterns.

For example, the following code, common in test files, used to trigger an error:

```go
const token = "fakeToken"
```

With this PR, a config can be set to exclude that issue when encountered in test files, based on the hardcoded value matching a pattern:

```json
{
  "exclude-rules": [{
    "path": ".+_test\\.go$",
    "G101": {
      "values": [
        "(?i)^test",
        "(?i)^fake"
      ]
    }
  }]
}
```

Variable names can be excluded as well, even if the value looks like a credential:

```go
testPassword := "|0oK5-l1k3_a=Pa5sw0Rd"
```

```json
{
  "exclude-rules": [{
    "path": ".+_test\\.go$",
    "G101": {
      "keys": ["(?i)^test"]
    }
  }]
}
```

### DIFF
--- a/exclusion_filter.go
+++ b/exclusion_filter.go
@@ -8,33 +8,33 @@ import (
 	"github.com/securego/gosec/v2/issue"
 )
 
-// PathExcludeRule defines rules to exclude for specific file paths
-type PathExcludeRule struct {
+// ExcludeRule defines rules to exclude for specific file paths
+type ExcludeRule struct {
 	Path  string   `json:"path"`  // Regex pattern for matching file paths
 	Rules []string `json:"rules"` // Rule IDs to exclude. Use "*" to exclude all rules
 }
 
-// compiledPathRule is a pre-compiled version of PathExcludeRule for efficient matching
-type compiledPathRule struct {
+// compiledExcludeRule is a pre-compiled version of PathExcludeRule for efficient matching
+type compiledExcludeRule struct {
 	pathRegex  *regexp.Regexp
 	ruleSet    map[string]bool // Set of rule IDs to exclude
 	excludeAll bool            // True if "*" was specified in rules
-	original   PathExcludeRule // Keep original for error messages
+	original   ExcludeRule     // Keep original for error messages
 }
 
-// PathExclusionFilter handles filtering of issues based on path and rule combinations
-type PathExclusionFilter struct {
-	rules []compiledPathRule
+// ExclusionFilter handles filtering of issues based on path and rule combinations
+type ExclusionFilter struct {
+	rules []compiledExcludeRule
 }
 
-// NewPathExclusionFilter creates a new filter from the provided exclusion rules.
+// NewExclusionFilter creates a new filter from the provided exclusion rules.
 // Returns an error if any path regex is invalid.
-func NewPathExclusionFilter(rules []PathExcludeRule) (*PathExclusionFilter, error) {
+func NewExclusionFilter(rules []ExcludeRule) (*ExclusionFilter, error) {
 	if len(rules) == 0 {
-		return &PathExclusionFilter{rules: nil}, nil
+		return &ExclusionFilter{rules: nil}, nil
 	}
 
-	compiled := make([]compiledPathRule, 0, len(rules))
+	compiled := make([]compiledExcludeRule, 0, len(rules))
 
 	for i, rule := range rules {
 		if rule.Path == "" {
@@ -58,7 +58,7 @@ func NewPathExclusionFilter(rules []PathExcludeRule) (*PathExclusionFilter, erro
 			}
 		}
 
-		compiled = append(compiled, compiledPathRule{
+		compiled = append(compiled, compiledExcludeRule{
 			pathRegex:  regex,
 			ruleSet:    ruleSet,
 			excludeAll: excludeAll,
@@ -66,12 +66,12 @@ func NewPathExclusionFilter(rules []PathExcludeRule) (*PathExclusionFilter, erro
 		})
 	}
 
-	return &PathExclusionFilter{rules: compiled}, nil
+	return &ExclusionFilter{rules: compiled}, nil
 }
 
 // ShouldExclude returns true if the given issue should be excluded based on
 // its file path and rule ID
-func (f *PathExclusionFilter) ShouldExclude(filePath, ruleID string) bool {
+func (f *ExclusionFilter) ShouldExclude(filePath, ruleID string) bool {
 	if f == nil || len(f.rules) == 0 {
 		return false
 	}
@@ -95,7 +95,7 @@ func (f *PathExclusionFilter) ShouldExclude(filePath, ruleID string) bool {
 
 // FilterIssues applies path-based exclusions to a slice of issues.
 // Returns the filtered issues and the count of excluded issues.
-func (f *PathExclusionFilter) FilterIssues(issues []*issue.Issue) ([]*issue.Issue, int) {
+func (f *ExclusionFilter) FilterIssues(issues []*issue.Issue) ([]*issue.Issue, int) {
 	if f == nil || len(f.rules) == 0 || len(issues) == 0 {
 		return issues, 0
 	}
@@ -117,12 +117,12 @@ func (f *PathExclusionFilter) FilterIssues(issues []*issue.Issue) ([]*issue.Issu
 // ParseCLIExcludeRules parses the CLI format for exclude-rules.
 // Format: "path:rule1,rule2;path2:rule3,rule4"
 // Example: "cmd/.*:G204,G304;test/.*:G101"
-func ParseCLIExcludeRules(input string) ([]PathExcludeRule, error) {
+func ParseCLIExcludeRules(input string) ([]ExcludeRule, error) {
 	if input == "" {
 		return nil, nil
 	}
 
-	var rules []PathExcludeRule
+	var rules []ExcludeRule
 
 	// Split by semicolon for multiple rules
 	parts := strings.Split(input, ";")
@@ -164,7 +164,7 @@ func ParseCLIExcludeRules(input string) ([]PathExcludeRule, error) {
 			return nil, fmt.Errorf("exclude-rules part %d: no valid rules specified", i+1)
 		}
 
-		rules = append(rules, PathExcludeRule{
+		rules = append(rules, ExcludeRule{
 			Path:  pathPattern,
 			Rules: cleanedRules,
 		})
@@ -175,7 +175,7 @@ func ParseCLIExcludeRules(input string) ([]PathExcludeRule, error) {
 
 // MergeExcludeRules combines exclude rules from multiple sources (config file + CLI).
 // CLI rules take precedence and are processed first.
-func MergeExcludeRules(configRules, cliRules []PathExcludeRule) []PathExcludeRule {
+func MergeExcludeRules(configRules, cliRules []ExcludeRule) []ExcludeRule {
 	if len(cliRules) == 0 {
 		return configRules
 	}
@@ -184,16 +184,16 @@ func MergeExcludeRules(configRules, cliRules []PathExcludeRule) []PathExcludeRul
 	}
 
 	// CLI rules first, then config rules
-	merged := make([]PathExcludeRule, 0, len(cliRules)+len(configRules))
+	merged := make([]ExcludeRule, 0, len(cliRules)+len(configRules))
 	merged = append(merged, cliRules...)
 	merged = append(merged, configRules...)
 	return merged
 }
 
 // String returns a human-readable representation of the filter
-func (f *PathExclusionFilter) String() string {
+func (f *ExclusionFilter) String() string {
 	if f == nil || len(f.rules) == 0 {
-		return "PathExclusionFilter{empty}"
+		return "ExclusionFilter{empty}"
 	}
 
 	var parts []string
@@ -209,5 +209,5 @@ func (f *PathExclusionFilter) String() string {
 		}
 	}
 
-	return fmt.Sprintf("PathExclusionFilter{%s}", strings.Join(parts, "; "))
+	return fmt.Sprintf("ExclusionFilter{%s}", strings.Join(parts, "; "))
 }
--- a/exclusion_filter.go
+++ b/exclusion_filter.go
@@ -19,12 +19,13 @@ type ExcludeRule struct {
 
 // compiledExcludeRule is a pre-compiled version of PathExcludeRule for efficient matching
 type compiledExcludeRule struct {
-	pathRegex    *regexp.Regexp
-	keyRegexes   []*regexp.Regexp
-	valueRegexes []*regexp.Regexp
-	ruleSet      map[string]bool // Set of rule IDs to exclude
-	excludeAll   bool            // True if "*" was specified in rules
-	original     ExcludeRule     // Keep original for error messages
+	pathRegex         *regexp.Regexp
+	keyRegexes        []*regexp.Regexp
+	valueRegexes      []*regexp.Regexp
+	ruleSet           map[string]bool // Set of rule IDs to exclude
+	excludeAll        bool            // True if "*" was specified in rules
+	noAddendaExcludes bool            // True if no keyRegexes or valueRegexes
+	original          ExcludeRule     // Keep original for error messages
 }
 
 // ExclusionFilter handles filtering of issues based on path and rule combinations
@@ -97,12 +98,14 @@ func NewExclusionFilter(rules []ExcludeRule) (*ExclusionFilter, error) {
 			return nil, fmt.Errorf("exclude-rules[%d].values: %w", i, err)
 		}
 
+		noAddendaExcludes := len(keyRegexes) == 0 && len(valueRegexes) == 0
+
 		ruleSet := make(map[string]bool)
 		excludeAll := false
 
 		for _, ruleID := range rule.Rules {
 			ruleID = strings.TrimSpace(ruleID)
-			if ruleID == "*" {
+			if ruleID == "*" && noAddendaExcludes {
 				excludeAll = true
 			} else if ruleID != "" {
 				ruleSet[ruleID] = true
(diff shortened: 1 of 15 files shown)


ccojocar:
Thanks for this contribution. This seems AI generated. 

Why was required to rename the methods/functions? Can you reduce the size of the changes only to the hardcode credentials?

chaimleib:
> Thanks for this contribution. This seems AI generated.
> 
> Why was required to rename the methods/functions? Can you reduce the size of the changes only to the hardcode credentials?

This was not AI generated. I did not even use AI for autocomplete.

I renamed some symbols because they mentioned "Path Exclusion", but I was adding functionality not related to paths. To avoid the misleading implication that PathExclusion symbols dealt only with paths, I removed "Path" from the names.

This is still WIP, and I'm looking into whether I should push the config deeper to be more tightly coupled to G101. I also want to write tests.

EDIT 2026-05-07: Since I really want to restrict the exclusions to particular paths, path_filters.go is the most natural place to trigger the key-value excluder. But I separated the logic out to its own files as much as I could.

ccojocar:
Please do first the minimum change required to incorporate the functionality you mentioned without refactoring. You can follow up with refactoring if it is needed afterwards. This will make reviewing the change much easier. Thanks

exclusion_filter.go
```diff
@@ -0,0 +1,308 @@
...
+
+// ExcludesAddenda returns whether an issue containing addenda
+// should be excluded from the results, assuming all of the following:
+// - The issue's path matches the exclude rule.
+// - The issue's ruleID matches the exclude rule.
+//
+// In other words, this method will further limit the exlude rule's scope,
+// if the issue has addenda matching either keys or values
+// which were specified in the rule.
+func (r *compiledExcludeRule) ExcludesAddenda(addenda any) bool {
```
ccojocar: I would use a type for `addenda` instead of `any`. It can be a join between Keyer and Valuer.

exclusion_filter.go
```diff
@@ -0,0 +1,308 @@
...
+			}
+		}
+	}
+
+	return false
+}
+
+// ShouldExclude returns true if the given issue should be excluded based on
+// its file path, rule ID, and addenda.
+func (f *ExclusionFilter) ShouldExclude(filePath, ruleID string, addenda any) bool {
```
ccojocar: ditto: `addenda` with type.

exclusion_filter.go
```diff
@@ -0,0 +1,308 @@
...
+		}
+
+		// Split by colon to separate path and rules
+		colonIdx := strings.LastIndex(part, ":")
+		if colonIdx == -1 {
+			return nil, fmt.Errorf("exclude-rules part %d: missing ':' separator in %q", i+1, part)
+		}
+
+		pathPattern := strings.TrimSpace(part[:colonIdx])
+		rulesPart := strings.TrimSpace(part[colonIdx+1:])
```
ccojocar, line 159: Does the part can be only of `len` == `1`? This indexing will generate a crash in that case.

ccojocar (commented):
Can you please add some info in the README regarding the usage of this exclusion filter?

Is this introducing any breaking changes for the existing filter format?

exclusion_filter.go
chaimleib: This is not my code. It was carried over from when file had a different name.

exclusion_filter.go
chaimleib: The idea (which I'm reconsidering) was to allow the different rules to provide different types of addenda. I imagined that some rules might only provide a key, for example, and not a value. And some rules might provide  addenda with completely different methods, or none at all.

This would go away if I tightly couple the new configs to G101.

rules/hardcoded_credentials.go
```diff
@@ -239,31 +239,52 @@ func (r *credentials) Match(n ast.Node, ctx *gosec.Context) (*issue.Issue, error
 	return nil, nil
 }
 
+func (r *credentials) issueForMatchedKey(n ast.Node, lhs string, rhs ast.Expr, ctx *gosec.Context) *issue.Issue {
+	val, err := gosec.GetString(rhs)
+	if err != nil {
+		return nil
```
ccojocar, line 248: Why is this error ignored? I would return it to the caller.

rules/hardcoded_credentials.go
```diff
@@ -239,31 +239,52 @@ func (r *credentials) Match(n ast.Node, ctx *gosec.Context) (*issue.Issue, error
...
+		iss := ctx.NewIssue(n, r.ID(), r.What, r.Severity, r.Confidence)
+		return iss
+	}
+	return nil
+}
+
+func (r *credentials) issueIfValueInSecretFormat(n ast.Node, lhs string, rhs ast.Expr, ctx *gosec.Context) *issue.Issue {
+	val, err := gosec.GetString(rhs)
+	if err != nil {
+		return nil
```
ccojocar, line 267: ditto

path_filter.go
```diff
@@ -27,6 +28,37 @@ type PathExclusionFilter struct {
 	rules []compiledPathRule
 }
 
+// CompileRegexes returns a slice of compiled regular expressions.
+// Returns nil if an empty patterns slice is provided.
+// Returns an error if any pattern is empty or failed to compile.
+func CompileRegexes(patterns []string) ([]*regexp.Regexp, error) {
```
ccojocar: I don't see where is this function used?

ccojocar (commented):
Also make sure to update the https://github.com/securego/gosec/blob/master/RULES.md#g101 with the added functionality when is implemented.

rules/hardcoded_credentials.go
chaimleib: I added an explanatory comment.

rules/hardcoded_credentials.go
chaimleib: Added an explanatory comment.

path_filter.go
chaimleib: It gets used in hardcoded_credentials_exclude_options.go, to compile the keys and values configs.

chaimleib:
> Can you please add some info in the README regarding the usage of this exclusion filter?
> 
> Is this introducing any breaking changes for the existing filter format?

Done.

There should be no breaking changes for the existing filter format. The new feature is only configurable via JSON right now, via a new field in the path filter config. I weighed allowing configs via CLI arguments, but that seemed too complicated at this stage.

chaimleib:
> Also make sure to update the https://github.com/securego/gosec/blob/master/RULES.md#g101 with the added functionality when is implemented.

Done!

issue/issue.go
```diff
@@ -119,20 +119,30 @@ var ruleToCWE = map[string]string{
...
+	File                       string                      `json:"file"`                                   // File name we found it in
+	Code                       string                      `json:"code"`                                   // Impacted code line
+	Line                       string                      `json:"line"`                                   // Line number in file
+	Col                        string                      `json:"column"`                                 // Column number in line
+	NoSec                      bool                        `json:"nosec"`                                  // true if the issue is nosec
+	Suppressions               []SuppressionInfo           `json:"suppressions"`                           // Suppression info of the issue
+	Autofix                    string                      `json:"autofix,omitempty"`                      // Proposed auto fix the issue
+	HardcodedCredentialAddenda *HardcodedCredentialAddenda `json:"hardcoded_credential_addenda,omitempty"` // Info for configurable Hardcoded Credential exclusion rules.
```
ccojocar, line 145: This file is generic. The name sounds very specific only for hardcoded credentials rule. 

Why this information is required to be added into an issue since the goal of the issue is only to add dictional filters on the matching.

Please remove this from the Issue structure. Thanks

rules/hardcoded_credentials.go
```diff
@@ -239,31 +239,68 @@ func (r *credentials) Match(n ast.Node, ctx *gosec.Context) (*issue.Issue, error
...
+	val, err := gosec.GetString(rhs)
+	if err != nil {
+		// There is no string value to be retrieved from rhs.
+		// This means the rhs is not a string (maybe a var, an expr, etc.).
+		// So, it isn't hardcoded and the node should be skipped silently.
+		return nil
+	}
+	if r.ignoreEntropy || r.isHighEntropyString(val) {
+		iss := ctx.NewIssue(n, r.ID(), r.What, r.Severity, r.Confidence)
+		iss.HardcodedCredentialAddenda = &issue.HardcodedCredentialAddenda{
```
ccojocar, line 252: What's the value of this?

path_filter.go
```diff
@@ -10,16 +10,18 @@ import (
 
 // PathExcludeRule defines rules to exclude for specific file paths
 type PathExcludeRule struct {
-	Path  string   `json:"path"`  // Regex pattern for matching file paths
-	Rules []string `json:"rules"` // Rule IDs to exclude. Use "*" to exclude all rules
+	Path  string                            `json:"path"`  // Regex pattern for matching file paths
+	Rules []string                          `json:"rules"` // Rule IDs to exclude. Use "*" to exclude all rules
+	G101  HardcodedCredentialExcludeOptions `json:"G101"`  // Per-path G101 options
```
ccojocar, line 15: This is generic code. Why `G101` is here? This should be named in a generic way, not specifically to a rule.

hardcoded_credential_exclude_options.go
```diff
@@ -0,0 +1,124 @@
...
+import (
+	"errors"
+	"fmt"
+	"regexp"
+)
+
+// HardcodedCredentialExcludeOptions configures whether
+// a Hardcoded Credentials issue should be excluded
+// from the results, based on options specific to this rule.
+type HardcodedCredentialExcludeOptions struct {
```
ccojocar, line 12: This should be named in a generic way, it's not only meant to be used only by hardcoded credentials rule.

hardcoded_credential_exclude_options.go
```diff
@@ -0,0 +1,124 @@
...
+	Keys []string `json:"keys"`
+
+	// Keys holds a list of Go regexp patterns.
+	// If on a Hardcoded Credentials issue any pattern matches a hardcoded value,
+	// the issue will be excluded from the report.
+	// If empty, we behave as if no exclude value pattern matched.
+	Values []string `json:"values"`
+}
+
+type CompiledHardcodedCredentialsRule struct {
```
ccojocar, line 27: ditto

path_filter.go
```diff
@@ -10,16 +10,18 @@ import (
...
 
 // compiledPathRule is a pre-compiled version of PathExcludeRule for efficient matching
 type compiledPathRule struct {
-	pathRegex  *regexp.Regexp
-	ruleSet    map[string]bool // Set of rule IDs to exclude
-	excludeAll bool            // True if "*" was specified in rules
-	original   PathExcludeRule // Keep original for error messages
+	pathRegex                    *regexp.Regexp
+	ruleSet                      map[string]bool // Set of rule IDs to exclude
+	hardcodedCredentialsExcluder *CompiledHardcodedCredentialsRule
```
ccojocar, line 22: ditto, this should be named generically maybe `CredentialsRule`.

ccojocar (changes requested):

issue/issue.go
chaimleib: In [one of my previous commits](https://github.com/securego/gosec/pull/1663/changes/BASE..8f6972a38a5c8d862526ce41135f4a3ed910f159#diff-78cb896736d386aba49a1e1f0edb2d2b36f0ea8a28051dac625f7bf58d4242c7R136), this field was `Addenda any`. The problem then was the use of `any`. I [got a comment](https://github.com/securego/gosec/pull/1663#discussion_r3195211432) seeming to say that it was too generic then. Should I change this back to `Addenda any` so that it becomes generic again? EDIT: The way I was envisioning things at that commit, different rules might add different Addenda fulfilling different interfaces than `Keyer` and `Valuer`, which I have since removed. Should I go back to that plan with `Addenda any`? Or should I stay with what I have, and refrain from imposing an opinion on later features which are not fleshed out yet?

`(*PathExclusionFilter).ShouldExclude(*issue.Issue)` needs some sort of addenda field, because otherwise it would lack key-value information about the violation. It wouldn't be able to call my new `rule.hardcodedCredentialsExcluder.ShouldExcludeKV`.

If I removed the call to `ShouldExcludeKV` from `ShouldExclude`, and instead made the exclusions at the rule level in `rules/hardcoded_credentials.go`, then the `excluded` counter for path-based excludes would fail to increment. In concrete terms, this section in `cmd/gosec/main.go` would not count the new exclusions:

```go
	// Apply path-based exclusions first
	var pathExcludedCount int
	issues, pathExcludedCount = pathFilter.FilterIssues(issues)
	if pathExcludedCount > 0 {
		logger.Printf("Excluded %d issues by path-based rules", pathExcludedCount)
	}
```

Do we want that?

path_filter.go
chaimleib: In the JSON configuration file, users are expected to know the rule codes, as in the `Rules` field just above. As a UI, I tried to stay consistent in how users will name rule-specific options, so I named the rule-specific path exclude options with `json:"G101"`.

(I did not want to change the type of the `Rules` field to a `map[string]any`. That would break compatibility with previous config files.)

I was also trying not to be confusing by naming the field in Go too differently from the JSON name. That's why the field name is also `G101`. Would you prefer the following?

```go
	CredentialExcludeOptions HardcodedCredentialExcludeOptions `json:"G101"`
```

rules/hardcoded_credentials.go
chaimleib: This ensures that all issues reported by this rule include key-value information in a consistent way.

hardcoded_credential_exclude_options.go
chaimleib: I thought of opening this up to allow other rules to use this. But when I looked at the other rules, they were different enough that either these options were irrelevant, or would increase the PR size by a lot.

Did you find another rule where you want to use this type?

To minimize the PR impact, and to keep type ownership and usage clear, I decided to not imply usefulness beyond "hardcoded credentials" (the name for the rule implied by the filename `rules/hardcoded_credentials.go`).

If a later PR does want to use this for a different rule, that PR can rename the types as needed.

rules/hardcoded_credentials.go
ccojocar: In which context is the key/value information required in addition to the issue itself?

path_filter.go
chaimleib: I like shorter names. Should this be like so?

```go
	credentailsRule *CompiledCredentialsRule
```

path_filter.go
ccojocar: My point is that this is generic code, not rule specific. You should use generic names for variables, and it should be written in a generic way that it can be reused.

For example this should be:
```
ExcludeOptions ExcludeOptions 
```

hardcoded_credential_exclude_options.go
ccojocar: See my comment above, use a generic name for the variables.

rules/hardcoded_credentials.go
ccojocar: I don't see what's the benefit of having this information into an issue.

chaimleib: We could avoid this, if path-based exclusions for this rule were processed in `rules/hardcoded_credentials.go`, rather than in `path_filter.go`. 

However, without some data passed down in the `Issue`, that [would mess up](https://github.com/securego/gosec/pull/1663#discussion_r3221218474) the `pathExcludedCount` statistics.

The existing logic in `path_filter.go` only has access to the configs and the issue, so that's why I injected the key and value there. I thought it was important for there to be a single place, `path_filter.go`, which computes path-based exclusions, and tallies them all up.

But maybe that is not so important. I could compute the path exclusion in the rule, and then have it signal in the issue that it was excluded by the `exclude-rules.G101` config:

-  A new `ExcludedByPath bool` field in the Issue struct could work. That's generic.
- Or, could I use the existing `Suppressions` field? I didn't use that, because I thought this was reserved for `#nosec` and `nolint:gosec` directives, which include a `Justification` string. If it's ok to use that field, maybe I should add a `justification` field to the new config file format?

```json
{
  "exclude-rules": [{
    "path": ".+_test\\.go$",
    "G101": {
      "justification": "Mocked credentials for tests are not real.",
      "values": [
        "(?i)^test",
        "(?i)^fake"
      ]
    }
  }]
}
```

path_filter.go
chaimleib: Right now, in the `master` branch, this is the best config I can manage for my purpose:

```json5
/* Config */ {
  "exclude-rules": /* []PathExcludeRule */ [{
    "path": ".+_test\\.go$",
    "rules": [ "G101" ]
  }]
}
```

My idea was to allow being more particular about the exclusions like this:
```json5
/* Config */ {
  "exclude-rules": /* []PathExcludeRule */ [{
    "path": ".+_test\\.go$",
    "G101": /* HardcodedCredentialExcludeOptions */ {
      "keys": [ "(?i)^test", "(?i)^fake" ],
      "values": [ "(?i)^test", "(?i)^fake", "\*{4}" ]
    }
  }]
}
```

Are you proposing this?
```json5
/* Config */ {
  "exclude-rules": /* []PathExcludeRule */ [{
    "path": ".+_test\\.go$",
    "options": /* ExcludeOptions */ {
      "G101": /* HardcodedCredentialExcludeOptions */ {
        "keys": [ "(?i)^test", "(?i)^fake" ],
        "values": [ "(?i)^test", "(?i)^fake", "\*{4}" ]
      }
    }
  }]
}
```
Isn't that more awkward than my idea?

chaimleib: Or, did you mean this?

```json5
/* Config */ {
  "exclude-rules": /* []PathExcludeRule */ [{
    "path": ".+_test\\.go$",
    "options": /* ExcludeOptions */ {
      "keys": [ "(?i)^test", "(?i)^fake" ],
      "values": [ "(?i)^test", "(?i)^fake", "\*{4}" ]
    }
  }]
}
```

What other rules would benefit from this?

What if you don't want all the rules to use a shared config? There would be no way to provide different rules with separate configs, unless each config was named uniquely to that rule:

```json5
/* Config */ {
  "exclude-rules": /* []PathExcludeRule */ [{
    "path": ".+_test\\.go$",
    "options": /* ExcludeOptions */ {
      "G101_keys": [ "(?i)^test", "(?i)^fake" ],
      "G101_values": [ "(?i)^test", "(?i)^fake", "\*{4}" ],
      "G999_keys": [ "(?i)^test", "(?i)^fake" ],
      "G999_values": [ "(?i)^test", "(?i)^fake", "\*{4}" ]
    }
  }]
}
```

But that would also get awkward and repetitive. Isn't it better to make a separate object for the configs of each rule?

```json5
/* Config */ {
  "exclude-rules": /* []PathExcludeRule */ [{
    "path": ".+_test\\.go$",
    "G101": /* HardcodedCredentialExcludeOptions */ {
      "keys": [ "(?i)^test", "(?i)^fake" ],
      "values": [ "(?i)^test", "(?i)^fake", "\*{4}" ]
    },
    "G999": /* TBDExcludeOptions */ {
      "keys": [ "(?i)^test", "(?i)^fake" ],
      "values": [ "(?i)^test", "(?i)^fake", "\*{4}" ]
    }
  }]
}
```

PR #1671, opened 2026-05-11
Add two options to require rule ID and justification for inline annotations

### LINKED ISSUE #1662: RFE: options to require rules and justfication for inline annotations
It would be nice if gosec had options to require specifying
- at least one rule to inline `#nosec`/`gosec:disable` annotations (i.e. ban "naked" directives without any rules), in order to avoid inadvertently suppressing other than intended rules that might apply
- a justification present in all `#nosec`/`gosec:disable` annotations, for rationale as it is not always obvious, and this is quite probably security sensitive stuff

For example, golangci-lint (nolint) and revive (revive:disable*) already have these (revive only for justification/reasoning at time of writing).

ccojocar:
fixes #1662

### DIFF
--- a/README.md
+++ b/README.md
@@ -514,6 +514,37 @@ the following:
 gosec -nosec=true ./...
 ```
 
+#### Requiring rule IDs and justifications
+
+To prevent annotations from inadvertently suppressing unrelated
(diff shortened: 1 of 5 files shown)


scop:
Thanks!

Issue #327, opened 2019-06-13
Parsing error in windows


cyw3:
### Summary 
gosec does not work in windows.

### Steps to reproduce the behavior
`
gosec F:/data/...
`
- in linux, it work.
- but in windows, failed:
```
 Including rules: default
 [gosec] 2019/06/13 11:34:09 Excluding rules: default
[gosec] 2019/06/13 11:34:09 Import directory: F:\data\src\back_media
[gosec] 2019/06/13 11:35:17 Checking package: main
[gosec] 2019/06/13 11:35:17 Checking file: F:\data\src\back_media\back_media.go
[gosec] 2019/06/13 11:35:17 Checking file: F:\data\src\back_media\main.go
[gosec] 2019/06/13 11:35:17 Import directory: F:\data\src\cdnutil
 [gosec] 2019/06/13 11:35:23 parsing errors in pkg "cdnutil": parsing line: strconv.Atoi: parsing "\\data\\src\\cdnutil\\ref_api.go": invalid syntax
```

### gosec version
2.0.0
### Go version (output of 'go version')
1.12.6
### Operating system / Environment
```
windows 7
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\xxx\AppData\Local\go-build
set GOEXE=.exe
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=F:\tools\go-1.12.6\go
set GOPROXY=
set GORACE=
set GOROOT=F:\tools\go-1.12.6
set GOTMPDIR=
set GOTOOLDIR=F:\tools\go-1.12.6\pkg\tool\windows_amd64
set GCCGO=gccgo
set CC=gcc
set CXX=g++
set CGO_ENABLED=1
set GOMOD=
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
set GOGCCFLAGS=-m64 -mthreads -fno-caret-diagnostics -Qunused-arguments -fmessag
e-length=0 -fdebug-prefix-map=C:\Users\codedog\AppData\Local\Temp\go-build106690
454=/tmp/go-build -gno-record-gcc-switches
```
### Expected behavior
output result.
### Actual behavior
parsing errors in pkg "cdnutil"

[...]

PR #1673, opened 2026-05-15
Refactor error position parsing to support path with colon.

jim-liu:
Refactor error position parsing with regex to support path with colon.

This is my attempt to solve https://github.com/securego/gosec/issues/327. To reproduce the issue, run **gosec** with a Go file with invalid syntax either on Windows, or on Linux but in a path with colons (such as `~/home/test:code`)

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -762,7 +762,7 @@ func ParseErrors(pkg *packages.Package) (map[string][]Error, error) {
 		return nil, nil
 	}
 	errs := make(map[string][]Error)
-	var posRegexp = regexp.MustCompile(`^(.*?)(?::(\w+))?(?::(\w+))?$`)
+	posRegexp := regexp.MustCompile(`^(.*?)(?::(\w+))?(?::(\w+))?$`)
 	for _, pkgErr := range pkg.Errors {
 		matches := posRegexp.FindStringSubmatch(pkgErr.Pos)
 		file := pkgErr.Pos
--- a/analyzer.go
+++ b/analyzer.go
@@ -29,6 +29,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"regexp"
 	"runtime/debug"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


ccojocar (changes requested):
Please can you add some unit tests for this change? Thanks

jim-liu:
> Please can you add some unit tests for this change? Thanks

Thanks for the heads-up - added a unit test and updated the regex to emit error messages with invalid line/column.

analyzer_test.go
```diff
@@ -1803,6 +1803,26 @@ func main() {
 			Expect(err).Should(HaveOccurred())
 		})
 
+		It("should properly parse the errors with colons in path", func() {
+			pkg := &packages.Package{
+				Errors: []packages.Error{
+					{
+						Pos: "C:\\file:1:2",
```
ccojocar, line 1815: Can you also add an Linux example of path?

For instance: 
```
touch "my:file:with:colons.txt"
```

It is a valid file name in Linux/Unix.

Can you also add a tests for a path without build errors.

analyzer.go
```diff
@@ -761,20 +762,24 @@ func ParseErrors(pkg *packages.Package) (map[string][]Error, error) {
 		return nil, nil
 	}
 	errs := make(map[string][]Error)
+	var posRegexp = regexp.MustCompile(`^(.*?)(?::(\w+))?(?::(\w+))?$`)
```
ccojocar: What about the Linux/Unix file systems? Does this regex cover the case of this file paths:

```
touch "my:file:with:colons.txt"
```

This is a valid path in LInux/Unix.

ccojocar (changes requested):

analyzer.go
jim-liu: The original regex used `\d`, but it will not throw errors for positions with invalid path like `path.go:one:two` and will instead treat the position as path. 

I updated to regex to use `\w` in order to throw errors when parsing positions with invalid line/column. However it will fail path with colons but without extension to fail if the position has no line nor column.

So the updated regex actually worked for your particular case (because of the `.txt`) but won't work if it's `my:file:with:colons`.

analyzer.go
```diff
@@ -761,20 +762,24 @@ func ParseErrors(pkg *packages.Package) (map[string][]Error, error) {
...
 		var err error
-		var line int
-		if len(parts) > 1 {
-			if line, err = strconv.Atoi(parts[1]); err != nil {
-				return nil, fmt.Errorf("parsing line: %w", err)
+		var line, column int
+		if len(matches) > 0 {
+			file = matches[1]
+			file = strings.TrimSuffix(file, ":")
+			if matches[2] != "" {
```
ccojocar: You still need to check if matches is greater than `2` here otherwise it will panic if there are only 2 matches.

```
If len(matches) > 2 && matches[2] != ""
```

analyzer.go
```diff
@@ -761,20 +762,24 @@ func ParseErrors(pkg *packages.Package) (map[string][]Error, error) {
...
+				if line, err = strconv.Atoi(matches[2]); err != nil {
+					return nil, fmt.Errorf("parsing line: %w", err)
+				}
 			}
-		}
-		var column int
-		if len(parts) > 2 {
-			if column, err = strconv.Atoi(parts[2]); err != nil {
-				return nil, fmt.Errorf("parsing column: %w", err)
+			if matches[3] != "" {
```
ccojocar: the same here `len(matches) > 3` otherwise, it will panic.

ccojocar (changes requested):
See my comments regarding the length check for matches, and also there are some lint warnings which needs to be fixed. Thanks

analyzer.go
jim-liu: I'm very new to Go, but my understanding is that FindStringSubmatch returns an array that is the 1 + the number of capturing groups. If an optional capturing group did not match, then the corresponding element is empty.

``` go
func main() {
	re := regexp.MustCompile(`^(.*?)(?::(\w+))?(?::(\w+))?$`)
	fmt.Printf("%d\n", len(re.FindStringSubmatch("file")))		// 4
	fmt.Printf("%d\n", len(re.FindStringSubmatch("file:1")))		// 4
	fmt.Printf("%d\n", len(re.FindStringSubmatch("file:1:2")))	// 4
}
```

jim-liu: I updated the variable declaration to fix the lint warning.

ccojocar: What if the behavior of this API changes, and the returned array is not any longer 4? It's always a good practice to make sure that the array has enough elements when accessing an index.

jim-liu: I don't know if the behavior of `FindStringSubmatch` is something that should be concerned about. I updated the code as requested.

ccojocar (approved):

PR #1697, opened 2026-06-24
fix: min+max G115 false positives

boekkooi-impossiblecloud:
Good day,

I was doing some work and noticed that using `min` + `max` triggered an incorrect G115 warning.

The root cause (in `analyzers/range_analyzer.go:1002-1034`) was that the min/max builtin handlers combined argument ranges using `&&` on both the minValueSet and maxValueSet flags. This was incorrect as only one direction needs `&&`.

per builtin:
- `min(a, b)`: upper bound is `min(a.max, b.max)` and either known max bounds the result. Should be `||` for `maxValueSet`.
- `max(a, b)`: lower bound is `max(a.min, b.min)` and either known min bounds the result. Should be `||` for `minValueSet`.

For `uint8(max(0, min(MaxUint8, data[i])))`, `data[i]` has no derivable range,
so the `&&` clobbered the constant bound from the other arg and the analyzer lost the `[0, 255]` envelope.

Thank you for reviewing this PR and maintaining this very useful tool!


boekkooi-impossiblecloud:
Thanks for the review and merge @ccojocar :smile:

PR #1702, opened 2026-07-11
feat(G101): detect AWS temporary access keys

Lhilove:
## Summary

This PR extends G101 hardcoded credential detection to identify AWS temporary access keys with the `ASIA` prefix.

Currently, G101 detects long-lived AWS access keys (`AKIA`) but does not detect temporary STS access keys (`ASIA`), which are commonly used with IAM roles, AWS IAM Identity Center (AWS SSO), and CI/CD workflows.

## Changes

- Added detection for AWS temporary access keys matching `ASIA[0-9A-Z]{16}`.
- Added unit tests covering:
  - Variable assignment
  - Struct literals
  - Map literals

This change is backward-compatible and extends the existing AWS credential detection.

### DIFF
--- a/rules/hardcoded_credentials.go
+++ b/rules/hardcoded_credentials.go
@@ -64,6 +64,10 @@ var secretsPatterns = [...]secretPattern{
 		regexp: regexp.MustCompile(`AKIA[0-9A-Z]{16}`),
 	},
 	{
+		name:   "AWS Temporary Access Key",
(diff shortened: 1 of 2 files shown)


Lhilove:
The barry-ai-security-review failure appears to be a pull_request_target security restriction on fork PRs rather than an issue with my changes test, taint-perf-guard, and coverage all passed. Happy to make any requested changes.

ccojocar (approved):

PR #1707, opened 2026-07-18
fix(G202): don't flag strings.Builder built from constants

### LINKED ISSUE #1701: G202 with strings.Builder
I've bumped into fun little issue with G202 and the way string concatenation in loop is handled, += in loop is considered inefficient so we have strings.Builder instead, but use of that triggers gosec's G202
tested with 2.27.1 and latest commit

```
package main

import (
    "database/sql"
    "strings"
)

func main() {
    db, err := sql.Open("sqlite3", ":memory")
    if err != nil {
        panic(err)
    }

    placeholders1 := "?, ?"
    q1 := "SELECT col1 FROM table1 WHERE id IN (" + placeholders1 + ")"
    rows1, err := db.Query(q1)
    if err != nil {
        panic(err)
    }
    defer rows1.Close()

    ids := []int{1, 2, 3}

    placeholders2 := ""
    args2 := make([]any, 0, len(ids))
    for i, id := range ids {
        if i > 0 {
            placeholders2 += ","
        }
        placeholders2 += "?"
        args2 = append(args2, id)
    }
    q2 := "SELECT col1 FROM table1 WHERE id IN (" + placeholders2 + ")"
    rows2, err := db.Query(q2, args2)
    if err != nil {
        panic(err)
    }
    defer rows2.Close()

    var phBuilder strings.Builder
    args3 := make([]any, 0, len(ids))
    for i, id := range ids {
        if i > 0 {
            phBuilder.WriteString(",")
        }
        phBuilder.WriteString("?")
        args3 = append(args3, id)
    }
    placeholders3 := phBuilder.String()
    q3 := "SELECT col1 FROM table1 WHERE id IN (" + placeholders3 + ")"
    rows3, err := db.Query(q3, args3)
    if err != nil {
        panic(err)
    }
    defer rows3.Close()
}
```

ravisastryk:
# fix(G202): don't flag `strings.Builder` built from constants

## Problem

G202 (SQL string concatenation) reports a false positive when a query is built by concatenating the result of a `strings.Builder` (or `bytes.Buffer`) whose content is only constant literals — the common, efficient idiom for building
SQL `IN (...)` placeholder lists.

The inconsistency: the equivalent `+=` construction is **not** flagged, but the `strings.Builder` replacement (recommended for efficiency) **is**.

Fixes: #1701 

```go
var ph strings.Builder
for i := range ids {
    if i > 0 {
        ph.WriteString(",")
    }
    ph.WriteString("?")
}
placeholders := ph.String()
q := "SELECT col1 FROM table1 WHERE id IN (" + placeholders + ")" // G202 (false positive)
```

## Root cause

`TryResolve` treats any function-call result as unresolved (tainted), so `builder.String()` was always considered non-constant, regardless of what was written into the builder.

## Fix

`resolveCallExpr` now recognizes `strings.Builder.String()` / `bytes.Buffer.String()` and resolves it to a constant **only when every write into that builder is a constant**. It stays conservative - a non-constant write, an opaque initializer, an escaping address, or a package-level builder all keep the query flagged. Genuinely tainted builders are still detected.

## How to test (quick)

Unit tests (covers the new positive + negative samples):

```bash
go test -run TestRules ./rules/
```

End-to-end — drop this into a file and scan it:

```go
// repro.go
package main

import (
	"database/sql"
	"os"
	"strings"
)

func main() {
	db, _ := sql.Open("sqlite3", ":memory:")

	// SAFE — builder fed only constants: should NOT be flagged
	var ok strings.Builder
	for i := 0; i < 3; i++ {
		if i > 0 {
			ok.WriteString(",")
		}
		ok.WriteString("?")
	}
	rows, _ := db.Query("SELECT * FROM t WHERE id IN (" + ok.String() + ")")
	defer rows.Close()

	// TAINTED — builder fed os.Args: should STILL be flagged
	var bad strings.Builder
	for _, a := range os.Args {
		bad.WriteString(a)
	}
	rows2, _ := db.Query("SELECT * FROM t WHERE id IN (" + bad.String() + ")")
	defer rows2.Close()
}
```

```bash
go run ./cmd/gosec/ -include=G202 ./path/to/repro/...
```

**Expected:** exactly **1 issue** — the tainted builder is reported, the constant-only builder is not. Before this change, both were reported.

### DIFF
--- a/resolve.go
+++ b/resolve.go
@@ -14,7 +14,11 @@
 
 package gosec
 
-import "go/ast"
+import (
+	"go/ast"
+	"go/token"
+	"go/types"
+)
 
 func resolveIdent(n *ast.Ident, c *Context) bool {
 	if n.Obj == nil || n.Obj.Kind != ast.Var {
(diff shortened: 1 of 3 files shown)


ravisastryk:
Heads-up: the barry-ai-security-review job is failing on the new actions/checkout@v7 "pull_request_target" guard, not on this PR's changes, the checkout step needs `allow-unsafe-pr-checkout: true` (or an equivalent adjustment). Happy to open a separate workflow PR if that's helpful @ccojocar so you can make that change and then this should pass.

ccojocar (approved):

PR #1719, opened 2026-08-15
fix(report): sarif and sonarqube paths depend on scan target order

VXNCXNX:
## What's broken

The paths in the `sarif` and `sonarqube` reports depend on the order the scan targets were given on the command line.

Same project, same issues, two orderings:

```
$ gosec -fmt=sarif ./... ./sub/...     ->  "uri": "a.go"   "uri": "b.go"
$ gosec -fmt=sarif ./sub/... ./...     ->  "uri": "a.go"   "uri": "sub/b.go"
```

And with two sibling directories whose names share a prefix, `p` and `p2`:

```
$ gosec -fmt=sarif /tmp/d/p2/... /tmp/d/p/...   ->  "uri": "/tmp/d/p2/a.go"
$ gosec -fmt=sarif /tmp/d/p/... /tmp/d/p2/...   ->  "uri": "a.go"
```

The first one emits an absolute path into a report format whose paths are supposed to be relative to the project root. SonarQube resolves file paths against the project, so an absolute one there does not match any file and the issue lands nowhere.

## Cause

Two separate defects in the same four-line loop, duplicated in `report/sarif/formatter.go` and `report/sonar/formatter.go`.

**The loop never stops.** It assigns `filePath` for every root that matches, so the last match wins rather than the right one. With `./... ./sub/...` the enclosing root is checked last and `sub/b.go` gets trimmed against the wrong one.

**`HasPrefix` has no path boundary.** Root `/tmp/d/p` is a string prefix of file `/tmp/d/p2/a.go`, so the branch is taken. `strings.Replace(file, root+"/", "", 1)` then looks for `/tmp/d/p/` inside that path, finds nothing, and returns the file unchanged, which is how an absolute path reaches the report.

## The fix

Match on `root + "/"` so a sibling cannot match, and keep the longest matching root instead of the last.

Longest rather than a plain `break`: roots are one `filepath.Abs` per CLI argument (`getRootPaths`, `cmd/gosec/main.go:310`), so nested arguments are legitimate and both genuinely contain the file. The report should be relative to the innermost root that contains it, which is a property of the paths and not of the argument order. A `break` on first match would still give two different answers for two orderings.

The `root + "/"` boundary also makes the slice safe, replacing a `Replace` that could silently no-op on a false prefix.

After, all four commands above agree, and `-fmt=sarif ./a.go` on a single-file argument is unchanged.

## Verification

A spec in each of `report/sarif/sarif_test.go` and `report/sonar/sonar_test.go`, covering both orderings of a nested pair and both orderings of a sibling pair.

Reverting either formatter fails its own spec and nothing else:

```
[FAILED] Expected
    <string>: sub/test.go
to equal
    <string>: test.go
```

`go test ./report/...` is green. The full `go test ./...` ends in FAIL only on `autofix / TestGenerateSolution_ClaudeProvider`, which expects an error from a live Claude API call and fails identically on an untouched tree here.

### DIFF
--- a/report/sarif/formatter.go
+++ b/report/sarif/formatter.go
@@ -186,9 +186,14 @@ func parseSarifLocation(i *issue.Issue, rootPaths []string) (*Location, error) {
 
 func parseSarifArtifactLocation(i *issue.Issue, rootPaths []string) *ArtifactLocation {
 	var filePath string
+	// Select the most specific root path which actually contains the file, so that the
+	// emitted path does not depend on the order the root paths were given on the command line.
+	longest := -1
 	for _, rootPath := range rootPaths {
-		if strings.HasPrefix(i.File, rootPath) {
-			filePath = strings.Replace(i.File, rootPath+"/", "", 1)
+		root := strings.TrimSuffix(rootPath, "/")
+		if len(root) > longest && strings.HasPrefix(i.File, root+"/") {
+			longest = len(root)
+			filePath = i.File[len(root)+1:]
 		}
 	}
 	return NewArtifactLocation(filePath)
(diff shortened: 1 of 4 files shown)


ccojocar (approved):

ccojocar:

VXNCXNX:

PR #1723, opened 2026-08-18
analyzers/G703: stop treating Clean/Abs/PathEscape as path-traversal sanitizers

### LINKED ISSUE #1721: Taint Analysis: `path.Clean()` is not a Sanitizer
## Summary

`path.Clean()` and `filepath.Clean()` are incorrectly treated as sanitizers for path traversal taint analysis.
They should be removed, and/or replaced with functions from [`os.Root`](https://pkg.go.dev/os#Root)

## Details

> // filepath.Clean normalizes and removes traversal components

[analyzers/pathtraversal.go#L69](https://github.com/securego/gosec/blob/master/analyzers/pathtraversal.go#L69)

While this comment is true, it is not "safe".
A better understanding of `path.Clean()` is that it *evaluates* `..` elements in a path.

> Eliminate each inner .. path name element (the parent directory) along with the non-.. element that precedes it.

[pkg.go.dev/path#Clean](https://pkg.go.dev/path#Clean)

While the `..` element is removed in the result, it also removes the parent element.
*The result has traversed the path!*

## Minimal Reproducer

```go
package main

import (
	"fmt"
	"net/url"
	"os"
	"path"
	"path/filepath"
)

func main() {
	pwd, err := os.Getwd()
	if err != nil {
		fmt.Printf("ERROR: os.Getwd() error=%v\n", err)
		os.Exit(1)
	}

	file := path.Join(pwd, "prefix", os.Args[1])

	filename := path.Clean(file)
	_, err = os.Stat(filename)
	fmt.Printf("path.Clean(%s) error=%v\n", filename, err)

	filename = filepath.Clean(file)
	_, err = os.Stat(filename)
	fmt.Printf("filepath.Clean(%s) error=%v\n", filename, err)

	filename, err = filepath.Abs(file)
	if err != nil {
		fmt.Printf("ERROR: filepath.Abs(%s) error=%v\n", file, err)
		os.Exit(1)
	}
	_, err = os.Stat(filename)
	fmt.Printf("filepath.Abs(%s) error=%v\n", filename, err)

	filename, err = url.PathUnescape(url.PathEscape(file))
	if err != nil {
		fmt.Printf("ERROR: url.PathUnescape(url.PathEscape((%s)) error=%v\n", file, err)
		os.Exit(1)
	}
	_, err = os.Stat(filename)
	fmt.Printf("url.PathEscape(%s) error=%v\n", filename, err)
}
```

After saving the above code to `main.go`:

```bash
# create a local file
$ mkdir prefix
$ touch prefix/file

# run with a safe parameter, nil errors is correct
$ go run . file
path.Clean(/example/absolue/path/prefix/file) error=<nil>
filepath.Clean(/example/absolue/path/prefix/file) error=<nil>
filepath.Abs(/example/absolue/path/prefix/file) error=<nil>
url.PathEscape(/example/absolue/path/prefix/file) error=<nil>

# run with a dangerous parameter, oops path traversal!
$ go run . ../../../../../../../../../../../../../../etc/passwd
path.Clean(/etc/passwd) error=<nil>
filepath.Clean(/etc/passwd) error=<nil>
filepath.Abs(/etc/passwd) error=<nil>
url.PathEscape(/etc/passwd) error=<nil>
```

### Gosec does not report an issue

```bash
# scan with gosec
$ go run github.com/securego/gosec/v2/cmd/gosec@latest .
[...]
Results:


Summary:
  Gosec  : dev
  Files  : 1
  Lines  : 43
  Nosec  : 0
  Issues : 0
```

## Suggested Actions

- Remove `path.Clean()` and `filepath.Clean()` from Sanitizers
- Also remove indirect sanitizers that depend on `Clean()` (e.g. `filepath.Abs()`)
- Remove `url.PathEscape()` from Sanitizers, since using `url.PathUnescape()` restores path traversal components

```diff
diff --git a/analyzers/pathtraversal.go b/analyzers/pathtraversal.go
index ba9b72e..bcaff20 100644
--- a/analyzers/pathtraversal.go
+++ b/analyzers/pathtraversal.go
@@ -66,22 +66,15 @@ func PathTraversal() taint.Config {
 			{Package: "net/http", Method: "ServeFileFS", CheckArgs: []int{3}},
 		},
 		Sanitizers: []taint.Sanitizer{
-			// filepath.Clean normalizes and removes traversal components
-			{Package: "path/filepath", Method: "Clean"},
-			// filepath.Abs calls Clean internally (per Go docs)
-			{Package: "path/filepath", Method: "Abs"},
 			// filepath.Base extracts just the filename, removing directory traversal
 			{Package: "path/filepath", Method: "Base"},
 			// filepath.Rel computes a relative path safely
 			{Package: "path/filepath", Method: "Rel"},
-			// url.PathEscape escapes path components
-			{Package: "net/url", Method: "PathEscape"},
 
-			// path.Base and path.Clean provide identical traversal-stripping
-			// semantics as their filepath counterparts (the only difference is
+			// path.Base provides identical traversal-stripping
+			// semantics as the filepath counterpart (the only difference is
 			// separator handling, which is irrelevant for security).
 			{Package: "path", Method: "Base"},
-			{Package: "path", Method: "Clean"},
 
 			// Integer conversions eliminate path traversal vectors entirely —
 			// the result can never contain "/" or ".." characters.
```

ChrisJr404:
This makes G703 stop treating `path.Clean`, `filepath.Clean`, `filepath.Abs` and `url.PathEscape` as path-traversal sanitizers, because none of them actually confine a path — they only normalize it, so tainted input still traverses.

`Clean` (both `path` and `path/filepath`) *evaluates* `..` elements instead of rejecting them, per the [docs](https://pkg.go.dev/path#Clean): "Eliminate each inner .. path name element (the parent directory) along with the non-.. element that precedes it." So `Clean("/prefix/../../../../etc/passwd")` returns `/etc/passwd` — the traversal has already happened. `filepath.Abs` calls `Clean` internally and inherits exactly this behavior, and `url.PathEscape` is reversible: `url.PathUnescape(url.PathEscape(p))` restores any `..` that was escaped. Treating these as sanitizers caused G703 to silently drop real traversal flows (a false negative), as reported in #1721 with a runnable reproducer.

The change drops those four from the `Sanitizers` set in `analyzers/pathtraversal.go` and rewrites the misleading "normalizes and removes traversal components" comment to explain why. `filepath.Base` / `path.Base` (which strip to the final path element and genuinely cannot traverse) and `filepath.Rel` are left in place.

Testing: the two G703 samples that previously flowed a tainted path through `filepath.Clean` and `filepath.Abs` and expected zero findings are now true positives (expected count updated to 1), and I added new samples covering `path.Clean` and the `url.PathEscape`/`url.PathUnescape` round-trip. The `path.Base` and `strconv` samples still expect zero, confirming the remaining sanitizers are unaffected. `go test ./analyzers/` (G703 spec) passes, and running the built binary against the issue's reproducer now reports the previously-missed flows through `Clean`/`Abs` while leaving `Base` clean.

Fixes #1721

### DIFF
--- a/analyzers/pathtraversal.go
+++ b/analyzers/pathtraversal.go
@@ -66,22 +66,25 @@ func PathTraversal() taint.Config {
 			{Package: "net/http", Method: "ServeFileFS", CheckArgs: []int{3}},
 		},
 		Sanitizers: []taint.Sanitizer{
-			// filepath.Clean normalizes and removes traversal components
-			{Package: "path/filepath", Method: "Clean"},
-			// filepath.Abs calls Clean internally (per Go docs)
-			{Package: "path/filepath", Method: "Abs"},
 			// filepath.Base extracts just the filename, removing directory traversal
 			{Package: "path/filepath", Method: "Base"},
 			// filepath.Rel computes a relative path safely
 			{Package: "path/filepath", Method: "Rel"},
-			// url.PathEscape escapes path components
-			{Package: "net/url", Method: "PathEscape"},
 
-			// path.Base and path.Clean provide identical traversal-stripping
-			// semantics as their filepath counterparts (the only difference is
-			// separator handling, which is irrelevant for security).
+			// path.Base provides identical traversal-stripping semantics as its
+			// filepath counterpart (the only difference is separator handling,
+			// which is irrelevant for security).
 			{Package: "path", Method: "Base"},
-			{Package: "path", Method: "Clean"},
+
+			// NOTE: Clean/filepath.Clean, filepath.Abs and url.PathEscape are
+			// deliberately NOT sanitizers. Clean only lexically normalizes a
+			// path: it evaluates ".." elements instead of rejecting them, so
+			// Clean("/prefix/../../etc/passwd") yields "/etc/passwd" — the
+			// traversal succeeds. filepath.Abs calls Clean internally and
+			// inherits the same behavior, and url.PathUnescape reverses
+			// url.PathEscape, restoring any ".." that was escaped. Confining a
+			// path to a directory requires os.Root or an explicit prefix check,
+			// not Clean. See https://github.com/securego/gosec/issues/1721.
 
 			// Integer conversions eliminate path traversal vectors entirely —
 			// the result can never contain "/" or ".." characters.
--- a/testutils/g703_samples.go
+++ b/testutils/g703_samples.go
@@ -84,6 +84,8 @@ func openFromArgs() {
 	}
 }
 `}, 1, gosec.NewConfig()},
+	// True positive: filepath.Clean only normalizes, it does not confine the
+	// path, so ".." in the user input still traverses out of any prefix.
 	{[]string{`
 package main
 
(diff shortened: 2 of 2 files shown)


nicholascapo (approved):

ccojocar (approved):

ccojocar:
@ChrisJr404 Can you please rebase this pull request? Thanks

ChrisJr404:
Rebased onto master, thanks.

PR #1733, opened 2026-08-28
fix(taint): make wide caller traversal deterministic

### LINKED ISSUE #1712: Taint analyzers still report findings nondeterministically on v2.28.0 when a sink has more than maxCallerEdges callers (follow-up to #1711)
**gosec version:** reproduced on **v2.28.0** (current release, 2026-07-14) and on v2.25.0; the responsible code is unchanged on `master` as of `5e72efc19` (2026-08-03)
**Go version:** 1.26.4
**OS:** reproduced on both darwin/arm64 and linux/amd64 (GitHub Actions `ubuntu-latest`)

This is a follow-up to #1711, which was closed by its author after they could not reproduce on
v2.28.0. It does still reproduce there — their reproduction depended on one particular finding in a
real codebase, and the rule that produced it changed. The reproducer below controls the one variable
that matters, the number of callers of the sink, so it does not depend on any rule staying the same.

#### What happens

On an unchanged tree, with the same binary, `gosec ./...` reports a G706 finding on roughly 5% of runs and reports nothing on the other 95%. In CI this makes the gate a coin flip: the same commit failed one run and passed a re-run of the same job with `Issues: 0`.

We measured 87 runs on one commit of our own codebase: **4 runs reported the finding, 83 did not.** `-concurrency 1` does not change this (1 finding in 20 runs).

#### Root cause

`taint/taint.go`, `isParameterTainted()`:

```go
const maxCallerEdges = 32
...
edgesChecked := 0
for _, inEdge := range node.In {
    if edgesChecked >= maxCallerEdges {
        break
    }
    ...
}
```

`node.In` comes from the CHA call graph, whose edge order is not stable between runs (it is built from map iteration). When the sink lives in a function with more than 32 callers, the analyzer examines an effectively random 32-caller subset, so whether it ever looks at the one tainted caller is chance.

The observed detection rate matches `maxCallerEdges / len(node.In)` closely. #1711 also points out
that `isTainted` writes `visited[v] = true` before it has decided whether `v` is tainted, and that
`isParameterTainted` shares that one map across every caller it tries — so whichever caller is
examined first can record a "not tainted" answer that blocks a later caller which would have proved
the opposite. Both effects are still present in v2.28.0 (`maxCallerEdges = 32`, no sorting of
`node.In`, and the same shared `visited` map).

#### Minimal reproducer

`main.go` — one sink helper, one tainted caller, N untainted callers (generator script and full file attached below):

```go
package main

import (
	"log/slog"
	"net/http"
)

// sink: taint must arrive through the parameter, so the analyzer has to walk
// incoming call-graph edges to find it.
func logIt(msg string) {
	slog.Warn(msg)
}

// the ONE tainted caller
func tainted(w http.ResponseWriter, r *http.Request) {
	logIt(r.URL.Path)
	_ = w
}

func filler0() { logIt("static 0") }
// ... filler1 .. fillerN-1, identical

func main() {
	http.HandleFunc("/", tainted)
}
```

Run `gosec ./...` repeatedly and count how many runs report G706:

| callers (`len(node.In)`) | `32 / callers` | v2.25.0 | **v2.28.0** |
|---|---|---|---|
| 21 (below the cap) | 100% | 12 / 12 | **6 / 6** |
| 101 | 32% | 2 / 12 | — |
| 601 | 5.3% | 1 / 20 | **3 / 20** |

Below the cap the finding is reported on every run, on both versions. Above it, the rate tracks the
ratio and the result differs run to run. On our own codebase (a central error helper with 604
callers) the same behaviour shows up as 4 findings in 87 runs on v2.25.0 and 2 in 20 on v2.28.0.

#### Why this matters beyond one rule

The same engine backs every taint-based analyzer — `sqlinjection`, `commandinjection`, `xss`, `pathtraversal`, `ssrf`, `ssti`, `smtpinjection`, `unsafe_deserialization`, `loginjection`, `form_parsing_limits`. A centralized helper with many callers is an ordinary Go pattern (ours has 604), so for that shape these rules are not a gate that can be relied on, in either direction: a real finding can be missed, and a clean run cannot be cited as evidence.

#### Suggested fixes

1. **Make the order deterministic** — sort `node.In` (e.g. by caller position/name) before truncating, so a given tree always produces the same result. This alone turns a flaky gate into a consistent one, even if coverage stays capped. The cap itself came in with #1610 to stop a hang on wide CHA graphs, which is a real problem worth keeping fixed — it is only the unordered truncation that makes the outcome vary.
2. **Prefer edges that can carry taint** — e.g. examine call sites whose argument at that index is not a constant first. In the reproducer that is 1 edge out of 601.
3. **Make the cap configurable** and report when it is hit, so users can tell "no finding" apart from "gave up after 32 of 604 callers".

We would be happy to test a patch against a real 126k-line codebase.

#### Generator for the reproducer

```python
#!/usr/bin/env python3
"""Write flaky/main.go: one sink helper, one tainted caller, N untainted callers."""
import os, sys

n = int(sys.argv[1]) if len(sys.argv) > 1 else 600
out = [
    'package main', '',
    'import (', '\t"log/slog"', '\t"net/http"', ')', '',
    '// sink: taint must arrive through the parameter, so the analyzer has to walk',
    '// incoming call-graph edges to find it.',
    'func logIt(msg string) {', '\tslog.Warn(msg)', '}', '',
    '// the ONE tainted caller',
    'func tainted(w http.ResponseWriter, r *http.Request) {', '\tlogIt(r.URL.Path)', '\t_ = w', '}', '',
]
out += [f'func filler{i}() {{ logIt("static {i}") }}' for i in range(n)]
out += ['', 'func main() {', '\thttp.HandleFunc("/", tainted)', '}', '']

os.makedirs('flaky', exist_ok=True)
with open('flaky/main.go', 'w') as f:
    f.write('\n'.join(out))
print(f'callers = {n + 1}   →  32/{n + 1} = {32 * 100 / (n + 1):.1f}%')
```

Then, for each N:

```sh
python3 gen-flaky.py 600
hits=0
for i in $(seq 1 20); do
  v=$(gosec ./flaky/ 2>&1 | grep -E '^  Issues :' | awk '{print $3}')
  [ "$v" != "0" ] && hits=$((hits+1))
done
echo "$hits/20 runs reported a finding"
```

A single run tells you nothing here — above the cap, both a red run and a green run are expected outcomes on the same code, which is the whole problem.

SetraTheXX:
Fixes #1712

## Summary

Taint analysis could miss a real flow when a helper or sink function has many incoming CHA caller edges. The incoming edge set is explicitly unordered, so truncating it at an arbitrary prefix made the result depend on call-graph construction order.

## Root cause

isParameterTainted previously examined only the first 32 incoming edges from node.In. CHA can produce a wide caller set, and node.In does not guarantee a stable order. A tainted caller beyond that prefix could therefore be skipped.

## Changes

- Keep bounded traversal, with maxCallerEdges set to 1024. The issue reproducer has about 604 incoming call sites; 1024 covers that case with headroom, while preserving the safety bound introduced by #1610.
- Sort incoming edges by caller function and call-site position only when the edge set exceeds the cap. Smaller caller sets avoid the additional sorting and allocation cost.
- Keep the focused issue-scale regression with roughly 600 incoming call sites.
- Add a cap-boundary regression with more than maxCallerEdges incoming edges. It moves the tainted edge to the end of the unordered input and verifies that deterministic ordering retains it.
- Extend the safe wide-caller regression beyond the cap.
- Add BenchmarkTaintAnalysisAtCallerEdgeCap to measure the sorting and bounded traversal path at the configured limit.

## Scope

This PR remains limited to taint caller-edge traversal and its regression/performance coverage. It does not change individual analyzer rules, remove the hang-prevention bound, redesign the taint engine, or introduce unrelated refactoring.

## Validation

Passed locally:

- go test ./taint -count=1
- Focused caller-edge regression tests repeated 20 times
- go test ./taint -run '^$' -bench '^BenchmarkTaintAnalysis(ManySinkCalls|AtCallerEdgeCap)$' -benchmem -count=1 -benchtime=1s
- go test ./... -run '^$' -count=1
- go vet ./...
- go build ./...
- gofmt and git diff --check

On the same Windows host, the cap-boundary benchmark measured approximately 4.4 ms/op, 1.25 MB/op, and 77k allocations/op. The existing many-sink benchmark remains below the cap and therefore does not sort incoming edges.

The full go test ./... -short -count=1 -timeout=120s run still exposes unrelated Windows/environment baseline failures: analyzer and report expectations, temporary-file locking, and timeouts in the cmd/gosec, cmd/gosecutil, and rules suites. The changed taint package passes. The local race command cannot run because this Windows environment has CGO disabled and no gcc toolchain.

### DIFF
--- a/taint/analyzer_internal_test.go
+++ b/taint/analyzer_internal_test.go
@@ -506,7 +506,8 @@ func caller%d(w W, data []byte) { w.Write(data) }
 const wideCallerCount = 600
 
 // buildWideCallerTaintFixture creates a sink helper with a wide set of incoming
-// callers, optionally including one caller that passes a source-typed field.
+// callers and multiple call sites, optionally including one caller that passes a
+// source-typed field.
 func buildWideCallerTaintFixture(tb testing.TB, fillerCount int, includeTainted bool) (*ssa.Program, *ssa.Function) {
 	tb.Helper()
 
@@ -522,7 +523,7 @@ func logIt(message string) { sink(message) }
 		src += "\nfunc tainted(r *Request) { logIt(r.Path) }\n"
 	}
 	for i := 0; i < fillerCount; i++ {
-		src += fmt.Sprintf("func filler%d() { logIt(\"static %d\") }\n", i, i)
+		src += fmt.Sprintf("func filler%d() { logIt(\"static %d\"); logIt(\"static duplicate %d\") }\n", i, i, i)
 	}
 
 	fset := token.NewFileSet()
--- a/taint/analyzer_internal_test.go
+++ b/taint/analyzer_internal_test.go
@@ -503,6 +503,8 @@ func caller%d(w W, data []byte) { w.Write(data) }
 	return prog, srcFuncs
 }
 
+// The issue reproducer has roughly 600 incoming call sites. Keep a focused
+// fixture at that scale to cover the reported false-negative case.
 const wideCallerCount = 600
 
 // buildWideCallerTaintFixture creates a sink helper with a wide set of incoming
@@ -520,10 +522,14 @@ func sink(message string) {}
 func logIt(message string) { sink(message) }
 `
 	if includeTainted {
-		src += "\nfunc tainted(r *Request) { logIt(r.Path) }\n"
+		src += "\nfunc aTainted(r *Request) { logIt(r.Path) }\n"
 	}
 	for i := 0; i < fillerCount; i++ {
-		src += fmt.Sprintf("func filler%d() { logIt(\"static %d\"); logIt(\"static duplicate %d\") }\n", i, i, i)
+		if i == 0 {
+			src += fmt.Sprintf("func filler%d() { logIt(\"static %d\"); logIt(\"static duplicate %d\") }\n", i, i, i)
+			continue
+		}
+		src += fmt.Sprintf("func filler%d() { logIt(\"static %d\") }\n", i, i)
 	}
 
 	fset := token.NewFileSet()
@@ -554,7 +560,7 @@ func logIt(message string) { sink(message) }
 	return prog, logFn
 }
 
-func TestTaintAnalysisFindsTaintedCallerRegardlessOfCallerEdgeOrder(t *testing.T) {
+func TestTaintAnalysisFindsTaintedCallerBeyondPreviousEdgeLimit(t *testing.T) {
 	t.Parallel()
 
 	prog, logFn := buildWideCallerTaintFixture(t, wideCallerCount, true)
@@ -565,11 +571,11 @@ func TestTaintAnalysisFindsTaintedCallerRegardlessOfCallerEdgeOrder(t *testing.T
 		t.Fatalf("expected logIt to have more than %d incoming edges, got %d", previousMaxCallerEdges, len(node.In))
 	}
 
-	// Put the tainted caller after the retained prefix. This models an
-	// otherwise valid CHA edge order in which the current cap hides the flow.
+	// Put the tainted caller after the previous retained prefix. This models an
+	// otherwise valid CHA edge order in which the old cap hid the flow.
 	taintedIndex := -1
 	for i, edge := range node.In {
-		if edge.Caller != nil && edge.Caller.Func != nil && edge.Caller.Func.Name() == "tainted" {
+		if edge.Caller != nil && edge.Caller.Func != nil && edge.Caller.Func.Name() == "aTainted" {
 			taintedIndex = i
 			break
 		}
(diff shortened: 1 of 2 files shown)


SetraTheXX:
Thanks for flagging the coverage gap. I pushed d4f05f8 with a test-only follow-up for the caller-edge comparator.

The wide-callers fixture now includes multiple call sites from the same filler function, so the `Site.Pos()` tie-breaker path is exercised without changing the production traversal logic or widening the scope of this PR.

Validated locally:
- focused caller-edge regression tests, 20 runs
- full `taint` package tests
- compile-only `go test ./... -run '^$'`
- `go vet ./...`
- `go build ./...`
- gofmt and `git diff --check`

The Windows environment cannot run the race detector locally because CGO is disabled and no gcc toolchain is installed; the CI run remains the source of truth for the race/coverage jobs.

taint/taint.go
```diff
@@ -26,8 +27,9 @@ const maxTaintDepth = 50
 // maxCallerEdges caps the number of incoming call graph edges examined per function
 // in isParameterTainted. CHA over-approximates call graphs (every interface method
 // call fans out to ALL implementations), so a function can have thousands of callers.
-// Real taint flows come from direct/nearby callers, not the 33rd+ CHA-generated edge.
-const maxCallerEdges = 32
+// Incoming edges are sorted before this cap is applied so that any truncation is
+// deterministic, while the higher limit covers ordinary functions with many direct callers.
+const maxCallerEdges = 4096
```
ccojocar: Why such a large increase from 32? Why not 512 or 1024? Do you have any validation tests which requires so many edges?

taint/taint.go
```diff
@@ -946,10 +948,36 @@ func (a *Analyzer) isParameterTainted(param *ssa.Parameter, fn *ssa.Function, vi
 		adjustedIdx = paramIdx
 	}
 
+	// Sort incoming edges before capping so the result does not depend on the
+	// iteration order used while CHA builds its graph.
+	inEdges := append([]*callgraph.Edge(nil), node.In...)
+	sort.SliceStable(inEdges, func(i, j int) bool {
```
ccojocar: What's the real benefit of this sorting beside increasing the upper limit? This creates some determinism but it does have performance penalties when the edge number is large. Do you have any guess what that might be?

My feeling is that the issue was caused the `32` upper limit which just cap the number of edges.

taint/analyzer_internal_test.go
```diff
@@ -503,6 +503,111 @@ func caller%d(w W, data []byte) { w.Write(data) }
 	return prog, srcFuncs
 }
 
+const wideCallerCount = 600
```
ccojocar, line 508: Why 600? This should be closed to max. Maybe considering adding also a benchmark test for max accepted edges?

taint/taint.go
SetraTheXX: Thanks for the pushback. I reduced maxCallerEdges to 1024. The reproducer in this issue has about 604 incoming call sites, so 512 would still truncate that case while 1024 leaves headroom and keeps the bounded traversal introduced by #1610. I added a focused ~600-edge regression and a boundary test that creates more than maxCallerEdges edges and verifies the tainted caller is retained after the unordered input is reordered.

taint/taint.go
SetraTheXX: The sorting is still needed for the bounded case: without it, the retained prefix depends on the unordered Node.In slice. I changed it to run only when len(node.In) > maxCallerEdges; caller sets within the cap use the original slice, so ordinary functions do not pay the sort/copy cost. The new cap-boundary benchmark measures the actual sorting/truncation path at 1024 (about 4.4 ms/op and 1.25 MB/op on my Windows host).

taint/analyzer_internal_test.go
SetraTheXX: Agreed. I kept 600 only for the issue-scale regression; it is not intended to validate the configured maximum. I added a separate boundary test and BenchmarkTaintAnalysisAtCallerEdgeCap using the configured 1024 limit plus overflow (the fixture has 1026 incoming call sites because one filler has two call sites). This separates issue reproduction from cap and benchmark coverage.

ccojocar (approved):

PR #1741, opened 2026-09-12
fix: make nosec suppression independent of comment order

### LINKED ISSUE #1740: #nosec suppression is nondeterministic (identical source, same binary)
`#nosec` suppression is nondeterministic: the same finding is suppressed on some runs and reported on others (identical source, same binary)

[reproduce.sh](https://github.com/user-attachments/files/32015457/reproduce.sh)

## Summary

Running gosec repeatedly over **unchanged** source produces **different results**.
A `G101` finding that is covered by a `#nosec G101` directive is correctly
suppressed on most runs, but leaks through as a reported issue on a minority of
runs — no source change, no flag change, same gosec binary.

This makes gosec unusable as a hard CI gate: a job that runs `gosec ./...` and
fails on any HIGH finding flakes red at random.

## Environment

- gosec: **v2.29.0** (also reproduces on `@latest` built from source)
- Go: `go1.26.8` (also seen on 1.22/1.23)
- OS: reproduced on macOS (arm64) and Ubuntu (GitHub-hosted `ubuntu-latest`)

## Minimal reproduction

Self-contained package (stdlib only, compiles, `go vet`-clean):

```go
package repro

import (
	"fmt"
	"net/url"
	"os"
)

// #nosec G101 -- example connection strings in help text are not real credentials
var serveCmd = &struct {
	Use, Long string
	RunE      func(args []string) error
}{
	Use: "serve",
	Long: `Config file format:
  databases:
    db1:
      url: "postgres://user:pass@host:5432/db1"
    db2:
      url: "postgres://user:pass@host:5432/db2"
`,
	RunE: func(args []string) error {
		cfgPath := os.Getenv("CFG")
		raw, err := os.ReadFile(cfgPath) // #nosec G304 -- operator-supplied path
		if err != nil {
			return fmt.Errorf("read: %w", err)
		}
		_ = raw
		u, _ := url.Parse(os.Getenv("DB_URL"))
		if u != nil && u.User != nil {
			u.User = url.UserPassword(u.User.Username(), "x")
		}
		return nil
	},
}

func main() { _ = serveCmd }
```

Run gosec ~50–60 times and count the G101 findings each run:

```bash
for i in $(seq 1 60); do
  gosec -fmt json -out out.json ./... >/dev/null 2>&1 || true
  python3 -c "import json;print(len([x for x in json.load(open('out.json'))['Issues'] if x['rule_id']=='G101']),end=' ')"
done; echo
```

A ready-made script (`reproduce.sh`) that generates the package, runs gosec N
times, and exits non-zero when the outcome varies is attached.

## Expected

Every run yields the identical result. The `#nosec G101` directive covers the
declaration, so the G101 finding should be suppressed on **every** run (count `0`).

## Actual

The G101 count varies run to run over identical source, e.g.:

```
1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 0 0
```

`0` = directive honored (correct); `1` = same directive ignored, finding leaks.
Leak rate is roughly 5–15% per run and rises with more concurrent CPU load.

On the real-world codebase where we first hit this, the same file flapped
between **9 and 10** total findings across 50 CI runs — one suppressed G101
appearing/disappearing — which is what led us here.

## What appears to matter (from bisecting)

The bug only surfaces with a specific combination; simplifying the file hides it:

1. **The finding is anchored to an outer node, not the trigger line.** The
   `#nosec G101` is on the outer `var` declaration, but the credential-looking
   strings are several lines deeper inside the `Long` raw string. gosec reports
   the G101 issue at the declaration's line range (e.g. `10-37`), not at the
   string's own line. Suppression then depends on range-matching, not an exact
   line hit.
2. **There is more than one `#nosec` directive in the file** (here G101 + G304).
   With a single directive the result is stable; adding the second directive is
   what makes the G101 suppression start flapping.
3. **It is line-position sensitive.** Inserting or removing lines (even
   comments) shifts the outcome — sometimes to stable-suppressed, sometimes to
   stable-leaked, sometimes to flapping. This points at nondeterministic
   ordering/iteration when matching findings against suppression ranges, rather
   than a fixed off-by-one.

Taken together this looks like the suppression pass matching issues against
`#nosec` ranges using a data structure whose iteration order (or concurrent
population) is not deterministic, so which range "wins" for a given
outer-node-anchored finding varies between runs.

## Impact

- gosec cannot be used as a blocking CI check: builds fail at random.
- Suppressions that developers added and verified locally silently stop working
  on some runs, so the reverse (a real finding randomly suppressed) is also
  plausible.

## Repro assets attached

- `reproduce.sh` — generates the package and runs the N-times comparison; exits
  `1` when nondeterministic.

## Few runs

1)
```
./reproduce.sh
gosec: Version: dev Git tag:  Build date:
go:    go version go1.26.8 darwin/arm64
runs:  50

Per-run count of the G101 finding that #nosec should suppress (expect all 0):
  0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 1 0 0 1 0 0 0 1 0 0 1 1

NONDETERMINISTIC: outcomes across identical source = { 0 1 }
The suppressed G101 leaked on 7/50 runs.
```
2)
```
 ./reproduce.sh
gosec: Version: dev Git tag:  Build date:
go:    go version go1.26.8 darwin/arm64
runs:  50

Per-run count of the G101 finding that #nosec should suppress (expect all 0):
  0 0 0 0 0 0 0 0 0 0 0 0 1 0 0 0 0 0 0 1 0 0 0 0 0 0 0 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0

NONDETERMINISTIC: outcomes across identical source = { 0 1 }
The suppressed G101 leaked on 3/50 runs.
```
3)
```
 ./reproduce.sh
gosec: Version: dev Git tag:  Build date:
go:    go version go1.26.8 darwin/arm64
runs:  50

Per-run count of the G101 finding that #nosec should suppress (expect all 0):
  0 0 0 0 0 0 0 1 0 0 0 0 1 0 0 1 0 0 1 0 1 1 0 0 0 0 0 0 0 0 0 0 0 1 0 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0

NONDETERMINISTIC: outcomes across identical source = { 0 1 }
The suppressed G101 leaked on 8/50 runs.
```

intSheep:
`#nosec` comments are read from a map, while lookup used only the first matching range. An inner `G304` directive could hide an outer `G101` suppression depending on iteration order. Merging nested ranges also widened the scope of inner directives.

Keep each directive's range, collect all matching suppressions, and process comments in source order so tracked justifications are stable. Added tests for both insertion orders, range boundaries, and multiple justifications, plus the issue's reproducer.

Tested on Go 1.26.8 with `make test-nocache`, `golangci-lint run`, and 60 runs of the reproducer with no G101 leaks. With `-nosec`, the expected G101 finding is still reported.

Fixes #1740. Thanks for the reproducer!

### DIFF
--- a/analyzer.go
+++ b/analyzer.go
@@ -16,6 +16,7 @@
 package gosec
 
 import (
+	"cmp"
 	"context"
 	"errors"
 	"fmt"
@@ -31,6 +32,7 @@ import (
 	"reflect"
 	"regexp"
 	"runtime/debug"
+	"slices"
 	"strconv"
 	"strings"
 
@@ -98,50 +100,30 @@ func (i ignores) parseLine(line string) (int, int) {
 }
 
 func (i ignores) add(file string, line string, suppressions map[string]issue.SuppressionInfo) {
-	is := []ignore{}
-	if _, ok := i[file]; ok {
-		is = i[file]
-	}
-	found := false
 	start, end := i.parseLine(line)
-	for _, ig := range is {
-		if ig.start <= start && ig.end >= end {
-			found = true
-			for r, s := range suppressions {
-				ss, ok := ig.suppressions[r]
-				if !ok {
-					ss = []issue.SuppressionInfo{}
-				}
-				ss = append(ss, s)
-				ig.suppressions[r] = ss
-			}
-			break
-		}
+	ig := ignore{
+		start:        start,
+		end:          end,
+		suppressions: make(map[string][]issue.SuppressionInfo, len(suppressions)),
 	}
-	if !found {
-		ig := ignore{
-			start:        start,
-			end:          end,
-			suppressions: map[string][]issue.SuppressionInfo{},
-		}
-		for r, s := range suppressions {
-			ig.suppressions[r] = []issue.SuppressionInfo{s}
-		}
-		is = append(is, ig)
+	for r, s := range suppressions {
+		ig.suppressions[r] = []issue.SuppressionInfo{s}
 	}
-	i[file] = is
+	// Keep each directive's range intact, even when it is nested in another.
+	i[file] = append(i[file], ig)
 }
 
 func (i ignores) get(file string, line string) map[string][]issue.SuppressionInfo {
 	start, end := i.parseLine(line)
-	if is, ok := i[file]; ok {
-		for _, i := range is {
-			if i.start <= start && i.end >= end || start <= i.start && end >= i.end {
-				return i.suppressions
+	suppressions := make(map[string][]issue.SuppressionInfo)
+	for _, ig := range i[file] {
+		if ig.start <= start && ig.end >= end || start <= ig.start && end >= ig.end {
+			for r, ss := range ig.suppressions {
+				suppressions[r] = append(suppressions[r], ss...)
 			}
 		}
 	}
-	return map[string][]issue.SuppressionInfo{}
+	return suppressions
 }
 
 // The Context is populated with data parsed from the source code as it is scanned.
@@ -911,7 +893,11 @@ func (v *astVisitor) Visit(n ast.Node) ast.Visitor {
 
 // updateIgnores parses comments to find and update ignored rules.
 func (v *astVisitor) updateIgnores() {
-	for c := range v.context.Comments {
+	// Keep tracked suppression justifications in source order.
+	nodes := slices.SortedFunc(maps.Keys(v.context.Comments), func(a, b ast.Node) int {
+		return cmp.Or(cmp.Compare(a.Pos(), b.Pos()), cmp.Compare(a.End(), b.End()))
+	})
+	for _, c := range nodes {
 		v.updateIgnoredRulesForNode(c)
 	}
 }
--- a/analyzer_ignores_internal_test.go
+++ b/analyzer_ignores_internal_test.go
@@ -1,0 +1,128 @@
+package gosec
+
+import (
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io"
+	"log"
+	"reflect"
+	"slices"
+	"testing"
+
+	"github.com/securego/gosec/v2/issue"
+)
+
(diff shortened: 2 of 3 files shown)


analyzer.go
```diff
@@ -911,7 +893,11 @@ func (v *astVisitor) Visit(n ast.Node) ast.Visitor {
 
 // updateIgnores parses comments to find and update ignored rules.
 func (v *astVisitor) updateIgnores() {
-	for c := range v.context.Comments {
+	// Keep tracked suppression justifications in source order.
+	nodes := slices.SortedFunc(maps.Keys(v.context.Comments), func(a, b ast.Node) int {
```
ccojocar, line 897: What's the performance impact of running sort on every AST node?

intSheep: Thanks for raising this. One clarification: `updateIgnores` is called once per source file, before `ast.Walk`; it is not called from `Visit` for every AST node. Also, `v.context.Comments` is an `ast.CommentMap`, so this only collects and sorts nodes that have associated comments.

The added cost is therefore O(C log C) per file, where C is the number of comment-associated nodes. In a focused microbenchmark on my machine, collecting and sorting 10, 100, and 1,000 entries took approximately 0.48 µs, 6.3 µs, and 92 µs respectively; even 10,000 entries took about 1.3 ms.

Based on that, I do not expect a material end-to-end impact for normal source files. I can also add an end-to-end benchmark if you think that would be useful.

ccojocar (approved):

